PR #13, opened 2018-05-14
add headers to cmake target

noamwies:
I updated the CMakeLists.txt to include the headers for better clion experience


gcarleo:
Thank you @noamwies !

PR #17, opened 2018-05-22
Mark AbstractHamiltonian::FindConn as const

femtobit:
`AbstractHamiltonian::FindConn` does not changes the internal state of the Hamiltonian (and judging from the documentation it is not meant to), so it should be marked as const.

(This pull request also contains a tiny unrelated change in a second commit. I could split this up if you want.)


gcarleo:
All good, thanks Damian.

PR #18, opened 2018-05-23
Add matrix wrapper classes for operators

femtobit:
This pull request adds a set of classes which provide a matrix representation of operators. See the commit messages (especially the description of 4004092) for further details.

Beside these classes, two small build system changes are included.


gcarleo:
Hi, could you see what's going wrong here? 

/home/travis/build/netket/netket/Test/Observable/unit-matrixwrapper.cc:140:26: error: 
      constexpr variable 'sqrt5' must be initialized by a constant expression
        constexpr double sqrt5 = std::sqrt(5);

gcarleo:
The rest seems solid, I just suggest to move the wrap class from Observable/ to Hamiltonian/

twesterhout:
@gcarleo C has no notion of `constexpr` and functions from `<cmath>` are thus not marked `constexpr`. Using `const double` instead should fix the error.

twesterhout:
1) Might I ask what the reason for `Apply` member function is? In C++ it is common practice to use `operator()` for this purpose. 
2) Is a matrix wrapper really the place to ask for the dimension of the Hilbert space? I.e. `GetDimension` does not belong in this class, I think.
But then, we have a class with a virtual call operator, which is just `std::function` -- no reason to reinvent the wheel.
I'm pretty sure I'm missing something here so an explanation would be highly appreciated.

gcarleo:
@twesterhout  I agree, that's the issue with `constexpr`

Regarding 1. Yes, maybe defining an operator would be better 

Regarding 2. , the matrix wrapper should at least know the size of the matrix it is wrapping. GetDimension just returns that size

femtobit:
Strictly speaking, the content of `<cmath>` is governed by the C++11 standard ([Section 26.8](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2011/n3242.pdf) and is actually a bit different from the C version but yes, `std::sqrt` is not `constexpr` according to the standard (it could be, in GCC this works, but in clang it doesn't).

In any case, I will change that quickly.

I will also move the implementation to Hamiltonian/ as suggested. In the future, I think we could discuss merging the implementations for Hamiltonian and the observables as a) they share a significant amount of code and b) semantically, a Hamiltonian is a special case of an observable.

twesterhout:
2. Of course it should. My concern is with the fact that it's becoming public interface. Is there a reason for that?

gcarleo:
@femtobit great, thanks! I think that it is more about Operators, which contain the more general notion of both Hamiltonian and Observable. We should rethink a bit how to organize this, agreed.

gcarleo:
@femtobit 

just another thing: 

void SaveEigenValues(const AbstractHamiltonian &hamiltonian,.. 

why not passing Hamiltonian instead of AbstractHamiltonian ? 

Generally speaking, I decide to have abstract classes waiting for Concepts to be adopted in future c++ standards. At that point, Hamiltonian etc, will become concepts, getting rid of those ugly abstract interfaces

twesterhout:
@gcarleo About Concepts TS: Wow! That's very important you've said that. Thanks a lot. I thought the reason you've been using virtual functions so far is because the type of the Machine or Hamiltonian used, depended on the values in configuration files.
If, however, you're trying to emulate Concepts by means of virtual base classes, might I suggest you to use SFINAE instead? One can get pretty close to concepts and avoid performance overhead of virtual function calls. For inspiration, have a look at Eric Niebler's amazing [range-v3](https://github.com/ericniebler/range-v3) library: Ranges TS requires Concepts TS, but range-v3 does a great job at emulating them.

twesterhout:
@femtobit `operator*` and `operator()` mean different things. In Eigen, `matrix * vector` means matrix-vector multiplication, not "applying matrix to vector". Calling a function on the other hand always means applying it to arguments. So the fact that `operator()` is used in Eigen to access elements is just for the lack of a better solution: `operator[]` with multidimensional indices is "impossible" in C++03 (and tricky in newer versions) :)
By no means do I suggest you throw away all the classes with extra functionality. I'm talking about the interface, i.e. functions merely requiring to be able to apply an Operator to a state vector can ask for `std::function` (if run-time polymorphism is required) or even just a template parameter.

femtobit:
> So the fact that operator() is used in Eigen to access elements is just for the lack of a better solution: operator[] with multidimensional indices is "impossible" in C++03 (and tricky in newer versions) :)
True, but still a possible source of inconsistency in the interface, since netket is using Eigen types in many places.

> By no means do I suggest you throw away all the classes with extra functionality. I'm talking about the interface, i.e. functions merely requiring to be able to apply an Operator to a state vector can ask for std::function (if run-time polymorphism is required) or even just a template parameter.
I agree, I think this is what I will do for the time evolution, as to perform a time step this is exactly the case you describe (only needing to apply a function to the state).

gcarleo:
@femtobit last commit is great, just a further small thing, sorry! (seems like dim is an int, see below)

In file included from /home/travis/build/netket/netket/Test/Hamiltonian/unit-matrixwrapper.cc:22:0:
/home/travis/build/netket/netket/NetKet/Hamiltonian/MatrixWrapper/sparse_matrix_wrapper.hpp: In instantiation of ‘void netket::SparseMatrixWrapper<Operator, WfType>::InitializeMatrix(const Operator&) [with Operator = netket::AbstractHamiltonian; WfType = Eigen::Matrix<std::complex<double>, -1, 1>]’:
/home/travis/build/netket/netket/NetKet/Hamiltonian/MatrixWrapper/sparse_matrix_wrapper.hpp:41:25:   required from ‘netket::SparseMatrixWrapper<Operator, WfType>::SparseMatrixWrapper(const Operator&) [with Operator = netket::AbstractHamiltonian; WfType = Eigen::Matrix<std::complex<double>, -1, 1>]’
/home/travis/build/netket/netket/Test/Hamiltonian/unit-matrixwrapper.cc:85:86:   required from here
/home/travis/build/netket/netket/NetKet/Hamiltonian/MatrixWrapper/sparse_matrix_wrapper.hpp:79:29: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
         for(size_t i = 0; i < dim_; ++i)
                             ^
/home/travis/build/netket/netket/NetKet/Hamiltonian/MatrixWrapper/sparse_matrix_wrapper.hpp: In instantiation of ‘void netket::SparseMatrixWrapper<Operator, WfType>::InitializeMatrix(const Operator&) [with Operator = netket::AbstractObservable; WfType = Eigen::Matrix<std::complex<double>, -1, 1>]’:
/home/travis/build/netket/netket/NetKet/Hamiltonian/MatrixWrapper/sparse_matrix_wrapper.hpp:41:25:   required from ‘netket::SparseMatrixWrapper<Operator, WfType>::SparseMatrixWrapper(const Operator&) [with Operator = netket::AbstractObservable; WfType = Eigen::Matrix<std::complex<double>, -1, 1>]’
/home/travis/build/netket/netket/Test/Hamiltonian/unit-matrixwrapper.cc:145:78:   required from here
/home/travis/build/netket/netket/NetKet/Hamiltonian/MatrixWrapper/sparse_matrix_wrapper.hpp:79:29: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]

femtobit:
@gcarleo Good point (and thanks for the quick fix).

The reason `size_t` seeped into my code was that `HilbertIndex::NStates()` returns `size_t` (but internally stores this number as an `int`). Other methods, e.g., `Hilbert::Size()` and `LocalSize`, return `int`s. (And of course, my IDE now rightfully warns about code like `for(int i = 0; i < hindex.NStates(); i++` in other places.)

Following the [Google Coding Standard](https://google.github.io/styleguide/cppguide.html#Integer_Types):

> [...] and try to avoid unsigned types (except for representing bitfields or modular arithmetic). Do not use an unsigned type merely to assert that a variable is non-negative.

Therefore, I think we should replace the uses of size_t in our interface (esp. `HilbertIndex::NStates()`) with a signed integer and agree on whether we want to use `int` or `int64_t` for this (see the Google guidelines above the part I quoted for a good discussion of this).

However, I do not think this needs to be addressed as part of this pull request, as build and tests are fine otherwise.

PR #28, opened 2018-05-29
Prevent duplicate connectors in Hamiltonian

femtobit:
Continuing the discussion from #27:

> But it is also because FindConn is not supposed to return duplicate terms (this is also checked for in one of the unit tests).

If I am not mistaken, the test you refer to currently checks for a different kind of duplication, namely, for duplicate target quantum numbers within each connector. Whether the `connectors` vector has duplicate entries, i.e., there exist `k != j` such that `connectors[k] == connectors[j]`, is not checked. That case happens for the 2x2 PBC cube example, as there e.g. site `(0,0)` has two connectors to `(0,1)` (one "direct" and one "through the boundary").

This pull request adds a check for this scenario and disallows periodic boundaries for L<=2 hypercubes to prevent duplicate connectors there.

The case of merging local operators in `CustomHamiltonian` is still open for now.

### DIFF
--- a/NetKet/Graph/hypercube.hpp
+++ b/NetKet/Graph/hypercube.hpp
@@ -54,6 +54,11 @@ class Hypercube : public AbstractGraph {
       : L_(FieldVal(pars["Graph"], "L")),
         ndim_(FieldVal(pars["Graph"], "Dimension")),
         pbc_(FieldOrDefaultVal(pars["Graph"], "Pbc", true)) {
+    if(pbc_ && L_ <= 2)
+    {
+        std::cout << "L<=2 hypercubes should not have periodic boundary conditions" << std::endl;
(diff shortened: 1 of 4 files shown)


gcarleo:
Thanks @femtobit. Basically what you are doing here is to check that the `Graph` does not have duplicate `Edges`, if I understand correctly. I should know better, but is a similar check already done in `CustomGraph` ? If not, it would be great to add that check to complete this PR!

gcarleo:
Actually the check is already performed here for `CustomGraph` 
https://github.com/netket/netket/blob/56ce86c886701867d2103ab290b5d68b0d18aed3/NetKet/Graph/custom_graph.hpp#L150 

This check would fail if duplicate edges exist

Issue #14, opened 2018-05-18
Exact unitary dynamics on small systems 


gcarleo:
Implement an exact solver for time-dependent Schroedinger equation, mostly for benchmarking purposes and small production results.

PR #29, opened 2018-05-30
Basic functionality for exact unitary time propagation

femtobit:
This pull request contains basic functionality for exact time propagation (Issue #14). The code is not necessarily production ready yet. However, I think this is a good point to discuss what is there so far.

Please let me know what you think of the structure and implementation of the code and whether this goes in the right direction.

If potential issues with this PR are resolved, I would suggest adding a `feature/time-evolution` branch to the main netket repo and merge it there, so we can continue to work on this before merging it into `master`.

#### General ODE solver
* Commits 54de9af, b5b15e8, and 33e948e.

The directory `NetKet/UnitaryTimePropagation/TimeSteppers` contains the code used for ODE propagation (as part of the namespace `netket::ode`), which is not limited to the Schroedinger equation but can in principle be used for any ODE system of the form `dx/dt = F(x, t)`. States should be Eigen vector (or matrix) types. A few standard algorithms (Euler and Runge-Kutta) are provided, though I think in practice one should use `Dopri54` (Dormand-Prince fifth-order Runge-Kutta scheme with adaptive step size). I am not sure how much sense it makes to include the other methods, which I mostly implemented for testing and demonstration purposes. The RK4 scheme might be useful if a fixed time step is desired for some reason.

The main time-stepping loop is implemented in
https://github.com/netket/netket/blob/537de3206c3b0ea38eb8710cef3513f274e1ecb5/NetKet/UnitaryTimeEvolution/TimeStepper/abstract_time_stepper.hpp#L44-L64

My custom implementation is compared in one unit test to the [Boost.Numeric.Odeint](https://www.boost.org/doc/libs/1_67_0/libs/numeric/odeint/doc/html/index.html) library and achieves comparable speed (though I did not test this systematically yet, so take this with a grain of salt).

If desired, additional time stepping methods can be added by subclassing `AbstractTimeStepper`. In principle we could also add time steppers delegating to some external library, if it can work with Eigen matrices as states.

#### Factory methods for JSON

The steppers can be created from JSON parameters via a factory function provided in `time_stepper.hpp`. (The same is now also true for the matrix wrappers, see `matrix_wrapper.hpp`.) Personally, I prefer this approach to directly passing `json` objects to the constructor for the following reasons:

* The Google C++ Style Guide recommends [avoiding complex initialization logic in constructors](https://google.github.io/styleguide/cppguide.html#Doing_Work_in_Constructors) and suggests using factory functions instead.
* Having a JSON constructor tightly couples the class to the configuration file format, making it harder to change that in the future (as many parts of the code are affected) and also making the initialization logic within the class harder to read, because it is mixed with JSON validation and processing.
* Passing the required data as arguments to the constructor reduces potential errors, as at least the types are checked at compile time (making it harder to, e.g., forget or mistype a parameter), while invalid JSON will only be detected at runtime (hopefully).

#### Time evolution code for NetKet systems
* Commit b694c0a

The source file `NetKet/UnitaryTimePropagation/time_evolution.hpp` contains the code for time-propagating quantum states via the Schroedinger equation using the general ODE solver. The class `TimeEvolutionDriver` handles most of the setup and actual time propagation. The time-stepping algorithm and tolerances as well as the desired matrix wrapper (i.e., sparse or dense) can be specified in a JSON config.

An brief example tutorial is included:
https://github.com/netket/netket/blob/537de3206c3b0ea38eb8710cef3513f274e1ecb5/Tutorials/TimeEvolution/ising1d.py#L1-L66

#### Parallelization
Currently, the set of initial states specified in the JSON file is distributed among the MPI processes and each one writes a separate output file for each of the initial states with the corresponding time evolution.

The ODE propagation itself is not parallelized.

#### Sanitizer

When the tests are built with `-DNETKET_Sanitizer=ON`, they fail because of memory errors such as:
```
3: Direct leak of 2400 byte(s) in 1 object(s) allocated from:
3:     #0 0x657c96 in __interceptor_malloc (/home/travis/build/femtobit/netket/build/Test/test-hamiltonian+0x657c96)
3:     #1 0x7f56a1683e96  (<unknown module>)
3:     #2 0x7f56a1683d98  (<unknown module>)
3:     #3 0x7f56a02076fb  (<unknown module>)
3:     #4 0x7f56a85d890b in mca_base_components_open (/usr/lib/libmpi.so.1+0xf890b)
3:     #5 0x7f56a8577b7b in mca_pml_base_open (/usr/lib/libmpi.so.1+0x97b7b)
3:     #6 0x7f56a853a198 in ompi_mpi_init (/usr/lib/libmpi.so.1+0x5a198)
3:     #7 0x7f56a855165f in MPI_Init (/usr/lib/libmpi.so.1+0x7165f)
3:     #8 0x873305 in main /home/travis/build/femtobit/netket/Test/unit-tests.cc:7:3
3:     #9 0x7f56a709af44 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x21f44)
3:     #10 0x57f64e in _start (/home/travis/build/femtobit/netket/build/Test/test-hamiltonian+0x57f64e)
```
See https://travis-ci.org/femtobit/netket/jobs/385823408 (or soon probably the Travis build of this pull request) for more. (Commit 3e44adb leads to the more detailed ASan output shown above. This makes running the tests a lot slower, unfortunately.) I am not sure why this occurs and whether something in my code is at fault. From the log shown, the errors all seem to be related to the MPI library and occur in all unit tests (most of which are not changed by this pull request). Yet, none of these show up in the builds of `netket/master`, which suggests something in this branch causes them. Any idea?

In any case, thanks for taking the time to read this somewhat long description. Looking forward to your feedback on this PR.


twesterhout:
1) Matrix wrapper related code leaks memory (this might very well be the reason why the tests fail).
2) Why do you re-implement different steppers rather than just use Boost.Numeric.Odeint?

femtobit:
Ad 1.: Could you point me to where it leaks memory?

Ad 2.: The implementation of the required algebra operations in `boost/numeric/odeint/external/eigen/eigen_algebra.hpp` does not work with current versions of Eigen (cf. headmyshoulder/odeint-v2#194, the solutions proposed there did not work for me) and it seems this issue has been present in Odeint for quite some time now. I tried to implement the required algebra operations to get `Eigen::Matrix` to work with controlled steppers myself but did not succeed so far. I would, however, welcome a pull request adding this support.

One could also use `std::vector` as state type for Odeint and wrap it with `Eigen::Map` as in the test below but I do not think this is elegant.
https://github.com/netket/netket/blob/54e190b853fd06f6dfabc0a20279cda9d3874a0a/Test/UnitaryTimeEvolution/unit-timeevolution.cpp#L267-L272

twesterhout:
1) I've tested it locally, and it might be not as obvious as I thought... My bad, sorry for the noise!
2) When you tried implementing the algebra yourself, what was the showstopper?

gcarleo:
@femtobit , thanks for the detailed PR, I am looking into it! 

A few comments already, more will come as soon as I settle down a bit: 

1. Please have a look at the warnings raised during compiling, for example 
```
/Users/travis/build/netket/netket/NetKet/UnitaryTimeEvolution/time_evolution.hpp:125:31: warning: comparison of integers of different signs: 'size_type' (aka 'unsigned long') and 'int' [-Wsign-compare]
``` 
and similar others. Those are nothing serious, but it is nice to have a warning-free experience!  

2. I 100% agree with the design choice of NOT using Boost (also in reply to @twesterhout). Boost is a monster library, and adding dependency to it just for a few lines of ode-int solver is something we don't want to do. Actually, let's make sure that unit tests do not depend on Boost either, unless some flag is explicitly set to ON in the CMake configuration! 

3. The memory leak is puzzling. I guess it is MPI related, but I can't understand why it manifests **in the other tests** only when you try your new functionality, that's really weird. 

4. What do you have in the output files at this point? We should be consistent with the other cases (in `Learning`) and have a Json output with the desired observables. The design goal here would be to have a section `Observables`, like for the other methods we have, and add there observables we want to be measured as a function of time. The output file would then contain those values, in Json format.

5. I would suggest to change `NetKet/UnitaryTimePropagation` in something a bit more general (see also comment 6.). Maybe the best would be to group together all classes solving similar problems, either in an exact way or with learning. The directory structure should reflect the input section. A first proposal is to remove the `Learning` section and just create a section `GroundState` for which as a method you can either specify `FullEd`, `SparseEd`,  `Sr`, `Gdesc` and more and a section `Dynamics` for which as a method you can specify `FullEd`, `SparseEd`, (with possibly sub-fields to ask for imaginary-time dynamics, see 6.), `t-VMC` (for the future) etc. For backward compatibility in all versions v1.x.x, we will still need to support the `Learning` field though. 

6. It is a straightforward (but important!) extension, to have this routine work also in imaginary-time, (in the non-unitary regime...) to converge to the ground state, as an alternative to the (being-implemented) Lanczos methods.

femtobit:
Some quick points on the remarks by @gcarleo (probably more later):

1. Sorry about the warnings. I will fix them when I'm back at my computer.

2. Boost is indeed only used in the unit test if it is found on the system. Otherwise that test case is not compiled.

4. Currently, just the quantum state is written to the output. This is only temporary, I plan to add computation of the observables. As this is related to the general design (cf. your points 5. and 6.) I wanted to discuss this first.

gcarleo:
@femtobit 

1. No problem, it's a minor thing
2. Fantastic, that's exactly the way to go. 
3. Yes, the output indeed is related to design. Let's stick to the design used for observables when doing `Learning`, basically 
```python
pars['Observables']={
    'Operators'      : sigmaxop,
    'ActingOn'       : sites,
    'Name'           : 'SigmaX',
}
```
then you might want to add an option whether to save or not the full state in the output Json file.

NetKet/UnitaryTimeEvolution/time_evolution.hpp
```diff
@@ -0,0 +1,142 @@
...
+          stepper_(std::move(stepper)),
+          t0_(t0), tmax_(tmax), dt_(dt)
+    {
+        ode_system_ = [this](const VectorType& x, VectorType& dxdt, double /*t*/) {
+            static const std::complex<double> im{0., 1.};
+            dxdt.noalias() = -im * hmat_->Apply(x);
+        };
+    }
+
+    void Run(VectorType& initial, ode::ObserverFunction<VectorType> observer_func)
```
gcarleo: Shouldn't be initial be `const` valued?

NetKet/UnitaryTimeEvolution/time_evolution.hpp
```diff
@@ -0,0 +1,142 @@
...
+    {
+        assert(initial.size() == GetDimension());
+
+        ode::Integrate<StepperType, VectorType>(
+                    *stepper_, ode_system_, initial, t0_, tmax_, dt_, observer_func);
```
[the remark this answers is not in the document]
gcarleo: Maybe stepper_.get() ?

gcarleo:
Something I forgot! My personal taste is against CRTP, since it introduces a lot of syntactic clutter and the code is much harder to read (at least for me). Also, modern compiler "devirtualize", and allow to obtain comparable performances. In practice, I don't think there is a measurable speed advance when using CRTP in this specific case, since calling those methods shouldn't be the computational bottleneck

NetKet/UnitaryTimeEvolution/time_evolution.hpp
```diff
@@ -0,0 +1,142 @@
...
+        double dt = pars_te["TimeStep"].get<double>();
+
+        return TimeEvolutionDriver(std::move(matrix), std::move(stepper),
+                                   t0, tmax, dt);
+    }
+
+    TimeEvolutionDriver(std::unique_ptr<MatrixType> matrix,
+                        std::unique_ptr<StepperType> stepper,
+                        double t0, double tmax, double dt)
+        : hmat_(std::move(matrix)),
```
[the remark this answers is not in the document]
femtobit: The point of using `unique_ptr` for these parameters is to signal that this transfers ownership of stepper and matrix wrapper to the TimeEvolutionDriver, which is then also responsible for deallocating them when it is destructed.

After calling `TimeEvolutionDriver(std::move(matrix), ...)` the ownership of `matrix` is transferred to the driver object and the `matrix` unique pointer is guaranteed to be empty. So, as you note, you should not dereference the outer pointer anymore (this would be undefined behaviour), but e.g. [checking whether it is empty](https://en.cppreference.com/w/cpp/memory/unique_ptr/operator_bool) is safe (and will return `true`).

The alternative would be to just pass a raw pointer, but then it would be less clear which part of the code is responsible for deallocating the objects in the end.

NetKet/UnitaryTimeEvolution/time_evolution.hpp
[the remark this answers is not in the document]
femtobit: Initial is passed directly to `ode::Integrate` which indeed changes the initial value. Alternatively, one could do something like
```cpp
void Run(const VectorType& initial, ...)
{
    VectorType state = initial; // copy
    ode::Integrate<...>(state, ...)
}
```
which leaves `initial` unchanged.

NetKet/UnitaryTimeEvolution/time_evolution.hpp
```diff
@@ -0,0 +1,142 @@
...
+    double tmax_;
+    double dt_;
+};
+
+void RunTimeEvolution(const json& pars)
+{
+    auto driver = TimeEvolutionDriver::FromJson(pars);
+
+    auto pars_te = pars["TimeEvolution"];
+    auto confs = pars_te["InitialStates"];
```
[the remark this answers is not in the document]
femtobit: Good point, thanks.

NetKet/UnitaryTimeEvolution/time_evolution.hpp
```diff
@@ -0,0 +1,142 @@
...
+            std::abort();
+        }
+
+        TimeEvolutionDriver::VectorType initial =
+                Eigen::Map<TimeEvolutionDriver::VectorType>(
+                    initial_vec.data(), initial_vec.size());
+
+        driver.Run(initial, observer_func);
+    }
+}
```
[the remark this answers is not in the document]
femtobit: I will add it.

NetKet/UnitaryTimeEvolution/TimeStepper/time_stepper.hpp
```diff
@@ -0,0 +1,62 @@
...
+std::unique_ptr<Stepper> ExplicitStepperFromJson(const json& pars, int dim)
+{
+    double dt = FieldVal(pars, "InternalTimeStep");
+    return std::unique_ptr<Stepper>(new Stepper(dt, dim));
+}
+
+}
+
+template<class State>
+std::unique_ptr<AbstractTimeStepper<State>>
```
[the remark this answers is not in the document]
femtobit: Returning `unique_ptr` from factory functions is safe and, e.g., recommended by Herb Sutter [in this blog post](https://herbsutter.com/2013/05/30/gotw-90-solution-factories/) which contains a nice discussion on why this is a good idea.

As far as I understand, what happens is the following: `std::unique_ptr` cannot be copied (if the pointer would be copied on return, this would indeed be unsafe as the original pointer would deallocate the object). However, returning it by value from this function actually results in the pointer being moved out of the function (which does not destroy its content) because of [copy elision](https://stackoverflow.com/questions/4316727/returning-unique-ptr-from-functions) and thus transferring ownership of the object to the calling code.

femtobit:
Good point regarding the CRTP, I have removed it now.

NetKet/UnitaryTimeEvolution/time_evolution.hpp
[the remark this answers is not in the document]
gcarleo: Sure you are right, I was just confused by the name `initial`, since I would have expected that the initial state is unchanged.

gcarleo:
@femtobit Overall the code is great, and it's also rather fast, which is also particularly important. It's just really puzzling to see this memory leak. Maybe we should try a different MPI implementation (say MPICH) to see if it is library specific?

gcarleo:
This memory leak is really annoying. I can't reproduce it locally on my laptop, and on Travis it seems independent on the MPI implementation (tried both openmpi and mpich). Has any of you tried with valgrind? (it's a pain to install it on Mac OS) @twesterhout  @femtobit @verticube

twesterhout:
@gcarleo Yes, I've run the example under `valgrind`. Just like LeakSanitizer, it reports quite a bit of "definitely lost" events. I haven't had a thorough look through the code yet, I'll do that if I find the time. In the meantime, here are two ideas how to proceed:
1) Step through each commit to locate the offending one.
2) Build MPI in debug mode to get a proper stacktrace.

NetKet/Hamiltonian/MatrixWrapper/matrix_wrapper.hpp
```diff
@@ -0,0 +1,40 @@
...
+template<class Wrapped>
+std::unique_ptr<AbstractMatrixWrapper<Wrapped>>
+ConstructMatrixWrapper(const json &pars, const Wrapped& wrapped)
+{
+    using WrapperPtr = std::unique_ptr<AbstractMatrixWrapper<Wrapped>>;
+
+    std::string wrapper_name = FieldOrDefaultVal<json, std::string>(pars, "MatrixWrapper", "Sparse");
+    if(wrapper_name == "Sparse")
+    {
+        return WrapperPtr(new SparseMatrixWrapper<Wrapped>(wrapped));
```
twesterhout: This is *not* exception safe. You're copying `wrapped` (which is a bad idea in the first place) whose copy constructor may throw an exception. Now, if an exception is thrown during construction, the newly allocated pointer will never be freed, i.e. you're leaking memory.

NetKet/Hamiltonian/MatrixWrapper/matrix_wrapper.hpp
```diff
@@ -0,0 +1,40 @@
...
+    }
+    else if(wrapper_name == "Dense")
+    {
+        return WrapperPtr(new SparseMatrixWrapper<Wrapped>(wrapped));
```
twesterhout: Are you sure a `SparseMatrixWrapper` should be created even though the user asked for a "Dense" wrapper?

NetKet/Hamiltonian/MatrixWrapper/matrix_wrapper.hpp
```diff
@@ -0,0 +1,40 @@
...
+    }
+    else
+    {
+        std::cout << "Unknown MatrixWrapper: " << wrapper_name << std::endl;
+        std::abort();
```
twesterhout: Why'd you call abort here? Throwing an exception should do the job.

NetKet/UnitaryTimeEvolution/TimeStepper/abstract_time_stepper.hpp
```diff
@@ -0,0 +1,86 @@
+#ifndef NETKET_ABSTRACT_TIME_STEPPER_HPP
+#define NETKET_ABSTRACT_TIME_STEPPER_HPP
+
+#include <functional>
+#include <iostream>
```
twesterhout: What do you need this for?

NetKet/UnitaryTimeEvolution/TimeStepper/abstract_time_stepper.hpp
```diff
@@ -0,0 +1,86 @@
...
+               Time t0, Time tmax, Time dt,
+               ObserverFunction<State> observer = NullObserver<State, Time>)
+{
+    assert(t0 < tmax);
+    assert(dt > .0);
+
+    observer(state, t0);
+
+    int step = 0;
+    double t = t0;
```
twesterhout: What's the point in the `Time` template parameter if you end up using `double`s anyway?

NetKet/UnitaryTimeEvolution/TimeStepper/abstract_time_stepper.hpp
```diff
@@ -0,0 +1,86 @@
...
+ * @param tmax The end time of the propagation. If at time t the next step t + dt
+ *      is > tmax, the state is only propagated to tmax.
+ * @param dt The time step. Time steppers may use a smaller time step internally
+ *      (e.g., as part of an adaptive algorithm).
+ * @param observer This callable is called at each time step, i.e., at times
+ *      t0, t0 + dt, t0 + 2*dt, ..., tmax.
+ */
+template<class Stepper, class State, typename Time = double>
+void Integrate(Stepper& stepper,
+               OdeSystemFunction<State> ode_system,
```
twesterhout: This function is a template function. Then why'd you use `std::function` for function parameters? First of all, you get an extra virtual call. Secondly, you end up copying the function even when an rvalue is passed. Finally, you're just over-constraining the arguments's type.
A simple `... class OdeSytemFunction ...` and `... OdeSystemFunction&& ode_system ...` should do the job.

NetKet/UnitaryTimeEvolution/TimeStepper/controlled_time_steppers.hpp
```diff
@@ -0,0 +1,266 @@
...
+
+public:
+    void Propagate(OdeSystemFunction<State> ode_system,
+                   State &x,
+                   Time t, Time dt) final override
+    {
+        assert(dt > .0);
+        Time tmax = t + dt;
+
+        last_x_.resize(x.size());
```
twesterhout: Is this `resize` really needed?

NetKet/UnitaryTimeEvolution/TimeStepper/controlled_time_steppers.hpp
```diff
@@ -0,0 +1,266 @@
...
+     * @param delta A local error estimate, usually computed as the difference between
+     *      the resulting state for two different-order methods.
+     * @param norm The norm of the current state, used for rescaling rtol.
+     * @return The relative local error of the solution. This will be a
+     *      non-negative number that is <= 1 iff the error is within the
+     *      bound given by atol and rtol.
+     */
+    double LocalError(const State& delta, double norm) const
+    {
+        assert(norm >= .0);
```
twesterhout: It might be a good idea to add something like `assert(delta.size() > 0)` here, because otherwise you might end up dividing by zero.

NetKet/UnitaryTimeEvolution/TimeStepper/controlled_time_steppers.hpp
```diff
@@ -0,0 +1,266 @@
...
+    using Base = ControlledStepperBase<State, Time>;
+
+    std::array<State, 7> k_;
+    State x_temp1_;
+    State delta_;
+
+    bool has_last_;
+
+    /* Butcher tableau */
+    static const int N_ = 7;
```
twesterhout: Why not `constexpr`?

NetKet/UnitaryTimeEvolution/TimeStepper/explicit_time_steppers.hpp
```diff
@@ -0,0 +1,141 @@
...
+                   Time t, Time dt) final override
+    {
+        assert(dt > .0);
+        Time tmax = t + dt;
+
+        int step = 0;
+        Time current_t = t;
+        while(current_t < tmax)
+        {
+            double next_dt = (current_t + internal_dt_ < tmax) ? internal_dt_ : tmax - current_t;
```
twesterhout: `Time` vs `double` again.

NetKet/UnitaryTimeEvolution/TimeStepper/time_stepper.hpp
```diff
@@ -0,0 +1,62 @@
...
+namespace netket { namespace ode {
+
+namespace detail {
+
+template<class Stepper>
+std::unique_ptr<Stepper> ControlledStepperFromJson(const json& pars, int dim)
+{
+    double atol = FieldOrDefaultVal(pars, "AbsTol", 1e-9);
+    double rtol = FieldOrDefaultVal(pars, "RelTol", 1e-9);
+    return std::unique_ptr<Stepper>(new Stepper(atol, rtol, dim));
```
twesterhout: Again, not exception-safe.

NetKet/UnitaryTimeEvolution/TimeStepper/time_stepper.hpp
```diff
@@ -0,0 +1,62 @@
...
+        return detail::ExplicitStepperFromJson<RungeKutta4Stepper<State>>(pars, dim);
+    }
+    else if(stepper_name == "Euler")
+    {
+        return detail::ExplicitStepperFromJson<EulerTimeStepper<State>>(pars, dim);
+    }
+    else
+    {
+        std::cout << "Unknown TimeStepper: " << stepper_name << std::endl;
+        std::abort();
```
twesterhout: `std::abort()` again.

NetKet/UnitaryTimeEvolution/time_evolution.hpp
```diff
@@ -0,0 +1,142 @@
...
+
+    TimeEvolutionDriver(std::unique_ptr<MatrixType> matrix,
+                        std::unique_ptr<StepperType> stepper,
+                        double t0, double tmax, double dt)
+        : hmat_(std::move(matrix)),
+          stepper_(std::move(stepper)),
+          t0_(t0), tmax_(tmax), dt_(dt)
+    {
+        ode_system_ = [this](const VectorType& x, VectorType& dxdt, double /*t*/) {
+            static const std::complex<double> im{0., 1.};
```
twesterhout: `constexpr`?

NetKet/Utils/math_helpers.hpp
```diff
@@ -0,0 +1,30 @@
...
+#include <functional>
+
+/**
+ * Returns v if it is in the interval [lo, hi] or the closest of the bounds,
+ * if v is outside. Uses comp to compare the values.
+ *
+ * This is a replacement of std::clamp which was introduced in C++17.
+ */
+template<class T, class Comparator>
+constexpr const T& clamp(const T& v, const T& lo, const T& hi, Comparator comp)
```
twesterhout: Please, Do not pollute the global namespace :)

NetKet/Utils/math_helpers.hpp
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Returns v if it is in the interval [lo, hi] or the closest of the bounds,
+ * if v is outside. Uses operator< to compare the values.
+ *
+ * This is a replacement of std::clamp which was introduced in C++17.
+ */
+template<class T>
+constexpr const T& clamp(const T& v, const T& lo, const T& hi)
+{
+    return clamp(v, lo, hi, std::less<T>());
```
twesterhout: This won't compile under C++17 because the call to `clamp` in ambiguous (ADL).

NetKet/netket.cc
```diff
@@ -22,7 +22,21 @@ int main(int argc, char *argv[]) {
 
   auto pars = netket::ReadJsonFromFile(argv[1]);
 
-  netket::Learning learning(pars);
+  if(netket::FieldExists(pars, "Learning"))
+  {
+      std::cout << "Hallo" << std::endl;
```
twesterhout: ??

NetKet/Utils/stopwatch.hpp
```diff
@@ -0,0 +1,33 @@
+#ifndef NETKET_STOPWATCH_HPP
+#define NETKET_STOPWATCH_HPP
+
+#include <chrono>
+
+namespace netket {
+
+class Stopwatch
```
twesterhout: Is this class actually used anywhere?

twesterhout:
Okay, it appears that there are no memory leaks in this PR (at least not until the removal of CRTP). This is an MPI implementation issue, or OpenMPI issue to be more specific.

NetKet/UnitaryTimeEvolution/TimeStepper/abstract_time_stepper.hpp
femtobit: Ah, I overlooked that one. Though, to be honest, I'm still thinking about removing the `Time` template parameter everywhere (it was a leftover from my initial version of this code which had even more template parameters), as it is probably unlikely that we will use anything other than `double` in the future. @gcarleo What do you think?

NetKet/Utils/stopwatch.hpp
femtobit: Yes, in the unit tests.

NetKet/Hamiltonian/MatrixWrapper/matrix_wrapper.hpp
femtobit: In principle I very much agree. Since calling `abort` on invalid input is currently done all over the code, however, I planned to do just the same here for now and discuss improving the error handling by using exceptions in another pull request.

NetKet/Hamiltonian/MatrixWrapper/matrix_wrapper.hpp
femtobit: The constructor of `SparseMatrixWrapper` takes `const Wrapped&` so there should not be a copy involved. Or am I missing something?

NetKet/UnitaryTimeEvolution/TimeStepper/abstract_time_stepper.hpp
gcarleo: Definitely remove it ! We will need to do imaginary-time evolution as well, but that's more natural to do in other ways (i.e. changing the o.d.e. function itself, which in your framework can be done easily)

NetKet/Hamiltonian/MatrixWrapper/matrix_wrapper.hpp
gcarleo: Indeed the `abort` solution I adopted is very bad, and an ad interim solution waiting for a better system to intercept errors

femtobit: @gcarleo I would suggest changing all places were `abort` is called to throw a suitable exception and then handle it in the main method by printing the error message and exiting. That should be sufficient for invalid input. I could prepare a pull request with that in the next few days, if you agree.

NetKet/UnitaryTimeEvolution/TimeStepper/abstract_time_stepper.hpp
femtobit: Hm, yes, in principle I would prefer doing that for `Integrate`. However, this would require changes in more places: The ODE function type is needed for `Stepper::Propagate` which is currently a virtual function and thus cannot have template parameters itself and making it non-virtual would change the design in more places, if I am not mistaken. (And since the rest of netket also uses abstract base classes to specify interfaces, I would like to keep doing that here as well.)

I would hope the overhead from using `std::function` is not too large in practice, though I have not checked.

femtobit:
When running the tests locally (Linux, clang 6.0.0, OpenMPI 3.1.0), I even get the MPI-related leak reports for the netket/master branch (currently e9fe700aa). Strangely, they do not appear in Travis for [the same commit](https://travis-ci.org/netket/netket/jobs/387196002).

Can someone else confirm that the leak reports are present with clang and `-DNETKET_Sanitizer=ON` on netket/master?

femtobit:
Aside from the openmpi issue, I think there are two main points still open here:

1. Reorganizing the code as suggested in https://github.com/netket/netket/pull/29#issuecomment-393520691 (points 5. and 6.)
2. Add sampling of observables during the time evolution and make output configurable.

As the discussion here is quite long already, I suggest addressing those in a second pull request later (and merging this to a development branch.like feature/time-evolution in the meanwhile).

femtobit:
Okay, merged into `feature/time_evolution`.

NetKet/Hamiltonian/MatrixWrapper/matrix_wrapper.hpp
twesterhout: There sure is. You're storing a matrix, not a reference to it. By passing `Wrapped const&` you're invoking a [copy constructor](http://en.cppreference.com/w/cpp/language/copy_constructor).

NetKet/Hamiltonian/MatrixWrapper/matrix_wrapper.hpp
twesterhout: Sounds good. Standard arguments against using exceptions (predictable performance etc.) do not apply to this projects, so I would suggest to stick to them.

NetKet/Utils/stopwatch.hpp
twesterhout: Then shouldn't this header be located somewhere in `Test/?

twesterhout:
@femtobit Ignore the errors you get with Open MPI (see https://stackoverflow.com/questions/8471773/mpi-memory-leak). Everything runs well for me with on Linux compiled with Clang-7 and MPICH. Hence my previous commit that your PR is okay.

gcarleo:
@twesterhout OK, indeed when I do things locally on my machine I also don't find those memory leaks. But I thought we also had this issue with MPICH on Travis?

twesterhout:
@gcarleo Could you pinpoint me to the commit which was built on Travis with MPICH?

gcarleo:
@twesterhout , this one https://travis-ci.org/netket/netket/jobs/385900803/config

twesterhout:
@gcarleo Thanks... Apparently, I can't simply checkout this revision. Could #30 have something to do with the errors? It hadn't been fixed at that point yet. Also, you're compiling tests in Release mode. Do you mind explaining why? I believe it's common to compile tests in Debug mode because you want as many checks as you can get. For example, in Release mode #30 wouldn't even be triggered (i.e. you'd only see errors from the AddressSanitizer).

gcarleo:
@twesterhout , I guess it would be good to test the latest commits agains MPICH then. You are right about Debug mode vs Release mode. The issue is that in Debug mode, the code is compiled with -O0 or something similar, and it takes ages to complete the tests. Maybe the best solution would be to modify optimisation flags only while in Debug mode... I am not an expert of CMakes, I would be very happy if you or @femtobit would generally help improving the current CMake configurations

PR #31, opened 2018-06-08
Throw exceptions for invalid user input

femtobit:
This pull request adds a custom exception class and replaces all calls to `std::abort` for invalid user input with throwing this exception.

This makes the code easier to read in most parts and allows us to handle invalid input in a single place (currently in `main` by printing an error message and exiting).

This PR also adds a point to the list of exceptions from the Google C++ Guide in `CONTRIBUTING.md` because the Google guide does not permit the use of exceptions (for reasons that in my opinion do not apply to netket).

### DIFF
--- a/NetKet/Hamiltonian/bosonhubbard.hpp
+++ b/NetKet/Hamiltonian/bosonhubbard.hpp
@@ -19,6 +19,10 @@
 #include <cmath>
 #include <iostream>
 #include <vector>
+
+#include "Utils/exceptions.hpp"
+#include "Utils/json_helper.hpp"
+
 #include "abstract_hamiltonian.hpp"
 
 namespace netket {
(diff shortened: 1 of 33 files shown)


NetKet/Utils/json_helper.hpp
```diff
@@ -21,40 +21,50 @@
...
 }
 
-template <class T>
-T FieldVal(const T &pars, std::string field) {
-  if (!FieldExists(pars, field)) {
-    std::cerr << "Field " << field << " is not defined in the input"
-              << std::endl;
-    std::abort();
+template<class Json>
+void CheckFieldExists(const Json& pars, const std::string& field, const std::string& context = "")
```
gcarleo: can you please add a small comment on the role of "context" here in this function? 
So that we remember in the future what's this about

NetKet/Hamiltonian/bosonhubbard.hpp
```diff
@@ -54,28 +58,17 @@ class BoseHubbard : public AbstractHamiltonian {
     if (FieldExists(pars["Hamiltonian"], "Nmax")) {
```
gcarleo: better to use the new `CheckFieldExists` function?

NetKet/Hamiltonian/bosonhubbard.hpp
```diff
@@ -54,28 +58,17 @@ class BoseHubbard : public AbstractHamiltonian {
...
       nmax_ = pars["Hamiltonian"]["Nmax"];
     } else {
-      std::cerr << "Nmax is not specified for bosons" << std::endl;
-      std::abort();
+      throw InvalidInputError("Nmax is not specified for bosons");
     }
 
     if (FieldExists(pars["Hamiltonian"], "U")) {
```
gcarleo: better to use the new `CheckFieldExists` function?

NetKet/Hamiltonian/custom_hamiltonian.hpp
```diff
@@ -29,36 +29,28 @@ class CustomHamiltonian : public AbstractHamiltonian {
 
   explicit CustomHamiltonian(const json &pars) : hilbert_(pars) {
```
gcarleo: better to use the new `CheckFieldExists` function?

NetKet/Hamiltonian/custom_hamiltonian.hpp
```diff
@@ -29,36 +29,28 @@ class CustomHamiltonian : public AbstractHamiltonian {
...
+      throw InvalidInputError("Hamiltonian: Local operators are not defined");
     }
 
     if (!pars["Hamiltonian"]["Operators"].is_array()) {
-      std::cerr << "Local operators is not an array" << std::endl;
-      std::abort();
+      throw InvalidInputError("Hamiltonian: Local operators is not an array");
     }
 
     if (!FieldExists(pars["Hamiltonian"], "ActingOn")) {
```
gcarleo: better to use the new `CheckFieldExists` function?

femtobit:
Thanks for the feedback, I have amended the PR accordingly. (I also added a short unit test to make sure the JSON helpers throw as expected.)

femtobit:
> there are places where exception are thrown only if (mynode == 0). Are we sure this is the good way to go? Wouldn't it be better instead to print the error message only if mynode==0, but throw for all nodes ?

Ah, thanks. Indeed, that is what I intended, I just overlooked a few places. Should be fixed now.

PR #32, opened 2018-06-08
Compile Travis tests without -DNDEBUG to trigger assertion errors

femtobit:
By passing `-DCMAKE_CXX_FLAGS_RELEASE=-O3` we leave the optimization level high but remove the `-DNDEBUG` flag that cmake sets by default for release builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -104,6 +104,6 @@ script:
 
   # compile and execute unit tests
   - mkdir -p build && cd build
-  - cmake .. ${CMAKE_OPTIONS} && cmake --build . --config Release
+  - cmake .. ${CMAKE_OPTIONS} -DCMAKE_CXX_FLAGS_RELEASE=-O3 && cmake --build . --config Release
   - ctest -C Release -V -j
   - cd ..


twesterhout:
First of all, what about debug symbols? Debug mode does not only imply less optimisations, but also the presence of debug symbols. Secondly, Release mode should not simply be `-DNDEBUG -O3` if you want the best performance. And finally, what's wrong with tests running for a long time? As long as you don't exceed Travis' (50 min. / job) timeouts, of course.

gcarleo:
I agree, but this change affects only the travis script, and is contained within the CI environment. The problem is that testing on travis currently already takes ages (we don't have a dedicated CI server **yet**, we should have one soon). Until we have faster hardware for testing, it makes sense to go with `-O3` to speed up a little bit the tests. Of course it would be better to work directly on the `CMake` configuration here, and make sure that when testing is enabled we don't miss any debug symbol. (Actually now I also realise that we are missing an option in CMake to turn on/off testing at compile time, that would be very useful to speed up compilation when tests are not needed).

PR #36, opened 2018-06-15
Feedforward

kchoo1118:


NetKet/Machine/activations.hpp
```diff
@@ -0,0 +1,73 @@
+#include <Eigen/Dense>
```
gcarleo: Please add the header here

NetKet/Machine/abstract_layer.hpp
```diff
@@ -0,0 +1,79 @@
...
+
+  virtual void InitLookup(const Eigen::VectorXd &v, LookupType &lt) = 0;
+
+  virtual void UpdateLookup(const Eigen::VectorXd &v,
+                            const std::vector<int> &tochange,
+                            const std::vector<double> &newconf,
+                            LookupType &lt) = 0;
+
+  /**
+  Member function doing..
```
gcarleo: Incomplete comments

NetKet/Machine/activations.hpp
```diff
@@ -0,0 +1,73 @@
...
+  // g = J * f = f
+  // Z = [z1, ..., zn], G = [g1, ..., gn], F = [f1, ..., fn]
+  // Note: When entering this function, Z and G may point to the same matrix
+  inline void apply_jacobian(const VectorType &Z, const VectorType &A,
+                             const VectorType &F, VectorType &G) {
+    for (int i = 0; i < G.size(); ++i) {
+      G(i) = F(i) * std::tanh(Z(i));
+    }
+    (void)A;
+    // G.array() = F.array()*Z.array().tanh();
```
gcarleo: This version should be faster, why is it commented?

NetKet/Machine/activations.hpp
```diff
@@ -0,0 +1,73 @@
...
+#ifndef NETKET_ACTIVATIONS_HPP
+#define NETKET_ACTIVATIONS_HPP
+
+namespace netket {
+
+class AbstractActivation {
+ public:
+  using VectorType = Eigen::Matrix<std::complex<double>, Eigen::Dynamic, 1>;
+
+  virtual inline void activate(const VectorType &Z, VectorType &A) = 0;
```
femtobit: The rest of netket uses `CamelCase` for member functions. Is there a reason to do it differently here?

NetKet/Machine/ffnn.hpp
```diff
@@ -0,0 +1,248 @@
...
+  using LookupType = typename AbstractMachine<T>::LookupType;
+
+  // constructor
+  explicit FFNN(const Hilbert &hilbert, const json &pars)
+      : nv_(hilbert.Size()), hilbert_(hilbert) {
+    from_json(pars);
+  }
+
+  void from_json(const json &pars) override {
+    auto layers_par = pars["Machine"]["Layers"];
```
gcarleo: Check that Layers is defined in the input otherwise throw error

NetKet/Machine/fullconn_layer.hpp
```diff
@@ -0,0 +1,246 @@
...
+  int mynode_;
+
+ public:
+  using StateType = typename AbstractLayer<T>::StateType;
+  using LookupType = typename AbstractLayer<T>::LookupType;
+
+  /// Constructor
+  FullyConnected(const int input_size, const int output_size,
+                 const bool use_bias = true)
+      : activation_(),
```
gcarleo: Why do you call the constructor here?

Tutorials/Heisenberg1d/heisenberg1d.py
```diff
@@ -50,11 +50,12 @@
 pars['Learning']={
```
gcarleo: Is it a faster setting for Tutorials? Does it reach the same accuracy as before?

kchoo1118: Oh I just changed it from AdaMax to Sgd.

NetKet/Machine/fullconn_layer.hpp
kchoo1118: That is the activation class.

kchoo1118: Because not all layers need an activation. Maybe you have some pooling layers?

NetKet/Machine/abstract_layer.hpp
```diff
@@ -0,0 +1,79 @@
...
+
+  virtual VectorType Output() const = 0;
+
+  virtual void Backprop(const VectorType &prev_layer_data,
+                        const VectorType &next_layer_data) = 0;
+
+  virtual const VectorType &Backprop_data() const = 0;
+
+  virtual void GetDerivative(VectorType &der, int start_idx) = 0;
+};
```
gcarleo: Please include a virtual destructor to cure the memory leaks dedected in Travis 
```c++
virtual ~AbstractLayer() {}
```

NetKet/Machine/activations.hpp
```diff
@@ -0,0 +1,73 @@
...
+  virtual inline void apply_jacobian(const VectorType &Z, const VectorType &A,
```
gcarleo: Please include a virtual destructor to cure the memory leaks dedected in Travis

NetKet/Machine/activations.hpp
kchoo1118: Thats true. I was testing, I forgot to change it back.

Tutorials/Heisenberg1d/heisenberg1d.py
gcarleo: It's better, we should just check that this is consistent with what is written in the webpage

NetKet/Machine/activations.hpp
kchoo1118: added

NetKet/Machine/ffnn.hpp
```diff
@@ -0,0 +1,248 @@
...
+  const Hilbert &hilbert_;
+
+ public:
+  using StateType = typename AbstractMachine<T>::StateType;
+  using LookupType = typename AbstractMachine<T>::LookupType;
+
+  // constructor
+  explicit FFNN(const Hilbert &hilbert, const json &pars)
+      : nv_(hilbert.Size()), hilbert_(hilbert) {
+    from_json(pars);
```
femtobit: Calling virtual functions from a constructor should be avoided (see https://google.github.io/styleguide/cppguide.html#Doing_Work_in_Constructors and https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rc-ctor-virtual for the reasons).

You could use a factory function instead (i.e., a static function initializing the Machine and then returning it).

NetKet/Machine/ffnn.hpp
femtobit: You can use FieldVal, which will provide a better error message if "Layers" is missing.
https://github.com/netket/netket/blob/66c24d47702bc0610e4fc92d2262475fcb03810a/NetKet/Utils/json_helper.hpp#L59

NetKet/Machine/activations.hpp
```diff
@@ -0,0 +1,73 @@
...
+
+  // Apply the Jacobian matrix J to a vector f
+  // J = d_a / d_z = I
+  // g = J * f = f
+  // Z = [z1, ..., zn], G = [g1, ..., gn], F = [f1, ..., fn]
+  // Note: When entering this function, Z and G may point to the same matrix
+  inline void apply_jacobian(const VectorType &Z, const VectorType &A,
+                             const VectorType &F, VectorType &G) {
+    G.noalias() = F;
+    (void)A;
```
femtobit: These lines are meant to suppress warnings about unused variables, right? I suggest not naming the variables in the function signature instead, i.e.
`inline void apply_jacobian(const VectorType &Z, const VectorType & /*A*/, const VectorType &F, VectorType & /*G*/)`
This is a somewhat common style in C++ (the commented name are just there to remind readers what the left-out variables where for).

Cf. https://google.github.io/styleguide/cppguide.html#Function_Declarations_and_Definitions

NetKet/Machine/activations.hpp
femtobit: Could you add a comment describing the purposes of the variables (`Z`, `A`, `F`, `G`) here? I think that would be helpful.

NetKet/Machine/fullconn_layer.hpp
```diff
@@ -0,0 +1,246 @@
...
+        usebias_(use_bias),
+        in_size_(input_size),
+        out_size_(output_size) {
+    Init();
+  }
+
+  explicit FullyConnected(const json &pars) : activation_() {
+    if (FieldExists(pars, "Inputs")) {
```
femtobit: Isn't this equivalent to something like
```
in_size_ = FieldVal(pars, "Inputs");
out_size_ = FieldVal(pars, "Outputs");
```
?

NetKet/Machine/ffnn.hpp
```diff
@@ -0,0 +1,248 @@
...
+#define NETKET_FFNN_HPP
+
+namespace netket {
+
+template <typename T>
+class FFNN : public AbstractMachine<T> {
+  using VectorType = typename AbstractMachine<T>::VectorType;
+  using MatrixType = typename AbstractMachine<T>::MatrixType;
+
+  std::vector<std::unique_ptr<AbstractLayer<T>>>
```
femtobit: In `layer.hpp` you add a typedef `using Ptype = unique_ptr<AbstractLayer<T>>`. Maybe do it here as well.

NetKet/Machine/abstract_layer.hpp
```diff
@@ -0,0 +1,81 @@
...
+#include <netket.hpp>
+#include <random>
+#include <vector>
+
+namespace netket {
+/**
+  Abstract class for Neural Network layer.
+  This class prototypes the methods needed
+  by a class satisfying the Layer concept.
+*/
```
femtobit: The second sentence seems redundant to me.

NetKet/Machine/activations.hpp
kchoo1118: Good point

NetKet/Machine/activations.hpp
kchoo1118: Good suggestion. I will make the changes

NetKet/Machine/fullconn_layer.hpp
kchoo1118: Yes

NetKet/Machine/activations.hpp
kchoo1118: Ok I will add comments

NetKet/Machine/ffnn.hpp
kchoo1118: Ok

NetKet/Machine/ffnn.hpp
gcarleo: I agree, but personally I find factory logic a bit too involved... I wonder if there is some easier alternative

femtobit: The case where problems with virtual function calls from the constructor could occur is when a subclass of `FFNN` overrides `from_json`. (This constructor would still call `FFNN::from_json`,  not the version in the subclass. That might even be the intended behaviour but can still be confusing.)

If `FFNN` is never subclassed, this is not a problem, so I think marking the whole class as `final` would be an alternative to changing the code here, i.e.,
```cpp
class FFNN final : public AbstractMachine<T> {
```
(cf. the second link in my comment above).

Generally, I think this discussion is a symptom of the fact that in the current design we are strongly coupling the initialization of our classes to the JSON representation which tends to lead to difficulties in terms of code design. However, this is not specific to this PR and changing things in this regard should be done consistently for all of netket, so I'd say this is a topic for another time.

NetKet/Machine/fullconn_layer.hpp
femtobit: Writing `activation_()` with an empty parameter list [will lead to the member being default-constructed](https://stackoverflow.com/questions/11164394/constructor-initialization-list-with-empty-initialization), which in this case [is equivalent](https://en.cppreference.com/w/cpp/language/data_members#Member_initialization) to not including `activation_` in the initializer list.

Whether to include it anyway is a matter of style, I guess.

NetKet/Machine/activations.hpp
```diff
@@ -0,0 +1,86 @@
...
+namespace netket {
+
+/**
+  Abstract class for Activations.
+*/
+class AbstractActivation {
+ public:
+  using VectorType = Eigen::Matrix<std::complex<double>, Eigen::Dynamic, 1>;
+
+  virtual inline void Activate(const VectorType &Z, VectorType &A) = 0;
```
femtobit: Another minor point: Naming this function `Activate` seems a bit unusual to me, it's like saying "we activate the activation" instead of "we apply the activation to this vector". I'd therefore suggest calling it `Apply` (which is also consistent with `ApplyJacobian`) (or, alternatively, `operator()` so that the activation can be called like a function).

NetKet/Machine/machine.hpp
```diff
@@ -77,6 +78,8 @@ class Machine : public AbstractMachine<T> {
       m_ = Ptype(new RbmSpin<T>(hilbert, pars));
     } else if (pars["Machine"]["Name"] == "RbmMultival") {
       m_ = Ptype(new RbmMultival<T>(hilbert, pars));
+    } else if (pars["Machine"]["Name"] == "FFNN") {
+      m_ = Ptype(new FFNN<T>(hilbert, pars));
```
femtobit: Also minor: If "Machine.Name" is neither of those values then `m_` will not be set which is an error. This should not happen in practise (as `CheckInput` should be called before passing the input to this function) but I'd add an `assert(CheckInput(pars));` at the start of this function to make sure this never happens (in release mode, the assert will not be compiled so there is no performance problem from checking the input multiple times there).

Tutorials/Heisenberg1d/heisenberg1d.py
```diff
@@ -35,7 +35,7 @@
 #defining the wave function
 pars['Machine']={
     'Name'           : 'RbmSpinSymm',
-    'Alpha'          : 1.0,
+    'Alpha'          : 1,
```
femtobit: Is there a reason for this change?

NetKet/Machine/activations.hpp
gcarleo: I prefer the `operator()` option, maybe returning `A` instead of passing it as a reference (hoping that move constructor is working properly). Otherwise just operator with two arguments is also good.

Tutorials/Heisenberg1d/heisenberg1d.py
gcarleo: Alpha is supposed to be an integer in that case, I guess the change makes sense

femtobit: Ah, okay.

carrasqu:
I don't know how to use this

PR #39, opened 2018-07-03
Adding GraphHamiltonians.

jamesETsmith:
- Added EdgeColors to custom graphs that allow users to specify colors of the edges.

- Added a new class of Hamiltonians called GraphHamiltonians that allow users to specify operators that only act on edges of a certain color. Users can also specify site operators that act on the vertices of the graph.

- Created tutorials for the transverse field Ising and J1 J2 models using graph hamiltonians.

- Added unittest for graph Hamiltonian.


gcarleo:
Thanks @jamesETsmith for the PR. Could you please fix the small issues found by Codacy? I will now have a look at the rest in more detail

gcarleo:
Great! Another small thing (sorry, it's in an effort to make the code consistent) : 

This PR introduces several re-formatting of the code which don't seem in line with the Coding style in [here](https://github.com/netket/netket/blob/master/CONTRIBUTING.md). We should maybe set up a server to run clang-format automatically at every PR, but in the meanwhile could you run it on your files locally? Notice that if you use the 'atom' editor, you can just have this [plugin](https://github.com/LiquidHelium/atom-clang-format ) do the job for you (just remember to set `google` in the style settings)

NetKet/Graph/hypercube.hpp
```diff
@@ -44,6 +44,9 @@ class Hypercube : public AbstractGraph {
   // adjacency list
   std::vector<std::vector<int>> adjlist_;
 
+  // Edge colors // TODO
```
gcarleo: What is the default behaviour here? Are the colours all the same by default?

Tutorials/CustomHamiltonian/custom_hamiltonian.py
```diff
@@ -36,6 +35,9 @@
     operators.append(mszsz)
     sites.append([i, (i + 1) % L])
 
+print(len(sites))
```
gcarleo: Why?

NetKet/Hamiltonian/graph_hamiltonian.hpp
```diff
@@ -0,0 +1,120 @@
...
+namespace netket {
+
+// BondHamiltonian on an arbitrary graph
+template <class G>
+class GraphHamiltonian : public AbstractHamiltonian {
+  std::vector<LocalOperator> operators_;
+  Hilbert hilbert_;
+
+  // Arbitrary graph
+  const G &graph_;
```
twesterhout: Why a reference? I would suggest you use a pointer instead as this would make `GraphHamiltonian` copy and move assignable (which is always a nice thing to have).

NetKet/Hamiltonian/graph_hamiltonian.hpp
```diff
@@ -0,0 +1,120 @@
...
+    if (bop.size() != op_color.size()) {
+      throw InvalidInputError(
+          "The bond Hamiltonian definition is inconsistent."
+          "The sizes of BondOps and BondOpColors do not match.");
+    }
+
+    if (bop.size() > 0) {
+      // Get adjacency list from graph
+      // auto adj = graph_.AdjacencyList();
+      auto ec = graph_.EdgeColors();
```
twesterhout: Do you really want a copy here?

NetKet/Hamiltonian/graph_hamiltonian.hpp
```diff
@@ -0,0 +1,120 @@
...
+
+      // Use adj to populate operators
+      for (std::map<std::vector<int>, int>::iterator it = ec.begin();
```
twesterhout: We're in C++11 land, so how about `for (auto const& p : ec)` instead?

NetKet/Hamiltonian/hamiltonian.hpp
```diff
@@ -41,6 +42,8 @@ class Hamiltonian : public AbstractHamiltonian {
         h_ = std::make_shared<Heisenberg<Graph>>(graph, pars);
       } else if (pars["Hamiltonian"]["Name"] == "BoseHubbard") {
         h_ = std::make_shared<BoseHubbard<Graph>>(graph, pars);
+      } else if (pars["Hamiltonian"]["Name"] == "Graph") {
+        h_ = std::make_shared<GraphHamiltonian<Graph>>(graph, pars);
```
twesterhout: I've only now noticed it, but why do we store a `shared_ptr`?

NetKet/Graph/custom_graph.hpp
```diff
@@ -56,6 +58,16 @@ class CustomGraph : public AbstractGraph {
         std::vector<std::vector<int>> edges =
             pars["Graph"]["Edges"].get<std::vector<std::vector<int>>>();
         AdjacencyListFromEdges(edges);
+
+        // Specify edge colors if there is more than one
+        if (FieldExists(pars["Graph"], "EdgeColors")) {
```
gcarleo: We enter here only if "Edges" are defined in the input. What about if AdjacencyList is defined instead? I guess we should call `EdgeColorsFromList` in that case too

Tutorials/GraphHamiltonian/Ising/ising.py
```diff
@@ -0,0 +1,96 @@
...
+
+pars['Graph'] = {
+    'Edges': list(G.edges),
+    # 'EdgeColors': edge_colors,
+}
+
+# first we choose a hilbert space for our custom hamiltonian
+pars['Hilbert'] = {
+    'QuantumNumbers': [1, -1],
+    'Size': len(list(G.edges)),  # TODO Reduce redundant input
```
femtobit: Is this `TODO` meant to be left here for this PR?

Tutorials/J1J2/j1j2.py
```diff
@@ -12,19 +12,18 @@
 # See the License for the specific language governing permissions and
```
femtobit: Only formatting changes in this file.

Tutorials/J1J2/plot_j1j2.py
```diff
@@ -4,50 +4,56 @@
```
femtobit: Only formatting changes in this file.

Test/Hamiltonian/unit-hamiltonian.cc
```diff
@@ -12,26 +12,23 @@
 // See the License for the specific language governing permissions and
```
femtobit: Only formatting changes in this file

NetKet/Graph/abstract_graph.hpp
```diff
@@ -49,6 +49,12 @@ class AbstractGraph {
   */
   virtual std::vector<std::vector<int>> SymmetryTable() const = 0;
 
+  /**
+  Member function returning edge colors of the graph.
+  @return ec[i][j] is the color of the edge between nodes i and j.
+  */
+  virtual std::map<std::vector<int>, int> EdgeColors() const = 0;
```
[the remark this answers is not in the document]
jamesETsmith: I've switched to using arrays to represent edges, but when I tried to use unordered_map I run into issues because there is no default hash function when arrays are the key (https://stackoverflow.com/questions/42701688/using-an-unordered-map-with-arrays-as-keys). Stack overflow suggests that one could write their own hash function. I am happy to do this, but is it worth the O(1) look-ups of the edges?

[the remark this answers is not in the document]
twesterhout: What's wrong with a custom hash function? I think something like this should do
```cpp
// Warning: untested code :)
struct edge_hasher {
    auto operator()(std::array<int, 2> const x) noexcept -> std::size_t
    {
        static_assert(2 * sizeof(int) == sizeof(std::size_t),
            "What kind of strange system is this?");
        return *reinterpret_cast<std::size_t const*>(x.data());
    }
};
```
This will be extremely fast as well (it's one of the reasons I suggested using `std::array<int, 2>`).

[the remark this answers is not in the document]
femtobit: By the way: Could we add `using Edge = std::array<int, 2>` somewhere instead of spelling out the type everywhere?

NetKet/Hamiltonian/graph_hamiltonian.hpp
jamesETsmith: I used a reference to be consistent with how graphs are stored in the BosonHubbard, Heisenberg, and Ising hamiltonians. If it's something that we want, should I also change it for those Hamiltonians?

gcarleo: Reference is fine

NetKet/Hamiltonian/hamiltonian.hpp
gcarleo: I think there was a problem with using 'unique_ptr' here, but I don't remember exactly what was the problem

jamesETsmith:
@twesterhout : Updated the eclist_ to be an unordered_map<array<int,2>, int>. Removed the TODOs that I'd left in the last PR.

@femtobit : Created a type alias for the unordered_map<array<int,2>, int> to make it less cumbersome, called ColorMap. I have also reverted the files where only the formatting has changed. 

@gcarleo : The default behavior for eclist_ (for all graph types) is to set all edge colors to 0. Custom edge colors can now be specified as a list of lists and will populate eclist_ regardless of whether the user creates a graph from edges or an adjacency list. Creating this input in python is straightforward and I've demonstrated it in Tutorials/GraphHamiltonian/J1J2/j1j2.py. I will try to update my PR for the netket-website shortly to include documentation for this!

Thanks for all your comments, if I missed anything, just let me know!

NetKet/Graph/abstract_graph.hpp
```diff
@@ -49,11 +67,17 @@ class AbstractGraph {
   */
   virtual std::vector<std::vector<int>> SymmetryTable() const = 0;
 
+  /**
+  Custom type for unordered_map<array<int,2>, int> w/ a custom hash function
+  */
+  using ColorMap =
+      std::unordered_map<std::array<int, 2>, int, std::ArrayHasher>;
```
femtobit: Maybe even add another alias for the edge? (Then the reader can be sure immediately what the `std::array<int, 2>` is for.)
```cpp
using Edge = std::array<int, 2>;
using ColorMap = std::unordered_map<Edge, int, std::ArrayHasher>;
```

NetKet/Graph/abstract_graph.hpp
```diff
@@ -15,9 +15,27 @@
...
 #include <vector>
 
+// Special hash functor for the EdgeColors unordered_map
+// Same as hash_combine from boost
+namespace std {
+struct ArrayHasher {
+  std::size_t operator()(const std::array<int, 2>& a) const {
+    std::size_t h = 0;
+
+    for (auto e : a) {
```
twesterhout: Why so difficult? This is the way you'd implement a hash function for `std::array<int, N>`, but `std::array<int, 2>` is just a piece of memory 8 bytes long, and a hash is 8 bytes long, so I don't see why not simply cast one into the other?

NetKet/Graph/abstract_graph.hpp
```diff
@@ -53,7 +71,9 @@ class AbstractGraph {
   Member function returning edge colors of the graph.
   @return ec[i][j] is the color of the edge between nodes i and j.
   */
-  virtual std::map<std::vector<int>, int> EdgeColors() const = 0;
+  using ColorMap =
+      std::unordered_map<std::array<int, 2>, int, std::ArrayHasher>;
+  virtual ColorMap EdgeColors() const = 0;
```
twesterhout: You're still copying the whole map, is there a reason for that?

NetKet/Graph/hypercube.hpp
```diff
@@ -113,6 +127,24 @@ class Hypercube : public AbstractGraph {
     }
   }
 
+  // Edge Colors from users specified map
+  void EdgeColorsFromList(const std::vector<std::vector<int>> &colorlist,
```
twesterhout: It looks like `EdgeColorsFromList` and `EdgeColorsFromAdj` are copy-pasted around (or am I missing something?). Wouldn't it be wiser to move the implementation to a common base class?

NetKet/Graph/abstract_graph.hpp
```diff
@@ -15,9 +15,27 @@
 #ifndef NETKET_ABSTRACTGRAPH_HPP
 #define NETKET_ABSTRACTGRAPH_HPP
 
+#include <array>
+#include <unordered_map>
 #include <vector>
 
+// Special hash functor for the EdgeColors unordered_map
+// Same as hash_combine from boost
+namespace std {
```
twesterhout: Sorry for being bold, but `namespace std` here is a bad idea. It's even considered undefined behavior by the C++ standard. I'd suggest to simply define the hash function inside the NetKet's namespace.

NetKet/Graph/abstract_graph.hpp
```diff
@@ -49,11 +67,17 @@ class AbstractGraph {
...
+  */
+  using ColorMap =
+      std::unordered_map<std::array<int, 2>, int, std::ArrayHasher>;
+
   /**
   Member function returning edge colors of the graph.
   @return ec[i][j] is the color of the edge between nodes i and j.
   */
-  virtual std::map<std::vector<int>, int> EdgeColors() const = 0;
+  virtual ColorMap EdgeColors() const = 0;
```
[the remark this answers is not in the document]
jamesETsmith: Were you thinking something like this: `const virtual ColorMap& EdgeColors() const = 0;`?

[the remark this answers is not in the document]
femtobit: Yes, exactly.

NetKet/Graph/abstract_graph.hpp
femtobit: Indeed, extending `std` is undefined behaviour except for a few cases. See e.g. https://en.cppreference.com/w/cpp/language/extending_std.

NetKet/Graph/hypercube.hpp
jamesETsmith: They are in two locations. I considered putting them in a base class, but I didn't want to clutter the AbstractGraph class and abstract class headers all contain only virtual functions. Since the same is done with the distances function, should I just leave these?

NetKet/Graph/abstract_graph.hpp
jamesETsmith: Thanks, that's good to know, I will move it to `namespace netket`.

NetKet/Graph/abstract_graph.hpp
jamesETsmith: I see now, I was making things more difficult than necessary. I've modified your initial suggestion so the hash functor is now 
```cpp
struct ArrayHasher {
  std::size_t operator()(const std::array<int, 2>& a) const {
    return *reinterpret_cast<std::size_t const*>(a.data());
  }
};
```

twesterhout: You might want to plug `noexcept` in there, but otherwise it looks good to me.

gcarleo: I guess `ArrayHasher` is not specific of `Graph`, and it is better moved in the `Utils/`. Maybe a new file `hashing.hpp` or something similar?

NetKet/Graph/graph.hpp
```diff
@@ -70,6 +68,8 @@ class Graph : public AbstractGraph {
     return g_->Distances();
   }
 
+  virtual ColorMap EdgeColors() const override { return g_->EdgeColors(); }
```
gcarleo: Why this method is virtual?

NetKet/Graph/hypercube.hpp
gcarleo: I think that it is OK to put non-virtual methods in the `Graph` class

NetKet/Hamiltonian/hamiltonian.hpp
gcarleo: We are working on this in another PR, do not worry about this here

Tutorials/GraphHamiltonian/J1J2/j1j2.py
```diff
@@ -0,0 +1,114 @@
...
+    'Name': 'Spin',
+    'S': 0.5,
+    'TotalSz': 0,
+    'Nspins': L,
+}
+
+#defining our custom hamiltonian
+pars['Hamiltonian'] = {
+    'Name': 'Graph',
+    'SiteOps': [],
```
gcarleo: Maybe we should just check that at least one among `SiteOps` and `BondOps` is defined in the constructor of  `GraphHamiltonian`. In this way, we can avoid to explicitly write 

````python
'SiteOps': [],
````

or similarly when there are only `SiteOps` but no `BondOps` (boring hamiltonian in that case though!)

File: NetKet/Graph/abstract_graph.hpp (hunk 14 above)
Comment (jamesETsmith) [reply NetKet/Graph/abstract_graph.hpp]: Sure, I'll create the `array_hasher.hpp` file in `Utils/` with the noexcept.

File: Tutorials/GraphHamiltonian/J1J2/j1j2.py (hunk 20 above)
Comment (jamesETsmith) [reply Tutorials/GraphHamiltonian/J1J2/j1j2.py]: Great point, I added some checks in `GraphHamiltonian` that create empty arrays for SiteOps or BondOps if they aren't specified (and check to make sure that at least one is specified). I've also added the behavior where users aren't required to specify `BondOpColors` if they are all the same color.

File: NetKet/Graph/graph.hpp (hunk 19 above)
Comment (jamesETsmith) [reply NetKet/Graph/graph.hpp]: This was an oversight on my part and I've removed the virtual keyword.

File: Tutorials/GraphHamiltonian/J1J2/j1j2.py [hunk 21]
```diff
@@ -0,0 +1,113 @@
...
+
+bond_color = [1, 2, 1, 2]
+
+# Define custom graph
+G = nx.Graph()
+for i in range(L):
+    G.add_edge(i, (i + 1) % L, color=1)
+    G.add_edge(i, (i + 2) % L, color=2)
+
+    G.add_edge((i + 1) % L, i, color=1)
```
gcarleo: Inserting twice the same edges? I guess the energy should be twice as large then?
I would insert bonds only once, to be consistent with the other hamiltonians

jamesETsmith: Sorry this was left from a test when I initialized the graph as a DiGraph and had to specify each edge in both directions. I will remove it.

gcarleo:
@femtobit , @twesterhout , are there some other changes to be considered in this PR? Thanks!

femtobit:
@gcarleo I think all my remarks have been addressed, so the PR looks good to me now.

twesterhout:
I'm okay with merging the code in its current state. Well done @jamesETsmith!

jamesETsmith:
Thanks for all the help!

PR #41, opened 2018-07-05
discard initial samples

wuyukai:
Add some initial sweeps to reach equilibrium. The number of sweeps is from "DiscardedSamples" in "Learning". By default it is 10% of "NSamples"

### DIFF
--- a/NetKet/Learning/ground_state.hpp
+++ b/NetKet/Learning/ground_state.hpp
@@ -116,9 +116,9 @@ public:
 
       setSrParameters(diagshift, rescale_shift, use_iterative);
     }
-    
-    double nskipsamples =
-          FieldOrDefaultVal(pars["Learning"], "SkipInitSamples", 0.);
+
+    double ndiscardedsamples =
+        FieldOrDefaultVal(pars["Learning"], "DiscardedSamples", 0.1 * nsamples);
 
     if (dosr_) {
       InfoMessage() << "Using the Stochastic reconfiguration method"
@@ -130,7 +130,7 @@ public:
       InfoMessage() << "Using a gradient-descent based method" << std::endl;
     }
 
-    Run(nskipsamples, nsamples, niter_opt);
+    Run(ndiscardedsamples, nsamples, niter_opt);
   }
 
   void Init() {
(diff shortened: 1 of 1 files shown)


NetKet/Learning/ground_state.hpp
```diff
@@ -153,6 +156,13 @@ class GroundState {
...
   }
 
+  void InitSweeps(double ndiscardedsamples) {
+    sampler_.Reset();
+
+    for (int i = 0; i < ndiscardedsamples; i++)
+      sampler_.Sweep();
+  }
+
   void Sample(double nsweeps) {
```
gcarleo: Each time `Sample` is called we should discard the initial samples, not only at the beginning of the optimisation.

wuyukai: Ideally it should be done at each iteration. But I think maybe it is not necessary, because at each iteration the parameters are updated only slightly, so that the equilibrium configuration at the last iteration will be close to the equilibrium for the current one. If we are to discard initial samples at each iteration, it may significantly slow down the program when multiple processes are used, because even though we discard only 1/10 of the samples, it needs to be done on all the processes; then it can take even longer time than the useful samples.

gcarleo: Yes but the changes in the variational parameters could be large, and discarding a 10% of the samples is not a big deal anyway, also because in that part we don't want to measure the local energy and do other more expensive operations that are required in the later sweeps. I think that skipping samples only in the first step of the optimization is not a clean enough solution, we should definitely do the skipping every time the variational parameters are changed. 

Basically in `Sample`, just add a loop 

```c++
  for (int i = 0; i < ndiscardedsamples; i++) {
      sampler_.Sweep();
    }
```

also, I would say that the number of discarded samples can be a fraction of the number of samples per processor `sweepnode`, not of the total number of samples. Not ideal, but better than having a process run a whole simulation only on discarded samples.

gcarleo: If you want an even finer tuning of these parameters, you can actually define a second input, something like `DiscardedSamplesOnInit`, which takes care only of the discarded samples in the initial iteration of the optimizer. By default, I would suggest to have this parameter equal to 0, so that those sweeps are performed **in addition** to the `DiscardedSamples` happening anyway at each iteration of the optimizer.

PR #46, opened 2018-07-13
Feature/feedforward

kchoo1118:


NetKet/Machine/machine.hpp
```diff
@@ -41,21 +42,18 @@ class Machine : public AbstractMachine<T> {
 
   explicit Machine(const Hilbert &hilbert, const json &pars)
       : hilbert_(hilbert) {
-    CheckInput(pars);
     Init(hilbert_, pars);
```
gcarleo: There should be a CheckInput(pars); for all the constructors

NetKet/Machine/machine.hpp
```diff
@@ -117,7 +118,8 @@ class Machine : public AbstractMachine<T> {
     CheckFieldExists(pars, "Machine");
     const std::string name = FieldVal(pars["Machine"], "Name", "Machine");
 
-    std::set<std::string> machines = {"RbmSpin", "RbmSpinSymm", "RbmMultival"};
+    std::set<std::string> machines = {"RbmSpin", "RbmSpinSymm", "RbmMultival",
```
gcarleo: URbm, not yet

NetKet/Machine/sum_output.hpp
```diff
@@ -0,0 +1,119 @@
...
+
+  // Using lookup
+  void Forward(const VectorType &prev_layer_data,
+               const LookupType & /*lt*/) override {
+    z_(0) = prev_layer_data.sum();
+  }
+
+  VectorType Output() const override { return z_; }
+
+  void Backprop(const VectorType & /*prev_layer_data*/,
```
gcarleo: Backprop is not implemented in this case? I know it does not make sense to backprob the final layer, but what if somebody wants this layer in the middle of a network?

NetKet/Machine/sum_output.hpp
```diff
@@ -0,0 +1,119 @@
...
+#include <random>
+#include <vector>
+#include "Lookup/lookup.hpp"
+#include "Utils/all_utils.hpp"
+#include "abstract_layer.hpp"
+
+namespace netket {
+
+template <typename T>
+class SumOutput : public AbstractLayer<T> {
```
gcarleo: Maybe some pooling layer would be a more standard name?

NetKet/Machine/sum_output.hpp
kchoo1118: It would still work. There is no parameters in this layer so backprop doesnt need to do anything. It should still work if you put it in the middle.

NetKet/Machine/machine.hpp
kchoo1118: I remembered we moved the CheckInput into Init.

NetKet/Learning/ground_state.hpp
```diff
@@ -120,6 +120,7 @@ class GroundState {
         FieldVal(pars["Learning"], "OutputFile", "Learning");
 
     int freqbackup = FieldOrDefaultVal(pars["Learning"], "SaveEvery", 50);
+
```
femtobit: This file only has formatting changes.

NetKet/Machine/abstract_layer.hpp
```diff
@@ -0,0 +1,96 @@
...
+  Machine: L = log(psi(v))
+  */
+  virtual void Backprop(const VectorType &prev_layer_data,
+                        const VectorType &next_layer_data, VectorType &der,
+                        int start_idx) = 0;
+  /**
+  Member function to return dL/d(in), where (in) is the input to the current
+  layer, and L = log(psi(v))
+  */
+  virtual const VectorType &BackpropData() const = 0;
```
femtobit: Is there a better name for this function? `BackprobData` does not seem very descriptive to me (`Data` is not very specific).

NetKet/Machine/abstract_machine.hpp
```diff
@@ -18,9 +18,9 @@
 #include <Eigen/Dense>
```
femtobit: This file only has formatting changes.

NetKet/Machine/activations.hpp
```diff
@@ -0,0 +1,192 @@
...
+  // G = J * F = F
+  inline void ApplyJacobian(const VectorType & /*Z*/, const VectorType & /*A*/,
+                            const VectorType &F, VectorType &G) {
+    G.noalias() = F;
+  }
+};
+
+class Lncosh : public AbstractActivation {
+ private:
+  using VectorType = Eigen::Matrix<std::complex<double>, Eigen::Dynamic, 1>;
```
femtobit: This is the same as `AbstractActivation::VectorType`. Is there a need to define it here again? (Would the code break if this was different from `AbstractActivation::VectorType`?)

Same comment for the other subclasses where this is also defined again.

NetKet/Machine/activations.hpp
```diff
@@ -0,0 +1,192 @@
...
+
+ public:
+  // A = Lncosh(Z)
+  inline void operator()(const VectorType &Z, VectorType &A) {
```
femtobit: You should add the [`override` specifier](https://en.cppreference.com/w/cpp/language/override) where applicable (i.e., `operator()` and `ApplyJacobian` here (and also for the other subclasses).

NetKet/Machine/activations.hpp
```diff
@@ -0,0 +1,192 @@
...
+  using VectorType = typename AbstractActivation::VectorType;
+
+  explicit Activation(const json &pars) { Init(pars); }
+  void Init(const json &pars) {
+    CheckInput(pars);
+    MPI_Comm_rank(MPI_COMM_WORLD, &mynode_);
+    if (pars["Activation"] == "Lncosh") {
+      m_ = Ptype(new Lncosh());
+      if (mynode_ == 0) {
+        std::cout << "# # Activation: "
```
femtobit: Use `InfoMessage` (or maybe `DebugMessage`, as this output is probably only useful for debugging) instead of `std::cout` here.

NetKet/Machine/ffnn.hpp
```diff
@@ -0,0 +1,267 @@
...
+  void Init(const json &pars) {
+    json layers_par;
+    if (FieldExists(pars["Machine"], "Layers")) {
+      layers_par = pars["Machine"]["Layers"];
+      nlayer_ = layers_par.size();
+    } else {
+      throw InvalidInputError("Field (Layers) not defined for Machine (FFNN)");
+    }
+
+    std::string buffer = "";
```
femtobit: What is the purpose of buffer here? As far as I can see it is always `""` and only passed as the `comment` argument to `InfoMessage`.

NetKet/Machine/fullconn_layer.hpp
```diff
@@ -0,0 +1,257 @@
...
+    z_.resize(out_size_);
+    a_.resize(out_size_);
+
+    if (usebias_) {
+      npar_ += out_size_;
+    } else {
+      bias_.setZero();
+    }
+
+    std::string buffer = "";
```
femtobit: See question on `buffer` above.

NetKet/Machine/fullconn_layer.hpp
```diff
@@ -0,0 +1,257 @@
...
+
+  int Ninput() const override { return in_size_; }
+
+  int Noutput() const override { return out_size_; }
+
+  void GetParameters(VectorType &pars, int start_idx) override {
+    int k = start_idx;
+
+    if (usebias_) {
+      std::memcpy(pars.data() + k, bias_.data(), out_size_ * scalar_bytesize_);
```
femtobit: Is this `memcpy` for efficiency? This can probably written in a safer way using Eigen methods. Are you sure it is necessary to sacrifice this safety by copying the raw memory?

NetKet/Machine/layer.hpp
```diff
@@ -0,0 +1,130 @@
...
+  using MatrixType = typename AbstractMachine<T>::MatrixType;
+  using StateType = typename AbstractMachine<T>::StateType;
+  using LookupType = typename AbstractMachine<T>::LookupType;
+
+  explicit Layer(const Graph &graph, const json &pars) { Init(graph, pars); }
+
+  void Init(const Graph &graph, const json &pars) {
+    CheckInput(pars);
+    if (pars["Name"] == "FullyConnected") {
+      if (pars["Activation"] == "Lncosh") {
```
femtobit: To improve readability: Maybe extract the value once as a variable and use that in every if statement (i.e., `const std::string activation = pars["Activation"]`).

NetKet/Machine/sum_output.hpp
kchoo1118: Ok sorry. It didnt work before but now it should be fine.

femtobit: Could you add test cases to check the back (and forward) propagation for all the layer types. If we have tests that catch errors such as the one here we could be a lot more confident that the implementation is correct.

NetKet/Machine/conv_layer.hpp
```diff
@@ -0,0 +1,435 @@
...
+#include <random>
+#include <vector>
+#include "Lookup/lookup.hpp"
+#include "Utils/all_utils.hpp"
+#include "abstract_layer.hpp"
+
+namespace netket {
+/** Convolutional layer with spin 1/2 hidden units.
+ Important: In order for this to work correctly, VectorType and MatrixType must
+ be column major.
```
femtobit: Where does this restriction come from?

In any case we should make sure this restriction is enforced (otherwise we have a source of bugs). I'm not an expert on template programming, but the storage order [is a template parameter](https://eigen.tuxfamily.org/dox/group__TopicStorageOrders.html) of the `Eigen::Matrix` class, so I think it should be possible to check this at compile time.

NetKet/Machine/conv_layer.hpp
```diff
@@ -0,0 +1,435 @@
...
+
+    Init(graph);
+  }
+
+  void Init(const Graph &graph) {
+    scalar_bytesize_ = sizeof(std::complex<double>);
+    // Construct neighbours_ kernel(k) will act on neighbours_[i][k]
+    std::vector<std::vector<int>> adjlist;
+    adjlist = graph.AdjacencyList();
+    for (int i = 0; i < nv_; ++i) {
```
femtobit: What exactly is the purpose of these nested loops?

carrasqu:
Everything looks fine to me. My understanding is that im2col is the best you can do in the general case, so I would say the Conv layer is implemented is optimal. One thought I had is that in the future it may be worth exploring automatic differentiation to save energy and avoid bugs in the backprop steps but this is not specific to conv layers but a more general issue.

NetKet/Machine/activations.hpp
kchoo1118: Sure I have changed it.

NetKet/Machine/layer.hpp
kchoo1118: But not all layers need an activation

NetKet/Machine/ffnn.hpp
kchoo1118: Yes because I didnt want the "#  "  in the standard messages

NetKet/Machine/conv_layer.hpp
kchoo1118: This restriction comes from the use of eigen::map.

kchoo1118: I added a check

NetKet/Machine/abstract_machine.hpp
gcarleo: Yes but the original file was not formatted correctly, so this is fine

NetKet/Machine/conv_layer.hpp
femtobit: > I added a check

Thanks. I took the liberty to change it to a compile-time check in 67fe7e6 so that the compilation will abort with an error message if the conv layer is used with a row-major matrix.

NetKet/Machine/conv_layer.hpp
```diff
@@ -0,0 +1,434 @@
...
+
+    dist_ = FieldVal(pars, "Distance");
+
+    usebias_ = FieldOrDefaultVal(pars, "UseBias", true);
+
+    Init(graph);
+  }
+
+  void Init(const Graph &graph) {
+    // Construct neighbours_ kernel(k) will act on neighbours_[i][k]
```
femtobit: See #47.

NetKet/Machine/conv_layer.hpp
```diff
@@ -0,0 +1,434 @@
...
+    std::vector<std::vector<int>> adjlist;
+    adjlist = graph.AdjacencyList();
```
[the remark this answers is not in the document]
gcarleo: I agree, this can be especially beneficial for large graphs. But somehow I was hoping that the move constructor would avoid copies

NetKet/Machine/conv_layer.hpp
```diff
@@ -0,0 +1,434 @@
...
+
+  bool usebias_;  // boolean to turn or off bias
+
+  int nv_;            // number of visible units in the full network
+  int in_channels_;   // number of input channels
+  int in_size_;       // input size: should be multiple of no. of sites
+  int out_channels_;  // number of output channels
+  int out_size_;      // output size: should be multiple of no. of sites
+  int npar_;          // number of parameters in layer
+  int kernelpar_;     // number of parameters in layer
```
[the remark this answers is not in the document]
gcarleo: @kchoo1118 ok to remove it?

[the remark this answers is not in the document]
gcarleo: I was wondering why the compiler doesn't pop up a warning for this though. Normally it should, if the variable is unused.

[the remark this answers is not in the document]
femtobit: > I was wondering why the compiler doesn't pop up a warning for this though. Normally it should, if the variable is unused.

Probably because it is assigned to in `Init`. My IDE did complain, though.

NetKet/Machine/conv_layer.hpp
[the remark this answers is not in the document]
femtobit: As far as I can see the adjacency list it is stored as a member variable of the graph. Moving the object out would make that member unusable. Therefore the compiler will not perform copy elision.

NetKet/Machine/conv_layer.hpp
```diff
@@ -46,7 +46,6 @@ class Convolutional : public AbstractLayer<T> {
   int out_channels_;  // number of output channels
   int out_size_;      // output size: should be multiple of no. of sites
   int npar_;          // number of parameters in layer
-  int kernelpar_;     // number of parameters in layer
```
femtobit: This breaks the build because `kernelpar_` is still used below. Fixed in the current version of #47.

NetKet/Machine/fullconn_layer.hpp
```diff
@@ -0,0 +1,257 @@
...
+    // Derivative for weights, d(L) / d(W) = [d(L) / d(z)] * in'
+    Eigen::Map<MatrixType> der_w{der.data() + k, in_size_, out_size_};
+
+    der_w.noalias() = prev_layer_data * dLz.transpose();
+
+    // Compute d(L) / d_in = W * [d(L) / d(z)]
+    din_.noalias() = weight_ * dLz;
+  }
+
+  const VectorType &BackpropData() const override { return din_; }
```
gcarleo: I still think that we should not have this BackpropData, and that din_ should just be returned by the Backprop method. Is there some case where this wouldn't work?

NetKet/Machine/fullconn_layer.hpp
```diff
@@ -0,0 +1,257 @@
...
+
+  // Using lookup
+  void Forward(const VectorType & /*prev_layer_data*/,
+               const LookupType &lt) override {
+    z_ = lt.V(0);
+    // Apply activation function
+    activation_(z_, a_);
+  }
+
+  VectorType Output() const override { return a_; }
```
gcarleo: Similarly to my comment on `BackPropData`, it seems to me that this method is redundant, since a_ should just be returned by the two Forward methods. Also, this is potentially error prone, because here one is implicitly assuming that every time one calls `Output` a call to `Forward` has been performed before. What if this is not the case?

NetKet/Machine/abstract_layer.hpp
gcarleo: I suggest to remove `BackpropData`, see my other comments below

NetKet/Machine/abstract_layer.hpp
```diff
@@ -0,0 +1,96 @@
...
+  @param prev_layer_data a constant reference to the output from previous layer.
+  */
+  virtual void Forward(const VectorType &prev_layer_data) = 0;
+
+  virtual void Forward(const VectorType &prev_layer_data,
+                       const LookupType &lt) = 0;
+  /**
+  Member function to return output of forward propagation.
+  */
+  virtual VectorType Output() const = 0;
```
gcarleo: Better remove this method, see below.

NetKet/Machine/fullconn_layer.hpp
kchoo1118: then i would need to store the output of a each layer in the machine class: FFNN.
You prefer that?

gcarleo: I think that would make more sense, yes (and actually even be slightly faster). We should just be careful when returning those vectors though, I am not sure Eigen supports move semantics. It might be a better option to pass the output by reference to the `Forward` method, though less elegant. Maybe @femtobit or @twesterhout know better.

gcarleo: Actually I am only now realising that returning a_ by value might be **very** bad for performance in the current implementation.

kchoo1118: But I dont really like that the layers dont know its own output

gcarleo: The layer shouldn't necessarily know about the output, since the output depends on the specific input, that somebody else is feeding to the layer. It's the FFNN that knows both input and output... it seems consistent to me. It's like implementing a class that computes sin(x). You have two options, say: 

```c++
class Sin{

public :

double operator() (double x){
   return sin(x); 
}
}; 
````

versus 
```c++
class Sin{
double y_; 
public :

void operator() (double x){
   y_=sin(x); 
}

double val(){
  return y_; 
}
}; 
```
I think the first option makes much more sense, and it's also safer! (imagine that you call the method Value() before calling the operator () in the second case...

kchoo1118: ok i will change it to pass by reference then

kchoo1118: should i also store the value of the output before applying the activation in FFNN?

So that FFNN stores the output, value before activation and din_?

NetKet/Machine/fullconn_layer.hpp
```diff
@@ -0,0 +1,256 @@
...
+    activation_(z_, output);
+  }
+
+  // Using lookup
+  void Forward(const VectorType & /*prev_layer_output*/, const LookupType &lt,
+               VectorType &output) override {
+    z_ = lt.V(0);
+    // Apply activation function
+    output.resize(out_size_);
+    activation_(z_, output);
```
gcarleo: At this point I would just do 
```c++
activation_(lt.V(0),output);
```
maybe resizing first if activation does not resize by default, or not resizing if output is already supposed to be of the good size

NetKet/Machine/fullconn_layer.hpp
```diff
@@ -0,0 +1,256 @@
...
+        const int sf = tochange[s];
+        lt.V(0) += weight_.row(sf) * (newconf[s] - v(sf));
+      }
+    }
+  }
+
+  void Forward(const VectorType &prev_layer_output,
+               VectorType &output) override {
+    // Linear term z = W' * in + b
+    z_ = bias_;
```
gcarleo: Now there is no need to keep this temporary ```z_``` I think.

You can do a one-liner

```c++
return activation_(bias_+weight_.transpose() * prev_layer_output,output);
```

Eigen should be smart enough to do a single Blas call to do the affine transformation optimally.

NetKet/Machine/fullconn_layer.hpp
```diff
@@ -0,0 +1,256 @@
...
+      k += out_size_;
+    }
+
+    // Derivative for weights, d(L) / d(W) = [d(L) / d(z)] * in'
+    Eigen::Map<MatrixType> der_w{der.data() + k, in_size_, out_size_};
+
+    der_w.noalias() = prev_layer_output * dLz.transpose();
+
+    // Compute d(L) / d_in = W * [d(L) / d(z)]
+    din_.noalias() = weight_ * dLz;
```
gcarleo: No need for ```din_```, just pass it by reference and modify it, removing the ```BackpropData``` method

NetKet/Machine/fullconn_layer.hpp
kchoo1118: no i want the temporary z_, for computing derivatives

NetKet/Machine/fullconn_layer.hpp
```diff
@@ -0,0 +1,256 @@
...
+  void Backprop(const VectorType &prev_layer_output,
+                const VectorType &this_layer_output,
+                const VectorType &next_layer_data, VectorType &der,
+                int start_idx) override {
+    // After forward stage, m_z contains z = W' * in + b
+    // Now we need to calculate d(L) / d(z) = [d(a) / d(z)] * [d(L) / d(a)]
+    // d(L) / d(a) is computed in the next layer, contained in next_layer_data
+    // The Jacobian matrix J = d(a) / d(z) is determined by the activation
+    // function
+    VectorType &dLz = z_;
```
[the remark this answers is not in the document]
kchoo1118: yea thats why i am asking if u wanna store the z_ in FFNN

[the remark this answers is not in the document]
gcarleo: But ```z_``` is somehow specific of this architecture, I am not sure it makes sense to store it in FFNN. If it is something necessary to speed up the calculations, and related to the current value of the input, then that is definitely something to be stored in a layer-wise LookupTable.

NetKet/Machine/fullconn_layer.hpp
kchoo1118: actually depending on the activation function, sometimes it is not necessary to use z_

NetKet/Machine/fullconn_layer.hpp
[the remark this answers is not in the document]
kchoo1118: i think it can be quite significant extra time, for example recomputing z_ would correspond to performing the whole convolution again.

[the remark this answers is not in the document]
kchoo1118: But anyway, the backprop is only called from FFNN, so the implementation of the FFNN gurantees that there is always first the forward pass which sets the z_ no?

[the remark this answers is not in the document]
kchoo1118: And isnt a layer-wise lookuptable exactly the same as z_?

gcarleo:
I think it's time to merge this! Thank you very much @kchoo1118 for the work on this important part of the code. Thank you @femtobit also for the long-needed changes to the graph class. I think that we will need to come back to these once the design of the new Lookup type / MonteCarloState will be finalized #52

PR #47, opened 2018-07-22
Simplify graph code in conv layer

femtobit:
As far as I understand, the following part of the code in the `Convolutional` class is supposed to store, for each node `i`, all nodes with distance of at most `dist_` to `i` in `neighbours_[i]`:
https://github.com/netket/netket/blob/67fe7e62c93abf4280d85c16023563868b3e41bd/NetKet/Machine/conv_layer.hpp#L115-L137

I found the logic there a bit hard to read an rewrote it in terms of a [breadth-first search](https://en.wikipedia.org/wiki/Breadth-first_search) started from each node, which I find easier to understand. (It could also be a bit faster, but I did not check this.)
The BFS code can potentially also be useful elsewhere.

If this looks good to you, accepting this PR will merge the changes into the branch for the feedforward PR #46 (so this is a PR on a PR, essentially).

### DIFF
--- a/NetKet/Graph/abstract_graph.hpp
+++ b/NetKet/Graph/abstract_graph.hpp
@@ -22,7 +22,7 @@
 
 namespace netket {
 /**
-    Abstract class for Graphs.
+    Abstract class for undirected Graphs.
     This class prototypes the methods needed
     by a class satisfying the Graph concept.
     These include lattices and non-regular graphs.
@@ -57,6 +57,13 @@ class AbstractGraph {
   @return true if lattice is bipartite.
   */
   virtual bool IsBipartite() const = 0;
+
+  /**
+   * Checks whether the graph is connected, i.e., there exists a path between
+   * every pair of nodes.
+   * @return true, if the graph is connected
+   */
+  virtual bool IsConnected() const = 0;
 
   /**
    * Perform a breadth-first search (BFS) through the graph, calling
--- a/NetKet/Graph/abstract_graph.hpp
+++ b/NetKet/Graph/abstract_graph.hpp
@@ -15,6 +15,9 @@
 #ifndef NETKET_ABSTRACTGRAPH_HPP
 #define NETKET_ABSTRACTGRAPH_HPP
 
+#include <cassert>
+#include <queue>
+#include <utility>
 #include <vector>
 
 namespace netket {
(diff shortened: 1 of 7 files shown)


gcarleo:
I like the idea of putting Graph-specific functions inside the Graph classes, and not in the Machine, thanks a lot for this! More comments soon

NetKet/Machine/conv_layer.hpp
```diff
@@ -111,28 +113,13 @@ class Convolutional : public AbstractLayer<T> {
...
-              }
-            }
-            if (!isin) {
-              neigh.push_back(m);
-            }
-          }
-        }
-      }
+      graph.BreadthFirstSearch(i, dist_, [&neigh](int node, int /*depth*/) {
+        neigh.push_back(node);
```
gcarleo: Nice lambda

Test/Graph/unit-graph.cc
```diff
@@ -16,10 +16,13 @@
 #include "netket.hpp"
 #include <fstream>
 #include <iostream>
+#include <unordered_set>
 #include <vector>
 
 #include "graph_input_tests.hpp"
 
+using namespace netket;
```
gcarleo: Uhm, maybe go back to namespace free?

Test/Graph/graph_input_tests.hpp
```diff
@@ -23,7 +29,7 @@ std::vector<netket::json> GetGraphInputs() {
   // Hypercube 3d
   pars = {
       {"Graph",
-       {{"Name", "Hypercube"}, {"L", 20}, {"Dimension", 3}, {"Pbc", true}}}};
+       {{"Name", "Hypercube"}, {"L", 10}, {"Dimension", 3}, {"Pbc", true}}}};
```
gcarleo: what was wrong with 20? Too slow to perform the BFS? (It shouldn't I guess)

Test/Graph/unit-graph.cc
```diff
@@ -30,9 +33,48 @@ TEST_CASE("graphs have consistent number of sites", "[graph]") {
...
+          INFO("ncall: " << ncall << ", start: " << start << ", v: " << v << ", depth: " << depth);
+          REQUIRE(visited.count(v) == 0);
+          visited.insert(v);
+          ncall++;
+        });
+      }
+    }
+  }
+
+  SECTION("each node is visited at least once on connected graph") {
```
gcarleo: Now that you mention this: we don't have a function to check if a graph is connected or not! That (simple) function would also make this test look nicer in terms of input

Test/Graph/graph_input_tests.hpp
femtobit: Well, with optimizations it was resonably fast, it just took some time in debug mode. But this probably shouldn't be left in for the PR. I can remove it.

Test/Graph/unit-graph.cc
femtobit: True. I can add that in this PR in a moment.

Test/Graph/graph_input_tests.hpp
gcarleo: It's OK, I was just wondering about production! Just leave it to 10, it doesn't make a difference anyway for testing

femtobit: Alright. Also note that for `N=20` in the test BFS started once for every node and then visits every node once which means `8000 * 8000 = 64e6` calls to the `visitor_function` in that test case.

gcarleo:
No directed graphs , we check that the adjacency list is symmetric
somewhere in the constructors (I have to double check that is global
thought and not only for hypercube)

NetKet/Graph/abstract_graph.hpp
```diff
@@ -55,6 +58,69 @@ class AbstractGraph {
...
+    // Pair of node and depth
+    using QueueEntry = std::pair<int, int>;
+
+    assert(start >= 0 && start < Nsites());
+    assert(max_depth > 0);
+
+    const auto adjacency_list = AdjacencyList();
+
+    // Store the already seen sites
+    std::vector<char> seen(Nsites());
```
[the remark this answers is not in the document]
femtobit: I think the `<...>` in you comment got lost.

gcarleo:
bool, vector of bool

kchoo1118:
This looks good thanks alot.
Its much better now

PR #56, opened 2018-08-15
Ground state calculation using imaginary time propagation

femtobit:
This PR adds a work-in-progress implementation of ground state calculations using imaginary time propagation with the Runge-Kutta time-stepping already implemented in the `feature/time_evolution` branch.

(It could probably use some more tests and documentation before merging to master, but I think this is a good time for some initial feedback.)

I have added a tutorial example for the 1d Ising model, which gives the same result for the ground state energy as the RBM version. (Currently the example uses N=12 as larger N were slow. The runtime was dominated however by the construction of the sparse matrix, so that should be solved by PR #54.)


femtobit:
I disagree with the Codacy's complaint here. It recommend that `TimeRange` should be passed by const reference. However, the object in question is just a small struct of three `double`s (i.e., 24 bytes) and would usually be passed as three separate double values (I just grouped them into a struct because they are used together in several places, though using the struct there as well is so far not part of this PR.)

femtobit:
With the speedup from PR #54 the Ising1d example now seems to work fine for `N=20`.

figure_1

gcarleo:
Thanks @femtobit, this is excellent! I am happy that the improvement in the sparse matrix also helped the imaginary-time evolution! I agree that codacy sometimes gives bogus warnings... let me see what I can do. 

I think there is something going wrong with the tests (some wrong field value for the Observable?) but anyway I will give a more in depth review soon. This mpi leaks are SO annoying... I'd be tempted to remove the sanitiser checks but that's a bit of a stretch. Maybe there is a way to silent just those mpi related warnings?

NetKet/UnitaryDynamics/TimeStepper/abstract_time_stepper.hpp
```diff
@@ -31,11 +31,29 @@ template<class State>
...
+     *      The function is called with the current state as const reference, a
+     *      reference to the derivative dxdt which ode_system should write to and
+     *      the current time t.
+     * @param x The value of x at t.
+     * @param t The current time.
+     * @param dt The time step. Time steppers may use a smaller time step internally
+     *      (e.g., as part of an adaptive algorithm).
+     */
     virtual void Propagate(OdeSystemFunction<State> ode_system,
                            State &x,
```
gcarleo: Here `State` here be a general object, right? I am already thinking at the case where we will need this to do the t-VMC and it's the variational parameters that we need to evolve, not the whole wave-function

Test/Hamiltonian/unit-matrixwrapper.cc
```diff
@@ -127,11 +126,11 @@ TEST_CASE("MatrixWrappers compute correct eigenvalues", "[matrix-wrapper]")
     pars["Observables"].push_back(observable_json);
 
     netket::Hilbert hilbert(pars);
-    netket::Observables obs(hilbert, pars);
+    netket::Observable obs(hilbert, pars);
 
     // check whether the correct eigenvalues are computed
     {
-        netket::DenseMatrixWrapper<netket::AbstractObservable> dense(obs(0));
```
gcarleo: Thanks, indeed this was ugly

gcarleo:
Just another comment, maybe `Observable` should still support input from a single observable from Json. I mean, if in the Json file there is only one observable specified, Observable should be constructed without looking for a vector, otherwise it would give an error? It would be enough to make a check in the Json contstructor, something like "if(... IsArray() )" or a similar method they have there, can't remember now

NetKet/GroundState/imaginary_time.hpp
```diff
@@ -0,0 +1,154 @@
...
+  void PrintOutput(int i, double time) {
+    auto jiter = json(obsmanager_);
+    jiter["Iteration"] = i;
+    jiter["Time"] = time;
+    outputjson_["Output"].push_back(jiter);
+
+    if (mpi_rank_ == 0) {
+      if (jiter["Iteration"] != 0) {
+        long pos = filelog_.tellp();
+        filelog_.seekp(pos - 3);
```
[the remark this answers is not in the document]
femtobit: One thing we could do is to just write one JSON object per line, as is done in my proof-of-concept code for the real-time evolution here:

https://github.com/netket/netket/blob/2f86307ab32f51b4397d66dabbdbb6d9b7433f8d/NetKet/UnitaryTimeEvolution/time_evolution.hpp#L113-L115

Though the whole file would not be valid JSON, it is still easy to read line by line, e.g.

```python
with open("...") as f:
    for line in f:
        obj = json.loads(line)
        times.append(obj["Time"])
        ...
```

Currently, the code for writing the output is also duplicated with the VMC class. I think I can extract it into a dedicated class so that it can be reused.

NetKet/UnitaryDynamics/TimeStepper/abstract_time_stepper.hpp
femtobit: In the current implementation, `State` can be an arbitrary `Eigen::Matrix`-like type (which includes the `Vector` classes). If the variational parameters are stored in a different way (like a set of matrices with different shapes) we will need to either put them into a large vector or generalize the ODE implementation.

femtobit:
Specifying a single observable in the input file should still work:
https://github.com/netket/netket/pull/56/commits/85771c33f0e8cab82a5207a63c220e014f96be1c#diff-c3fd182bda1a58c17284d5641b6b86caR56
(I copied that code directly from the `Observables` constructor to `Observable::FromJson`.)

NetKet/GroundState/imaginary_time.hpp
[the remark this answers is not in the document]
gcarleo: Wait wait! I wouldn't throw away JSON validity in this case though. I think there might be ways to do this using the json library, maybe getting a diff between two json objects, or something similar. I will look into this, and see if there is a built-in solution, otherwise we can always ask the developers of nlhoman:json if they know a way to do it (the development of that library is very active)

gcarleo:
AT LAST, I think I solved the MPI leaks issue. Basically, in Travis they have a rather old Ubuntu distribution (Trusty) which comes with a ~4 years old distribution of openmpi. Those old versions are [known](http://users.open-mpi.narkive.com/YHnzgOhU/ompi-users-valgrind-reports-lots-of-memory-leakage) to have memory leaks!!!! I just switched to a newer version, on OsX. The only problem is that it takes a lot of time to compile. I will try to see if we can get a dedicated CI server.

In the meanwhile... why why don't we make also Codacy happy ? ehe

NetKet/GroundState/imaginary_time.hpp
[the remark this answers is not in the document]
femtobit: This is solved, I think, the new `JsonOutputWriter` class I just added to this PR. It writes the JSON line by line but still includes some code to make sure the file is always valid JSON:
https://github.com/femtobit/netket/blob/7ee7cfbb977c5f5d4dd8dfd9fc6de2bdab956306/NetKet/GroundState/json_output_writer.hpp#L62-L77

femtobit:
I have now extracted the common output code from the VMC and imaginary time classes. Also, Codacy is now happy as well.

> AT LAST, I think I solved the MPI leaks issue.

Great! Indeed this makes the test output a lot more readable again and now everything passes.

Regarding MPI: So far the code here is not parallelized in any way. There are several ways that could be done for the imaginary time propagation, but I'm not sure what the best strategy is.

One option is to parallelize the matrix-vector product in the EOM, but I think the systems need to be quite large to actually benefit from this (also, it is not trivial to do that for the Eigen matrices - though Eigen can use OpenMP, if we'd want that).

Test/UnitaryDynamics/unit-timeevolution.cpp
```diff
@@ -8,9 +8,9 @@
 #include <Eigen/Eigen>
 
 #include "netket.hpp"
-#include "UnitaryTimeEvolution/time_evolution.hpp"
-#include "UnitaryTimeEvolution/TimeStepper/explicit_time_steppers.hpp"
-#include "UnitaryTimeEvolution/TimeStepper/controlled_time_steppers.hpp"
+#include "UnitaryDynamics/time_evolution.hpp"
```
gcarleo: Can we change the name to `Dynamics` or just `TimeEvolution`? For the imaginary time case this is not a unitary evolution for example

gcarleo:
@femtobit , it's almost all good (I just suggested to remove the `Unitary` word in the filenames)!! We can merge this soon, then we will also need some documentation... but we don't need to have it before the merge I guess... thanks again for this!

femtobit:
I'm not sure whether I will have time to do that in the next days, but I agree that some documentation would be nice. (And it would also be interesting to compare the performance and accuracy of this to the other methods.) But I also think we can do this in another PR later.

gcarleo:
OK no problem, let me know when you have time to write the documentation, some PR to the website will be required but it should be straightforward (I hope!) 

Thanks for this implementation!

Issue #61, opened 2018-08-21
Cannot fix the total magnetization to an odd value?


twesterhout:
It seems that something's wrong with MC sampling for odd magnetizations. Example config:
```json
{ "Optimizer": {"Name": "AdaMax"}
, "Hamiltonian": {"Name": "Heisenberg", "TotalSz": 1}
, "Graph": {"Dimension": 1, "Name": "Hypercube", "Pbc": true, "L": 5}
, "GroundState": {"OutputFile": "test", "Diagshift": 0.1, "NiterOpt": 20, "UseIterative": true, "Nsamples": 2000.0, "Method": "Sr"}
, "Machine": {"Alpha": 2, "Name": "RbmSpin"}
, "Sampler": {"Name": "MetropolisHamiltonian"}
}
```
`netket` executable then terminates with the following error:
```
Cannot fix the total magnetization
```
It looks like a bug to me, because clearly 5 spins _can_ have a total magnetisation of 1. Am I missing something here?

everthemore:
If they have quantum numbers +0.5 and -0.5 they won't have total Sz = 1, right? So single basis states can't be made to obey this. Does it work for L = 6?

twesterhout:
> If they have quantum numbers +0.5 and -0.5 they won't have total Sz = 1, right?

Right! It makes sense, thanks. I assumed the one-halves weren't taken into account here but rather absorbed by the weights.

However, if I follow your advice and use `TotalSz = 0.5` instead, I get the very same error. So it seems that there's more to the story.

PR #62, opened 2018-08-22
A fix for #61

twesterhout:
The algorithm is basically a copy-paste from [here](https://github.com/twesterhout/tcm-swarm/blob/master/cbits/spin_state.hpp#L149) and seems to be working.


gcarleo:
Does it work only for spins 1/2?

gcarleo:
Ok @twesterhout , no problem and thanks for fixing this! @kchoo1118 is your initialisation part affected by these changes/ the bug reported by @twesterhout in #61 ?

kchoo1118:
Looks good. Hmm I never encountered those issues.

PR #66, opened 2018-09-05
Refactor JSON output to prevent concurrency bugs

femtobit:
The JSON output bug fixed by #64 was caused because converting an `ObsManager` to JSON performs an MPI reduction and thus has to be called from all processes, while in the first version of my code it was only called by rank 0.

This PR adds a few changes to the output code to make errors like this less likely.


gcarleo:
@femtobit , thanks, I will look into this. A further issue we just identified is that now we are appending to the wave function file, instead of rewriting it from scratch, as we were doing before. The expected behaviour is to save the wave-function for backup purposes, so appending to the old file leads to undefined behaviour when loading the wave-function in further calculations.

femtobit:
> A further issue we just identified is that now we are appending to the wave function file, instead of rewriting it from scratch, as we were doing before.

Ah, I see. I changed the code to rewrite the wavefunction every time.

Maybe we should make this configurable, as it could be useful for tests with small systems to have the complete state information saved?

gcarleo:
@kchoo1118 what's the current status with the wave function log file then? You mentioned that this issue might be already solved in more recent commits?

kchoo1118:
yea i just modified it in my local branch so that the wavefunction is not constantly appending

gcarleo:
OK then I will merge this PR, and maybe we can do another PR for the wavefuncion appending issue @kchoo1118  :)

PR #67, opened 2018-09-07
Introduced Lanczos ED from IETL library

alexwie:
Dear all, 

I wrote a wrapper for using the IETL Lanczos code (http://alps.comp-phys.org/static/doc2.1.0/ietl.html). The ground state energy computation in NetKet would now be done with this library. Probably we should discuss some more detailed user interface for the algorithm (how to set parameters, like precision for the Lanczos iteration. There are also some test, which could be extended. 

Best,
Alex


gcarleo:
Thanks @alexwie , excellent work! The Lanczos part integrates very nicely via the IETL, and diagonalising a 20 spins Hamiltonian takes about 20 seconds on my laptop, which seems a pretty good performance to me.  

Do we have tests on complex hamiltonians? In the 'GroundState' section of the input I would allow the user to chose among two different methods 'SparseEd' and 'FullEd'. Then, you can easily retrieve custom Lanczos parameters from the input, but I would maybe leave the current values as default values? Is there a warning or something similar if the Lanczos is not converged within the given number of iterations? 

Something we should also think about is how to work with the Observables in the case of ED. @femtobit already did some work in the context of Imaginary time propagation, and I guess that could be easily exported also to the 'SparseEd' case

gcarleo:
However it seems that there is an issue with memory usage: diagonalising 22 spins on my laptop uses about 5GB of memory, whereas I would have expected something of the order of hundred MB... I guess there is some memory inefficient part to be corrected, maybe in the hamiltonian wrapper

alexwie:
Currently, I did not check for complex Hamiltonians, and it might well be that this is not working correctly yet. Are there any predefined complex Hamiltonians we could use? I think it would be quite easy to add this to the test. 

It's a good idea to let the user choose between full and sparse ED. We could then just allow for some parameters in the input. The convergence check is also a good idea. The observables can probably be measured when returning the eigenvector from the IETL library, a feature we can definitely use. 

We should check where exactly the memory is used up and fix this issue. The IETL should not use much memory (3 or 4 Lanczos vectors, should be around 50 MB I suppose for 20 spins), if used correctly. Of course, I am using the matrix generated by the SparseWrapper, I am not sure whether this is sparse. This probably uses all the memory.

NetKet/GroundState/exact_diagonalization.hpp
```diff
@@ -0,0 +1,74 @@
...
+    using iter_t = ietl::lanczos_iteration_nlowest<double>;
+    using lanczos_t = ietl::lanczos<matrix_t, vectorspace_t>;
+
+    // TODO: make these parameters accessible to user
+    int max_iter = 1000;
+    int seed = 42;
+    double precision = 1e-8;
+
+    SparseMatrixWrapper<Hamiltonian> matrix(hamiltonian);
+    const matrix_t numerical_matrix = matrix.GetMatrix();
```
femtobit: This copies the whole sparse matrix instead of taking a reference to the one store in `matrix`, so at least this will take twice as much memory as necessary (since the matrix wrapper is still in scope).

NetKet/GroundState/ground_state.hpp
```diff
@@ -12,102 +12,95 @@
...
-  explicit GroundState(const json &pars) {
-    std::string method_name;
-
-    if (FieldExists(pars, "GroundState")) {
-      method_name = FieldVal(pars["GroundState"], "Method", "GroundState");
-    } else if (FieldExists(pars, "Learning")) {
-      method_name = FieldVal(pars["Learning"], "Method", "Learning");
-      // DEPRECATED (to remove for v2.0.0)
-      WarningMessage()
+  class GroundState {
```
femtobit: The indentation is not consistent with the style guide. I suggest you run `clang-format` as describe in https://github.com/netket/netket/blob/master/CONTRIBUTING.md#c-coding-style on the files changed in this PR.

Test/GroundState/unit-exact-diagonalization.cc
```diff
@@ -0,0 +1,40 @@
...
+    std::string name = input_tests[i].dump();
+    SECTION("Exact Diagonalization test (" +
+            std::to_string(i) + ") on " + name) {
+      netket::json pars = input_tests[i];
+      netket::Graph graph(pars);
+      netket::Hamiltonian hamiltonian(graph, pars);
+
+      std::vector<double> eigs_lanczos = eigenvalues_lanczos(hamiltonian);
+      std::vector<double> eigs_full = eigenvalues_full(hamiltonian);
+      REQUIRE(std::abs(eigs_full[0] - eigs_lanczos[0]) < 1e-12);
```
femtobit: Catch2 provides a class for approximate floating-point comparisons: https://github.com/catchorg/Catch2/blob/master/docs/assertions.md#floating-point-comparisons
The result should be the same, of course, but maybe it'd be more idiomatic to use that.

gcarleo:
Indeed the problem is that the matrix, for the Ising model, has (N+1)*(2^N) non-zero entries. This on 22 spins is already about 2 GB, including the IETL vectors. The remaining allocated memory is maybe the temporary vector used to initialise the sparse matrix more efficiently. 

It would maybe make sense to do without the SparseMatrix, and define the operations directly on the Hamiltonian, to save memory...  let's think about this, maybe later. 

For complex Hamiltonians, no, we don't have any predefined one at the moment. Maybe a very simple one would be a spin model with a term X + Y

femtobit:
Also, I think we still store the whole Hamiltonian matrix. Since it is Hermitian, storing only the upper or lower triangle would save space. I don't know hwther IETL supports that. The Eigen [`SelfAdjointEigenSolver`](http://eigen.tuxfamily.org/dox-3.2/classEigen_1_1SelfAdjointEigenSolver.html) does:
> Only the lower triangular part of the input matrix is referenced.

gcarleo:
Indeed, now I am returning the sparse matrix by reference, this indeed shaves a factor of two in memory consumption, but we still have a factor of 2 kicking back, since we have the vector of triplets when we construct the sparse matrix. I am not sure there is a good way to remove this further factor of two without paying a substantial price in efficiency. 

I think that the only solution to really save memory is to create a direct Wrapper for the Hamiltonian (supporting matrix vector multiplication), without forming a SparseMatrix, which is intrinsically at least a factor of N more memory intensive, because of the structure of the Hamiltonian. 

Also, Eigen doesn't seem to support Sparse hermitian matrices, which would have allowed us to save a factor of two. (IETL just cares about the fact that the user-defined matrix type has matrix vector operations, does not store the matrix itself).

femtobit:
> [...] but we still have a factor of 2 kicking back, since we have the vector of triplets when we construct the sparse matrix. I am not sure there is a good way to remove this further factor of two without paying a substantial price in efficiency.

It seems that using `mat.coeffRef(i,j) = ...` can be sped up by reserving enough space via `mat.reserve(nnz)` (see [this StackOverflow post](https://stackoverflow.com/questions/18154027/sparsematrix-construction-in-eigen)). For that, we'd need to be able to estimate the number of non-zero entries per row or at least get a reasonable upper bound.

> Also, Eigen doesn't seem to support Sparse hermitian matrices, which would have allowed us to save a factor of two. (IETL just cares about the fact that the user-defined matrix type has matrix vector operations, does not store the matrix itself).

The way it is handled in Eigen is to just use the lower triangular part of the matrix for the self-adjoint eigenvalue solver (so the upper triangle can just be left empty when initializing the matrix). If we can get IETL to do that, it would be fine. Otherwise we could also try specializing or current wrapper classes to support this and look into passing them directly to IETL. Do you know what exactly IETL requires for the matrix type? Just matrix-vector multiplication via `operator*`?

femtobit:
> I think that the only solution to really save memory is to create a direct Wrapper for the Hamiltonian (supporting matrix vector multiplication)

Indeed, if we can change the matrix wrapper classes so that they could be passed to IETL directly, we could also use the existing `DirectMatrixWrapper` class to compute the matrix elements on-the-fly, as you suggested.

alexwie:
It's definitely possible to combine the IETL with computing Matrix elements on-the-fly. Still, this will make the computation much slower. So the question is whether to go for speed or memory efficiency or let the user decide. 
Allocating the memory of the sparse matrix in advance and then filling it is probably a good way to save memory. For the IETL library, only a routine called ietl::mult has to be defined for the given datatypes. If we implement this routine only using the upper half of the Hermitian matrix, we can use it. This would save again a factor 2 in memory.

femtobit:
> So the question is whether to go for speed or memory efficiency or let the user decide.

That should be left to the user. Some may want fast computation and has a cluster node with lots of memory available while others might need to optimize for less memory usage. That was the general idea for providing the matrix wrapper classes as a way for the suer to choose between dense, sparse and no storage for the matrix elements.

> For the IETL library, only a routine called ietl::mult has to be defined for the given datatypes. If we implement this routine only using the upper half of the Hermitian matrix, we can use it. This would save again a factor 2 in memory.

Great, then lets implement `ietl::mult` directly for `AbstractMatrixWrapper` so we can choose which of its subclasses to use at runtime, depending on the config.

gcarleo:
I think the best is to implement a direct method for ```ietl::mult``` , this seems like the only solution to save memory. The others might shave a factor of 2 or so, but do not solve the real issue, namely the factor of ```N``` or more associated with the sparse matrix representation of the Hamiltonian. Of course the new method might be slower, but we can just give the user the choice on whether to use one or the other one... Also, I am not so sure the direct multiplication method would be dramatically slower...!

alexwie:
I'll take this part. Could you just first implement the User interface you want, such that I could fill in this functionality? Also, do you have a predefined complex Hamiltonian for testing?

NetKet/GroundState/exact_diagonalization.hpp
```diff
@@ -0,0 +1,74 @@
...
+
+  normal_dist_t dist(-1., 1.);
+  ietl::random_generator<std::mt19937, normal_dist_t> random_gen(dist, seed);
+  iter_t iter(max_iter, num_eigenvalue_to_converge, precision, precision);
+  lanczos_t lanczos(numerical_matrix, ietl_vecspace);
+
+  lanczos.calculate_eigenvalues(iter, random_gen);
+
+  std::vector<double> eigs_lanczos = lanczos.eigenvalues();
+  eigs_lanczos.resize(first_n);
```
twesterhout: I'm curious about this line. It seems that `tmatrix::eigenvalues` computes *all* eigenvalues. Or am I missing something? Isn't it a waste of computational resources to throw away all but `first_n` first eigenvalues if they've already been computed anyway?

twesterhout:
And could you perhaps follow #70 and avoid copying the sources of IETL?

NetKet/GroundState/exact_diagonalization.hpp
alexwie: It's true, that tmatrix::eigenvalues computes all eigenvalues of the T-matrix, which are an approximation to the real eigenvalues. Only the first_n of them are converged to real eigenvalues, so the question remains, how to deal with these possibly not converged eigenvalues. One could either keep then as approximative eigenvalues (although it's not clear how good the approximation is) or just throw them away, since they have not converged yet to the prescribed precision.

twesterhout: Oh, you're absolutely right @alexwie! I missed a few lines in your implementation. All makes sense now (except for the copying, but ALPS is way too old for move semantics :grinning:).

gcarleo:
@alexwie , in 'Hamiltonian/MatrixWrapper/directmatrixwrapper.hpp' there is a (untested) direct wrapper of the Hamiltonian which could be used for IETL purposes (basically the method 
'Apply' should be adapted to be used with IETL). Basically at the moment I would just put a boolean switch that allows you to chose either this `DirectMatrixWrapper` or the `SparseMatrixWrapper`. Then we can think later on how to refine the user interface from Json... I guess now it is important just to check that `DirectMatrixWrapper` is not several orders of magnitude slower than the sparse matrix one!! And no, we don't have yet a test for complex hamiltonians, I will work on that.

femtobit:
>  Then we can think later on how to refine the user interface from Json...

You can use `ConstructMatrixWrapper` (though the direct wrapper needs to be added to the function):
https://github.com/netket/netket/blob/a8d9cabbe097317909e93b0500c50967928f119a/NetKet/Hamiltonian/MatrixWrapper/matrix_wrapper.hpp#L17

Users can specify the desired wrapper via e.g. `"MatrixWrapper": "Sparse"` in the JSON.

External/ietl/eigeninterface.h
```diff
@@ -40,8 +40,14 @@ namespace ietl {
             const Eigen::Matrix<TCoeffs, Eigen::Dynamic, 1>& x,
             Eigen::Matrix<TCoeffs, Eigen::Dynamic, 1>& y)
   { y = a*x; }
-
-
+  
+  template <class matrix_t, class TCoeffs>
+  void mult(const matrix_t& a,
```
femtobit: This modifies the source file of an external dependency, which can become a problem when updating the dependencies. (It can also be confusing if our version of IETL is subtly different from the original withour clear documentation.) Also note that in the current master the external dependencies are not stored in the repository anymore, see PR #70.

You could move the `ietl::mult` function to `AbstractMatrixWrapper.hpp`, where the matrix wrapper base class is defined.

NetKet/GroundState/exact_diagonalization.hpp
```diff
@@ -23,52 +23,98 @@
...
+
+    normal_dist_t dist(-1., 1.);
+    random_t random_gen(dist, seed);
+
+    // Converge the first_n eigenvalues to precision
+    iter_t iter(max_iter, first_n, sqrt(precision), sqrt(precision));
+
+    std::vector<double> eigs_lanczos;
+  
+    if (matrix_free) { // Matrix-free computation
```
femtobit: This part has some code duplication. How about doing something like this at an appropriate point in the code:
```cpp
auto matrix = ConstructMatrixWrapper(pars);
// ...
detail::eigenvalue_lanczos_run(..., *matrix)
```
The `ConstructMatrixWrapper` function returns `unique_ptr<AbstractMatrixWrapper>` and any child of `AbstractMatrixWrapper` will work with your version of `ietl::mult` (which is the idea behind having an abstract base class for the matrix wrappers).
Users can specify the desired wrapper via e.g. `"MatrixWrapper": "Sparse"` or  `"MatrixWrapper": "Direct"` in the JSON, just like for the imaginary time propagation.

NetKet/GroundState/ground_state.hpp
```diff
@@ -83,22 +83,38 @@ class GroundState {
...
-      file_ed << j << std::endl;
-      file_ed.close();
-
+    } else if (method_name == "EdFull") {
+      double precision;
+      int n_eigenvalues, max_iter, random_seed;
+      get_ed_parameters(pars, precision, n_eigenvalues, random_seed, max_iter);
+      std::vector<double> eigs = eigenvalues_full(hamiltonian, n_eigenvalues);
+      write_ed_eigenvalues(pars, eigs);
+    } else if (method_name == "EdSparse") {
```
femtobit: I would suggest leaving the method name as `Ed` (instead of having `EdSparse` and `EdFly`) and let the user select between sparse and matrix-free in the config as mentioned above.
`EdFull` on the other hand seems fine to me, as it uses the Eigen eigenvalue solver instead of lanczos.

NetKet/GroundState/exact_diagonalization.hpp
```diff
@@ -23,52 +23,98 @@
 #include "Hamiltonian/MatrixWrapper/matrix_wrapper.hpp"
 
 namespace netket {
+  
+  namespace detail {
+    template <class matrix_t, class iter_t, class random_t>
+    std::vector<double>eigenvalues_lanczos_run(const matrix_t& matrix,
```
femtobit: A minor point regarding formatting: According to the Google Style Guide [function names](https://google.github.io/styleguide/cppguide.html#Function_Names) should use `MixedCaseLikeThis`, which is used for the rest of NetKet as well.

alexwie:
Dear all,

I implemented the on-the-fly ED. The user can now choose between "EdFull", "EdSparse" and "EdFly". The on-the-fly version uses only little memory but is quite slow. The user can now also specify the precision, maximum number of iterations, number of eigenvalues to converge and the initial random seed, see the ExactDiag Tutorial for the usage. 

Best,
Alex

femtobit:
> I implemented the on-the-fly ED. The user can now choose between "EdFull", "EdSparse" and "EdFly". The on-the-fly version uses only little memory but is quite slow. The user can now also specify the precision, maximum number of iterations, number of eigenvalues to converge and the initial random seed, see the ExactDiag Tutorial for the usage.

Hi Alex, thanks for the work. I added a few review comments in between your two commits.

femtobit:
I have added a few changes related to the direct wrapper:
* I noticed that the index ordering was swapped in `DirectMatrixWrapper::Apply` (i.e., it was computing `H.transpose() * v` instead of `H*v`). Now there is a test comparing the result of direct and sparse wrapper.
* The `HilbertIndex` is now stored between calls to `Apply`.
* `HilbertIndex` now uses `std::unordered_map` instead of `std::map`.

> The on-the-fly version uses only little memory but is quite slow.

The last two points should have an impact on performance. @alexwie, could you check whether the performance has changed now compared to your previous tests?

gcarleo:
I did a quick test before this changes, and unordered_map is actually slower

femtobit:
> I did a quick test before this changes, and unordered_map is actually slower

Interesting. Though the results might be different now because previously the direct wrapper created a new `HilbertIndex` for each call to `Apply`. So in case the setup of the `unordered_map` is slow, this might have influenced the result. Now the setup should only be done once.

femtobit:
Not sure why Travis shows the build as failed. The build for the latest commit (which is c28171a) works:
https://travis-ci.org/netket/netket/builds/429634815

Not sure why it links to [another build for a different commit](https://travis-ci.org/netket/netket/builds/429634842) in the interface here.

gcarleo:
@femtobit , indeed  the new version is about 9% slower than the one with ```std::map``` on a L=16 spin systems... we should see what we can further improve, but I am not sure we can make it as fast as the sparse one

NetKet/GroundState/ground_state.hpp
gcarleo: I would put everything under ```Ed``` for consistency, and allow to chose between ```Sparse```, ```Full```, ```Direct``` ?

NetKet/Hamiltonian/MatrixWrapper/direct_matrix_wrapper.hpp
```diff
@@ -26,24 +31,25 @@ namespace netket {
  * computed from Operator::FindConn every time Apply is called.
  */
 template <class Operator, class WfType = Eigen::VectorXcd>
-class DirectMatrixWrapper : public AbstractMatrixWrapper<WfType> {
+class DirectMatrixWrapper : public AbstractMatrixWrapper<Operator, WfType> {
   const Operator& operator_;
+  HilbertIndex hilbert_index_;
```
gcarleo: Then maybe we should store also Hilbert & hilbert_; here @femtobit

NetKet/Hamiltonian/MatrixWrapper/direct_matrix_wrapper.hpp
```diff
@@ -53,9 +59,9 @@ class DirectMatrixWrapper : public AbstractMatrixWrapper<WfType> {
       for (size_t k = 0; k < connectors.size(); ++k) {
         auto vk = v;
```
gcarleo: Most likely this and the following line are quite slow... not sure how to improve on this without losing generality

NetKet/Hilbert/hilbert_index.hpp
```diff
@@ -32,7 +32,7 @@ class HilbertIndex {
 
   const int size_;
 
-  std::map<double, int> statenumber_;
+  std::unordered_map<double, int> statenumber_;
```
gcarleo: maybe go back to ```std::map``` , I measured a consistent slowing down with ```unordered::map```, but this might be allievated changing the hashing function?

femtobit: > @femtobit , indeed the new version is about 9% slower than the one with std::map on a L=16 spin systems...

Yes, I get the same result now (actually, the difference is even larger on my machine):
```
                  map   N Mean Runtime (s) Std Dev (s)
0  std::unordered_map  14           8,1442  0,0103
2            std::map  14           3,0302  0,0142
1  std::unordered_map  16          47,1230  0,5000
3            std::map  16          16,4941  0,0435
```
For now I can change it back to `std::map`, then.

NetKet/Hamiltonian/MatrixWrapper/direct_matrix_wrapper.hpp
femtobit: Since this is just a reference, storing it has no noticeable performance impact so I think this is not necessary.

gcarleo: True, agreed

gcarleo:
The last commit speeds up the ```Apply``` function of about 30% or more. I am now using a function to compute number representations avoiding temporary copies of visible vectors, nominally improving of a factor ```N``` over previous implementation. In practice the speed up is smaller, but still noticeable

gcarleo:
So, to sum up on this PR: I think this is quite solid and we are very close to merging! 

Performances of the Sparse version are very good, and the Direct version is just fine (I think). It is about a factor of 20 slower, but it also uses a factor of 10 less memory. Ideally we might improve upon it a bit more, but maybe there is no need to do that in this PR. 

Before merging, there is just a few remaining things mainly related to the input mode (basically choose the type of ```Ed``` to be performed), the output (we don't support measurement of observables now, is that correct? @femtobit how hard it is to use the infrastructure you put on for the time evolution also in this case?). Then, there is the minor (for the moment) problem of deciding where to put the IETL headers. For this one, I would suggest to move the parts @alexwie has written inside the NetKet folder, and leave the pure IETL headers where they are/ possibly add an option to download them in the CMake, as we are now doing for the other external libraries. This also raises the question of how to guarantee that those libraries are available for download at compile time, and if we want to have backup copies of those libraries on our servers. But maybe I would leave these last points (not strictly related to ED) for another PR.   

What do you guys think? Who is willing to do some of those changes?

femtobit:
Regarding the output: We will need the eigenvectors to compute the observables, while the current version of the code only computes the eigenvalues, if I'm not mistaken: https://github.com/netket/netket/blob/a9ccd441e4c83a760f0fa9fa26e50cc852729513/NetKet/GroundState/exact_diagonalization.hpp#L38-L39

There is also the question of whether we want to compute the observables at every step of the Lanczos solver (then we'd need a way to run the IETL solver and pause every fixed number of steps to compute and save the observable values) or if we just compute then once in the end after the iteration is done?

> This also raises the question of how to guarantee that those libraries are available for download at compile time, and if we want to have backup copies of those libraries on our servers.

We could fork the respective Github repositories and put them under the netket organization. Then we can use these forks in the cmake file.

Still, I'm not perfectly happy with cmake downloading the dependencies only when compiling the code, but changing that would probably mean using git submodules for the dependencies or something similar.

alexwie:
I am not sure how to deal with the dependency of the IETL. Basically there are 3-4 header files that are original code by the ALPS team which comprises the core algorithm. The rest is an interface I wrote to both the Wrappers you had in netket and also doing the eigenvalue calculations with Eigen instead of Lapack (as originally intended by the IETL). Hence the code is somehow mixed. One could split this up, but the full original IETL contains quite a bit more than just the Lanczos routine. I am not sure whether the IETL code will be developed further and whether there is a github repo. Couldn't we just keep it as it is? I think that will cause the least trouble. What do you think?

twesterhout:
> Couldn't we just keep it as it is? I think that will cause the least trouble. What do you think?

Short answer is **no, we can't**. The reason is that ALPS License 1.0 is incompatible with Apache License 2.0. And now reading ALPS License a bit more carefully I would argue that using ALPS is a huge pain :( We have to redistribute their LICENSE, CITATION.txt etc., print a message on start-up saying that we're using ALPS... I would even ask a professional to clarify whether it's at all possible to use ALPS in this project.

Sorry for not raising this issue earlier. Please, take this licensing stuff seriously!

gcarleo:
@twesterhout we were aware of the licensing issue from the beginning, but the IETL is actually also licensed under the Boost license (I know the developer and asked him directly). The issue is that we didn't find online the corresponding files with the boost headers (there is a broken link), we will correct this as soon as possible.

gcarleo:
Thanks @fabienalet , I figured you might know better about this! 
Maybe the version you mentioned is already the one @alexwie started from. As far as I understand that is a GPL 2.0 or higher (i.e. 3.0 is allowed, meaning we have less compatibility issues with Apache). However, I recently asked Matthias, and the newer IETL version here http://alps.comp-phys.org/static/doc2.1.0/ietl.html should be under Boost license, which is less of a headache for us. Only problem is that the download link is broken, hopefully we will have a fix soon.

twesterhout:
> @twesterhout we were aware of the licensing issue from the beginning, but the IETL is actually also licensed under the Boost license (I know the developer and asked him directly).

@gcarleo, @fabienalet thanks, This is great news!

> The issue is that we didn't find online the corresponding files with the boost headers (there is a broken link), we will correct this as soon as possible.

Then I'm afraid I must insist on deferring the merge until we have either a download link or a copy of the package.

Jumping back to performance stuff, does anyone have an up-to-date profile data for this PR (otherwise I'll create one myself, but it might take me a few days). This whole business of `HilbertIndex` seems too heavy-weight for the task and I'm wondering whether it shows up in the profiling results.

gcarleo:
@twesterhout , agreed ```HilbertIndex``` is certainly not optimal, there is large room for improvement. The only thing is that we also need to keep it general enough (I can think of several ways of making it blazing fast for spins 1/2, using binary operations and integers only, but that would be very problem specific). Coming to your question: no, I don't have profiling data, maybe @femtobit ? I just did a few changes that I could see "by eye" were affecting the performance, and managed to improve it of a sizeable margin. My bet is that now the bottleneck is ```FindConn`` in the Hamiltonian, but we can't do much about it if we want to keep this thing general, I suspect.  Of course if you have a faster version that would be great.

femtobit:
Okay, I have done a bit of profiling. I am not an expert in interpreting these results, but here is a call graph for lancoz with the direct wrapper on a `N=16` ising chain:
prof
The raw profiling data (from `gprof`) is here: https://gist.github.com/femtobit/ed1072a49bee865b87dfa0d8663e3503

I am not an expert in interpreting profiling data, but I suspect there is still something to gainm from optimizing the `HilbertIndex` code. But @gcarleo is also right that `FindConn` plays a significant role. Note that this method seems to spend some time with `std::vector` insertion operations. I suspect that there is room for some improvements.

femtobit:
Just for comparison, here is the same call graph for the imaginary time propagation. The bottleneck is exactly the same, also with a splitting of about 40% time spent in `FindConn`, 20% in `NumberToState`, and 10% in `StateToNumber`.

prof_it

alexwie:
@femtobit, this profiling looks amazing! You can really see where the problem is. I think, the performance issue is that while applying an operator, some abstract objects (this also includes std::vectors) are created and copied, which is slow, would you agree? In my codes, I try to have this part as fast as possible. Usually, I split up an operator in non-branching operators, such that applying one operator to a configuration yields only one different configuration. On small lattices you can code configurations as integers, which is fast. Of course, this does not work anymore for larger lattices. Since this is a quite essential part to maybe also the other computations, like the variational Monte Carlo, it could be worth thinking more about how to have this part as fast as possible, while retaining a certain degree of freedom in the possible operators.

femtobit:
> I think, the performance issue is that while applying an operator, some abstract objects (this also includes std::vectors) are created and copied, which is slow, would you agree?

Indeed, that seems to have been the case, as PR #71 illustrates.

alexwie:
Dear all, there is now support for computing ground state observables using ED. See Tutorials/ExactDiag on how to use it. Again, the user can choose between Dense/Sparse/Direct Matrix Wrappers. I added tests to check whether the energy of the ground state is the lowest eigenvalue. Please let me know what you think.

gcarleo:
Thanks @alexwie , it looks good to me, I just pushed a few changes related to what we did upstream with the CMake etc. I was doing some test though, and found that the "free" solver used as much memory as the sparse one on 24 sites? Maybe I did something wrong, do you have the same memory footprint?

gcarleo:
@femtobit , it also seems that this version breaks compatibility with intel compilers (intel c++ 17), because of some issue with move constructors etc. Do you know how to fix this?  

```
In file included from /mnt/home/gcarleo/netket1/NetKet/GroundState/ground_state.hpp(28),
                 from /mnt/home/gcarleo/netket1/NetKet/netket.hpp(20),
                 from /mnt/home/gcarleo/netket1/NetKet/netket.cc(15):
/mnt/home/gcarleo/netket1/NetKet/GroundState/imaginary_time.hpp(43): error: function "netket::JsonOutputWriter::JsonOutputWriter(const netket::JsonOutputWriter &)" (declared implicitly) cannot be referenced -- it is a deleted function
                                      std::move(output), range);
                                      ^

In file included from /mnt/home/gcarleo/netket1/NetKet/GroundState/ground_state.hpp(28),
                 from /mnt/home/gcarleo/netket1/NetKet/netket.hpp(20),
                 from /mnt/home/gcarleo/netket1/NetKet/netket.cc(15):
/mnt/home/gcarleo/netket1/NetKet/GroundState/imaginary_time.hpp(54): error: function "netket::JsonOutputWriter::JsonOutputWriter(const netket::JsonOutputWriter &)" (declared implicitly) cannot be referenced -- it is a deleted function
          output_(std::move(output)),
                 ^

In file included from /mnt/home/gcarleo/netket1/NetKet/GroundState/ground_state.hpp(29),
                 from /mnt/home/gcarleo/netket1/NetKet/netket.hpp(20),
                 from /mnt/home/gcarleo/netket1/NetKet/netket.cc(15):
/mnt/home/gcarleo/netket1/NetKet/GroundState/variational_montecarlo.hpp(121): error: no operator "=" matches these operands
            operand types are: nonstd::optional_lite::optional<netket::JsonOutputWriter> = netket::JsonOutputWriter
        output_ = JsonOutputWriter::FromJson(pars_gs);
```

femtobit:
@gcarleo Interesting. I will look into this.

gcarleo:
> @gcarleo Interesting. I will look into this.
> (Automatic tests with the Intel compiler would be very useful to have. In principle this seems to actually be [possible to do with Travis](https://github.com/nemequ/icc-travis), but might take some effort to get right.)

@femtobit , thanks. I agree that this would be a very valuable option. The problem is that this is a bit tricky since intel compilers are proprietary, and it's not clear if with an open-source license we can set it up on Travis or not. I will try to learn more about that... in the meanwhile I guess you can just download a copy of the compiler on your own machine, if you don't have it already!

alexwie:
I now added the License Files with modifications for the IETL library. I also fixed a small bug, since the matrix format was chosen wrongly from the input (sparse matrices were always selected). There is also a test for complex Hamiltonians.

femtobit:
@gcarleo Regarding the compiling issue you mention, I tried to compile netket on a cluster machine using the following compiler version:
```
$ CC=icc CXX=icpc I_MPI_CC=$CC I_MPI_CXX=$CXX cmake -DCMAKE_BUILD_TYPE=Debug cmake ..
-- The C compiler identification is Intel 17.0.7.20180403
-- The CXX compiler identification is Intel 17.0.7.20180403
[...]
```
Using these versions of the compiler, everything seems to work fine for me (see https://gist.github.com/femtobit/2874810ffe23c4c0b83dc4726f905c94 for a log). Is this the same Intel compiler version you have been using?

gcarleo:
Hi @femtobit , I was trying with 

'''
icc (ICC) 17.0.4 20170411
''' 

but well, if the 17.0.7 works, then maybe that was some internal bug in the compiler? .... possible

gcarleo:
In any case, I am merging this now, since the license issues were solved

gcarleo:
@femtobit I tried with a more recent Intel compiler and I have the same issue 
```
-bash-4.2$ CC=icc CXX=icpc I_MPI_CC=$CC I_MPI_CXX=$CXX cmake -DCMAKE_BUILD_TYPE=Debug cmake ..
-- The C compiler identification is Intel 19.0.0.20180804
-- The CXX compiler identification is Intel 19.0.0.20180804
```

With result 
```
In file included from /mnt/home/gcarleo/netket1/NetKet/GroundState/ground_state.hpp(28),
                 from /mnt/home/gcarleo/netket1/NetKet/netket.hpp(20),
                 from /mnt/home/gcarleo/netket1/NetKet/netket.cc(15):
/mnt/home/gcarleo/netket1/NetKet/GroundState/imaginary_time.hpp(43): error: function "netket::JsonOutputWriter::JsonOutputWriter(const netket::JsonOutputWriter &)" (declared implicitly) cannot be referenced -- it is a deleted function
                                      std::move(output), range);
                                      ^

In file included from /mnt/home/gcarleo/netket1/NetKet/GroundState/ground_state.hpp(28),
                 from /mnt/home/gcarleo/netket1/NetKet/netket.hpp(20),
                 from /mnt/home/gcarleo/netket1/NetKet/netket.cc(15):
/mnt/home/gcarleo/netket1/NetKet/GroundState/imaginary_time.hpp(54): error: function "netket::JsonOutputWriter::JsonOutputWriter(const netket::JsonOutputWriter &)" (declared implicitly) cannot be referenced -- it is a deleted function
          output_(std::move(output)),
                 ^

In file included from /mnt/home/gcarleo/netket1/NetKet/GroundState/ground_state.hpp(29),
                 from /mnt/home/gcarleo/netket1/NetKet/netket.hpp(20),
                 from /mnt/home/gcarleo/netket1/NetKet/netket.cc(15):
/mnt/home/gcarleo/netket1/NetKet/GroundState/variational_montecarlo.hpp(121): error: no operator "=" matches these operands
            operand types are: nonstd::optional_lite::optional<netket::JsonOutputWriter> = netket::JsonOutputWriter
        output_ = JsonOutputWriter::FromJson(pars_gs);
                ^

compilation aborted for /mnt/home/gcarleo/netket1/NetKet/netket.cc (code 2)
make[2]: *** [CMakeFiles/netket.dir/NetKet/netket.cc.o] Error 2
make[1]: *** [CMakeFiles/netket.dir/all] Error 2
make: *** [all] Error 2
```

femtobit:
@gcarleo Okay, strange that it worked for me on our cluster then. Maybe it's not just the compiler but something about the environment. I can try whether I can reproduce the problem on my own machine.

gcarleo:
It might be that this is related also to the MPI environment... are you using an MPI-intel compiler?

femtobit:
@gcarleo Okay, I have a suspicion about what might be going on:
* When I try to compile netket on a machine with Intel 19 (I think the exact version doesn't matter to much as we have seen the same problem with Intel 17 and 19 now) with GCC 4.8 in the environment it fails with the same message that you got.
* When I add, on the same machine, a newer version of GCC (6.2, to be specific) to the environment it suddenly compiles (without even another call to cmake).

According to the [Intel documentation](https://software.intel.com/en-us/node/522750), icc seems to adapt its behavior to the system version of GCC for compatibility reasons:
> **How the Intel Compiler Uses GCC**
> The Intel compiler uses the GNU* tools on the system, such as the GNU header files, including stdio.h, 
> and the GNU linker and libraries. So the Intel compiler has to be compatible with the version of gcc or 
> g++ you have on your system. For example, if you have gcc version 4.6 on your system, icc behaves
> like gcc 4.6, with the compatible features and behaviors.

And indeed, GCC 4.8 or similarly old versions cannot build netket due to insufficient C++11 support.

Does you test system have an old version of GCC and, if so, could you try compiling netket with a newer version of GCC present to see whether this is the cause?

gcarleo:
@femtobit , I don't know how you pulled this out, but that seems to be precisely the problem!!  

```
g++ (GCC) 4.8.5 20150623 (Red Hat 4.8.5-16)
```

I need to ask the cluster people to update the gcc compiler used to bootstrap the intel one, or something like that...

femtobit:
@gcarleo Yeah, I stumbled upon this by chance because I had a different GCC version in the environment when I looked at the issue again today and was surprised that I could suddenly reproduce it.

If your cluster has a module system or a similar to manage the environment, it should be sufficient to just load another version of GCC into the path.

gcarleo:
It worked at last! thanks

PR #70, opened 2018-09-08
Modernize CMake

twesterhout:
I had a go at handling dependencies using CMake. Although this PR is not really finished yet (I'm curious what Travis CI has to say about it), any feedback at this point is highly appreciated.


CMakeLists.txt
```diff
@@ -1,32 +1,80 @@
...
 
 project(NetKet)
 
-set (CMAKE_CXX_STANDARD 11)
-set (NETKET_SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/NetKet/)
-set (EXTERNAL_INCLUDE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/External/)
+include(CTest) # adds option BUILD_TESTING (default ON)
+include(FetchContent)
+
+# set (CMAKE_CXX_STANDARD 11)
```
gcarleo: Why this is not necessary anymore?

CMakeLists.txt
```diff
@@ -1,32 +1,80 @@
-cmake_minimum_required(VERSION 3.1)
-
-
-set(CMAKE_DISABLE_SOURCE_CHANGES ON)
-set(CMAKE_DISABLE_IN_SOURCE_BUILD ON)
+cmake_minimum_required(VERSION 3.9)
```
gcarleo: I already have trouble with 3.1 being too new on some platforms :) 
Is 3.9 strictly needed or can we do with older versions?

CMakeLists.txt
gcarleo: Indeed it seems that compilation fails without this?

twesterhout: You're very right. I meant to use the `target_compile_feature` instead, but never actually got to adding it... Should be fixed in the last commit. The reason why this approach is preferable nowadays is that it doesn't use global variables.

CMakeLists.txt
twesterhout: Even worse, it looks like FetchContent is only available since 3.11 :) ... We can use ExternalProject instead, but then the download etc. will happen at build time, not configure time. If that's not a problem, I can try to lower the required version. On the other hand, if 3.1 is already too new, does it really matter which version we use? I.e. one has to compile CMake from scratch anyway, they most likely will use the latest (3.12) version.

twesterhout:
> Indeed this is the same approach used in Google Benchmark, if I am not wrong, and It makes a lot of sense also to keep dependencies more easily up to date... (Also I prefer it largely to using submodules)

It's okay as long as we don't have to install the dependencies, only the executable. I do like submodules better, but both Eigen3 and json are huge, which would make NetKet a couple of hundreds of MBs in size -- probably not what we want.

twesterhout:
I had a look at the code coverage part here https://github.com/netket/netket/blob/master/Test/CMakeLists.txt#L6 and it seems quite out of date. It's not tested in Travis either. Does it make sense to just remove it?

CMakeLists.txt
```diff
@@ -1,32 +1,79 @@
 cmake_minimum_required(VERSION 3.1)
 
-
-set(CMAKE_DISABLE_SOURCE_CHANGES ON)
-set(CMAKE_DISABLE_IN_SOURCE_BUILD ON)
```
gcarleo: Why remove these? It's a good idea to prevent builds in the source dir

CMakeLists.txt
```diff
@@ -1,32 +1,79 @@
...
+# MPI
+################################################################################
 find_package(MPI REQUIRED)
-if (MPI_FOUND)
-    include_directories(SYSTEM ${MPI_INCLUDE_PATH})
-else (MPI_FOUND)
-    message(SEND_ERROR "This application cannot compile without MPI")
-endif (MPI_FOUND)
 
-include_directories(${MPI_INCLUDE_PATH})
```
gcarleo: I think that's why tests are failing on some compilers. It seems you need to restore the MPI_INCLUDE_PATH

CMakeLists.txt
twesterhout: Good point.

CMakeLists.txt
twesterhout: 71fe221 should fix the failures

twesterhout:
Seeing that the CI builds fail due to `ctest` timing out rather than actual build failures, I think this PR has reached "mergeable" state. Need that approving review now ;)

gcarleo:
OK great. I restarted the builds to see why `test-timeevolution_default` is failing on some machines. It didn't happen before, there might be some issue related to this CMake changes

twesterhout:
It did pass now, didn't it? But any ideas what might have gone wrong? I haven't really changed the way things are compiled, merely the structure of the CMake files.

PR #71, opened 2018-09-21
Speed-up calculation of the matrix elements

femtobit:
My initial [profiling results](https://github.com/netket/netket/pull/67#issuecomment-423522846) seemed to confirm my suspicion that there is some noticeable overhead from the `std::vector` insertion operations in `FindConn`. To get rid of this overhead, I propose to change the way the matrix elements are computed. This is noticeably faster, at least on my machine (compare `FindConn` with `ForEachConn` ):
```
   Model   N       Method Runtime (s) StdDev (s)  #Runs
0  Ising  14     FindConn       1,986      0,012     16
1  Ising  14  ForEachConn       0,795      0,003     16
2  Ising  16     FindConn      10,175      0,168      5
3  Ising  16  ForEachConn       3,461      0,020      5
4  Ising  20     FindConn     246,630      2,620      2
5  Ising  20  ForEachConn      84,738      0,712      2
```

This PR provides a new method `AbstractHamiltonian::ForEachConn` that iterates over all reachable configurations calling a callback function each time. It should be called like this:
https://github.com/netket/netket/blob/77ef5c2124c02115eee730a0a9e050bd583a5d5c/NetKet/Hamiltonian/MatrixWrapper/direct_matrix_wrapper.hpp#L51-L54

For backwards compatibility, `ForEachConn` is implemented for the `AbstractHamiltonian` in terms of `FindConn`:
https://github.com/netket/netket/blob/77ef5c2124c02115eee730a0a9e050bd583a5d5c/NetKet/Hamiltonian/abstract_hamiltonian.hpp#L87-L88
Thus, all existing Hamiltonians will work with `ForEachConn`. To get an actual speedup, subclasses need to override `ForEachConn`. In this PR, this is implemented for `Ising`:
https://github.com/netket/netket/blob/77ef5c2124c02115eee730a0a9e050bd583a5d5c/NetKet/Hamiltonian/ising.hpp#L140-L165

This is a PR on the `lanczos` branch (PR #67) because it already contains changes in the direct matrix wrapper compared to `master`.

### DIFF
--- a/NetKet/Hamiltonian/abstract_hamiltonian.hpp
+++ b/NetKet/Hamiltonian/abstract_hamiltonian.hpp
@@ -19,7 +19,18 @@
 #include <complex>
 #include <vector>
 
+#include "Hilbert/hilbert.hpp"
+
 namespace netket {
+
+/**
+ * This struct represents a non-zero matrix-element H(v,v') of an operator for a
+ * given visible state v.
+ */
+struct MatrixElement {
+  std::complex<double> weight;    /// The matrix element H(v,v')
+  ConfigurationUpdateRef update;  /// The update neccessary to obtain v' from v
+};
 
 /**
       Abstract class for Hamiltonians.
@@ -50,6 +61,20 @@ class AbstractHamiltonian {
                         std::vector<std::vector<int>> &connectors,
                         std::vector<std::vector<double>> &newconfs) const = 0;
 
+  using ConnCallback = std::function<void(MatrixElement)>;
+  /**
+   * Iterates over all states reachable from a given visible configuration v,
+   * i.e., all states v' such that H(v,v') is non-zero.
+   * @param v The visible configuration.
+   * @param callback Function void callback(MatrixElement mel) which will be
+   * called once for each reachable configuration v'. The parameter mel contains
+   * the value H(v,v') and the information to obtain v' from v. Note that the
+   * member mel.update has reference character and can only be savely used
+   * inside the callback. It will become invalid once callback returns.
+   */
+  virtual void ForEachConn(const Eigen::VectorXd &v,
+                           ConnCallback callback) const;
+
   /**
   Member function returning the hilbert space associated with this Hamiltonian.
   @return Hilbert space specifier for this Hamiltonian
@@ -58,6 +83,22 @@ class AbstractHamiltonian {
 
   virtual ~AbstractHamiltonian() {}
 };
+
+void AbstractHamiltonian::ForEachConn(const Eigen::VectorXd &v,
+                                      ConnCallback callback) const {
+  std::vector<std::complex<double>> weights;
+  std::vector<std::vector<int>> connectors;
+  std::vector<std::vector<double>> newconfs;
+
+  FindConn(v, weights, connectors, newconfs);
+
+  for (size_t k = 0; k < connectors.size(); k++) {
+    const ConfigurationUpdateRef update{connectors[k], newconfs[k]};
+    const MatrixElement mel{weights[k], update};
+    callback(mel);
+  }
+}
+
 }  // namespace netket
 
 #endif
--- a/NetKet/Hamiltonian/abstract_hamiltonian.hpp
+++ b/NetKet/Hamiltonian/abstract_hamiltonian.hpp
@@ -27,9 +27,15 @@ namespace netket {
  * This struct represents a non-zero matrix-element H(v,v') of an operator for a
  * given visible state v.
  */
-struct MatrixElement {
-  std::complex<double> weight;    /// The matrix element H(v,v')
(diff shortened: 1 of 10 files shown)


gcarleo:
@femtobit thanks for this PR, it looks like a great way to avoid moving and constructing temporary vectors around! I am not entirely sure though this will make a huge impact on the Monte Carlo part, but it would be nice to see more in detail once we settle down a bit these two PRs. Are the benchmarks you have reported here for the full Lanczos or just for the single routines?

gcarleo:
I did a couple of benchmarks, for the Lanczos par this is a fantastic improvement over the old version! Now the sparse solver runs 20 sites in 13 seconds and the "on-the-fly" one in about 80 seconds (before there was about a factor of 20 between the two running times!). 
I just have some doubts about 'ConfigurationUpdateRef', since it introduced a concept which should be handled more homogenously also in another parts of NetKet then...

femtobit:
> I just have some doubts about ConfigurationUpdateRef, since it introduced a concept which should be handled more homogenously also in another parts of NetKet then...

I wasn't too happy with the `ConfigurationUpdateRef` either. I simplified the structure a bit. Now everythin needed is contained in the `MatrixElement` struct that is passed to the callback:
https://github.com/netket/netket/blob/2aa204ec154bc0acfedbe9814d818d72281e158a/NetKet/Hamiltonian/abstract_hamiltonian.hpp#L31-L37
Thus, the interface is largely the same as previously, but instead of repeatedly filling and passing `std::vector`s around, we now use `span`s (i.e., non-owning references to existing memory).

femtobit:
> Are the benchmarks you have reported here for the full Lanczos or just for the single routines?

The benchmarks are from running the example from `Tutorials/ExactDiag` with `EdFly`.

> I am not entirely sure though this will make a huge impact on the Monte Carlo part,

Indeed, I wouldn't expect a huge impact there. The runtime of the VMC code is currently mostly determined by `UpdateParameters` and a few other functions. `FindConn` is not significant here. You can see this from the following call graph I obtained for the `Ising1d` example with VMC:

prof

NetKet/Hamiltonian/abstract_hamiltonian.hpp
```diff
@@ -19,8 +19,24 @@
 #include <complex>
 #include <vector>
 
+#include "Hilbert/hilbert.hpp"
+
 namespace netket {
 
+// clang-format off
```
gcarleo: This is the last thing I am not entirely convinced about! I mean, not the clang-format off (btw, why do you have this here?) but rather the MatrixElement struct. 
The thing is that I would expect a MatrixElement object to carry information both about v and v', not only v'! Maybe we can add a span pointing to v? Would it make sense?

gcarleo:
Anyways @femtobit , thanks for the profiling on the Monte Carlo part too! It is crazy that 75% of the time is taken by the linear algebra solver. We will need to do something about that in the future ...

NetKet/Hamiltonian/abstract_hamiltonian.hpp
femtobit: Not sure it makes sense to point to `v`. In places where this struct makes sense to appear, there should always be a known `v'`.

We also could just as well directly pass the matrix element value and two `span`s to the callback. I found the struct a bit easier to read, though. Maybe the name implies a generality that is not not there (and not needed). Would calling it something like `Connection` help?

> the clang-format off (btw, why do you have this here?)

Hm, `clang-format` did add an awkward line break in the struct (due to the comments).

gcarleo: Alright agreed, ```MatrixConnector``` then is a better concept?

gcarleo: In this way in the future we can maybe also have that ```FindConn``` returns a vector of ```MatrixConnector```

femtobit: > In this way in the future we can maybe also have that FindConn returns a vector of MatrixConnector

That would not work well, I think. Since the `span`s do not own the memory they refer to, one needs to be careful with their lifetime. Hence the comment here:
https://github.com/netket/netket/blob/baa7c0d6a29ebd2194b31b2c475fab3848f3d4a0/NetKet/Hamiltonian/abstract_hamiltonian.hpp#L77-L80
That the parameter in `ForEachConn` is only usable inside the callback seems reasonably safe to me. However, I would expect to be able to pass the return value of `FindConn` around safely, without worrying about its lifetime. Thus, a `vector` seems to be the better choice here.

Therefore, I do not see a good way to unify these at the moment. I also propose `ConnectorRef` as a name. The `Ref` part makes it clear that one has to think about the lifetime if this object is passed around.

PR #72, opened 2018-09-30
Add possibility to link against Lapack 

gcarleo:
This PR adds the possibility to link Eigen against native Lapack libraries, typically offering a speed-up over internal Eigen implementations. 
If Lapack is not installed, the CMake falls back to the Eigen internals. 
This PR also removes testing enabled by default, since this substantially increases compilation time (mostly because of the Boost monster used in some unit test).


gcarleo:
LOL, it seems that `nonstd::optional` has a weird conflict with complex numbers in c++11? 
Maybe we need another implementation of optional @femtobit 
```
/home/travis/build/netket/netket/build/External/optional-lite/nonstd/optional.hpp:313:23: error: 
      expected ')'
template< std::size_t I >
                      ^
/usr/include/complex.h:49:11: note: expanded from macro 'I'
#define I _Complex_I
```

femtobit:
That's a strange build failure. It doesn't happen on my machine. Do you now which LAPACK implementation that build is using? Something seems to be including the C header `complex.h` which is strange. In C++, the [`<complex>`](https://en.cppreference.com/w/c/numeric/complex/I) header should not define that `I` macro if I'm not mistaken.

In any case, if needed we could just patch `optional-lite`, renaming that parameter `I` to something else.

femtobit:
> In any case, if needed we could just patch optional-lite, renaming I to something else.

Okay, this is getting more ugly. We also use `I` in some places in our codebase (at least in some of the tests). There has to be a better way than replacing all these occurrences.

gcarleo:
I think this is some weird CMake incostincency. I can reproduce that error
on my machine only if I don't compile with c++11 flags from the command
line. For some reason on Travis CMake is maybe not using c++11 flags

femtobit:
> For some reason on Travis CMake is maybe not using c++11 flags

I added `VERBOSE=1` to the environment variables to get the command line make is actually executing. The `-std=c++11` flag seems to be set correctly:
`/usr/bin/g++-5  -DEIGEN_USE_BLAS -DEIGEN_USE_LAPACKE -I/home/travis/build/femtobit/netket/NetKet -isystem /usr/include/mpich -isystem /home/travis/build/femtobit/netket/build/External/Eigen3 -I/home/travis/build/femtobit/netket/build/External/json -I/home/travis/build/femtobit/netket/build/External/optional-lite  -Wall -Wno-long-long -pedantic -Wextra -Wshadow -O3   -std=c++11 -o CMakeFiles/netket.dir/NetKet/netket.cc.o -c /home/travis/build/femtobit/netket/NetKet/netket.cc`
(From https://travis-ci.org/femtobit/netket/jobs/435814690)

gcarleo:
Alright, I have investigated more the issue, and found that it happens for gcc<8.0. 
I opened an issue on optional-lite, and it has been already addressed and solved. 
Now I guess we need to address it also for our tests...

CMakeLists.txt
```diff
@@ -30,7 +36,7 @@ target_include_directories(nlohmann_json
 ################################################################################
 if (NOT EXISTS "${CMAKE_BINARY_DIR}/External/optional-lite/nonstd/optional.hpp")
     file(DOWNLOAD
-        "https://github.com/martinmoene/optional-lite/releases/download/v3.0.0/optional.hpp"
+        "https://raw.githubusercontent.com/martinmoene/optional-lite/master/include/nonstd/optional.hpp"
```
femtobit: Maybe change this from `master` to the specific commit (that is, use the URL https://raw.githubusercontent.com/martinmoene/optional-lite/e240fdaaa4071a980926dea4ed5b610aa8fbc538/include/nonstd/optional.hpp)?
That way, all users get the same version of `optional-lite` when they build `netket`.

Issue #73, opened 2018-10-02
Look ups for derivative


stavros11:
Hello,

Thanks to everyone for developing and releasing this code.

Please excuse my C++ inexperience, I would like  to ask a quick (and possibly simple) question: In the `abstract_machine.hpp`, is there any reason that there is no `DerLog` function that takes look ups as argument, similarly to `LogVal` and `LogValDiff`? I am assuming that it has to do with the way the derivative and the look up update is handled in the VMC part (for which I haven't checked the code), however wouldn't in principle to be possible to use already calculated look ups for more efficient calculation of derivative?

I am implementing some custom machines based on MPS, where all the left and right contractions are required for the full derivative. These contractions can also serve as look ups to assist `LogValDiff` calculations (particularly in the one-flip case). I am wondering whether it is possible to use them directly in `DerLog` without contracting from scratch every time. I tried defining private look up variables within my class for this and update them within `UpdateLookup` but this did not work, I am guessing because `UpdateLookup` is not called before every derivative calculation.

gcarleo:
Hi Stavros, this is an excellent point. 

Indeed we were thinking of adding lookup-based derivatives as well. These might be useful for example also for feed-forward neural networks. The way to go is to add a method to the abstract machine, something like ```DerLogDiff``` taking as an input the lookup table and the visible configuration. In order for this to work, one has to change a little bit how things are implemented in the Variational code (at the moment I think we are not computing derivatives after each sweep, but at the end, once all the sampled visible configurations have been collected. I would be happy to help you with this part, it is an improvement which could speed-up calculations for some families of variational states

stavros11:
Hi Giuseppe, thank you for the response.

Yes, I can see that the gradient is calculated after sampling, so this excludes the use of lookup. I guess one should move the calculation of `Ok_` and `elocs_` vectors in the sweep loop for this to work, assuming that the sampler updates the lookup every time a flip is accepted. Are there any plans to do this in the future? Also changing the abstract machine could complicate things, as every machine should probably be updated then.
I am not how much it would help in terms of speed, at least for the MPS it might already be costly to constantly update the lookup, however since this is done anyway for `LogValDiff`, it could be used in the derivative too.

PR #74, opened 2018-10-08
Periodic MPS / SBS machines

stavros11:
Hello everyone,

Trying not to increase the volume of pulls with unfinished code, I was planning to propose this pull after having everything working a bit better and well optimized. Given the current timeline, it might be a good idea to share earlier so that you also have some time to review the progress and decide whether this contribution is going to be useful or how we can proceed.

The uploaded files are the following:
- `mps_periodic.hpp`: Implements a periodic MPS machine where all matrices are square with the same (bond) dimension given by user. It is possible to set translational symmetry (same set of matrices repeated across the chain).
- `mps_diagonal.hpp`: Implements a machine identical to `mps_periodic.hpp` where all matrices are diagonal (using `cwiseProduct` instead of matrix multiplication).
- `abstract_mps.hpp`: Contains virtual methods so that the above classes can be used for SBS calculations.
- `sbs.hpp`: Implements a SBS (String Bond State = Product of MPS strings) machine. All SBS strings are periodic.

And a [README explaining the JSON settings.](https://github.com/stavros11/Variational-Monte-Carlo/blob/master/cpp/UserSettings.md)

**WARNING: Unit test failures**
- The real test of `DerLog` for all machines. The imaginary part has no problems and the machines seem to converge to the correct energies for small (`N<10`) 1D Ising and Heisenberg Hamiltonians. I am not sure where the unit test error comes from - in my computer the Jastrow machines fail the same test too.
- I use exceptions in the SBS machine which I think are not allowed in the Google style. This is only in the `from_json` function and does not affect calculations, so it will be replaced soon (sorry for that).

Some additional notes:
1. The look up tables store all left and right contractions. This is efficient in the one-flip case but might not be the optimal when more than one spins are flipped. Storing middle contractions could assist in computations for large systems but would add time in the look up update. It is not straightforward (to me) what is the best to do. Optimizing this is among my priorities.
2. All contractions are calculated from scratch every time `DerLog` is called. [Using look up tables in `DerLog`](https://github.com/netket/netket/issues/73) would help this and other machines, but this requires small changes in the VMC part (and possibly some other parts too). 
3. User can explicitly assign which sites go to each string in SBS through the JSON. This allows the greatest flexibility but might be a bit hard/non-automatic from the user's perspective and also allows non-physical cases (eg. strings that jump sites). An easy way to bypass this is by writing a simple [Python wrapper](https://github.com/netket/netket/pull/65) that creates the "sites to strings" lists for the JSON. A more formal way could be to use the graph object directly in C++. I am not yet sure what's the best way to simplify things while still keeping the settings quite general.
4. The "sites to strings" option in SBS can be used to generalize MPS too (1-string SBS = MPS). Hence the SBS machine can be used to get different MPS shapes (eg. snake MPS), which is not supported in my pure MPS implementations.
5. Weight loading does not work. Machines initialize randomly after loading.
6. My machines are possibly slower than all current machines and the number of parameters is typically larger (eg. ~N^3 for MPS vs ~N^2 for RBM). In any case, I am far from C++ expert, so please comment if you find any "newbie" errors in my code. It seems that the diagonal version is much faster even for small systems, which might be just because the number of parameters is smaller or because the contractions are inefficient.
7. I have an implementation for open boundary MPS which allows different matrix dimensions across the chain. If there is interest I add that one too, probably after I manage to add a feature to transform it to its canonical form.

Thank you very much and I hope this is useful. Many thanks to I. Glasser for getting me involved and explaining many things related to these kind of states. Note that the code hasn’t been reviewed, therefore I am responsible for any issues!

Stavros


gcarleo:
Hi Stavros, thanks for these efforts. We will review it and propose some changes and improvements.

For the moment, it seems that the tests are failing because the code is not compiling correctly (see the output of the Travis build/or run a complete compilation of the tests on your local machine, with ```make test```) 

```
/home/travis/build/netket/netket/NetKet/Machine/mps_diagonal.hpp:485:23: error: 
      use of overloaded operator '=' is ambiguous (with operand types
      'Eigen::DenseCoeffsBase<Eigen::Matrix<std::complex<double>, -1, 1, 0, -1,
      1>, 1>::Scalar' (aka 'std::complex<double>') and 'const
      nlohmann::basic_json<std::map, std::vector, std::basic_string<char>, bool,
      long, unsigned long, double, std::allocator, adl_serializer>::value_type'
      (aka 'const nlohmann::basic_json<std::map, std::vector,
      std::basic_string<char>, bool, long, unsigned long, double,
      std::allocator, adl_serializer>'))
          W_[i][j](k) = pars["Machine"]["W"][seg_init + (d_ * i + j) * D_ + k];
```

gcarleo:
It seems that the machine test fails because of a segmentation fault, most likely because of accessing some unallocated region of memory? 
```
test-machine_default (SEGFAULT)
```

stavros11:
Hi Giuseppe, thank you for your comments. Sorry for the compiling errors. I thought I had a working version before pulling because at least in my local computer I was able to successfully run the machines with several different settings and even run the unit tests (but got the errors I wrote above). I was probably doing something differently because all my computers have Windows and nothing is the same as the instructions.
Anyway, I have been working on the feedback from Travis and I hope to fix this soon so that you have a working version for review. I try to minimize the commits (I hope github is not spamming everyone with emails!)

NetKet/Machine/mps_diagonal.hpp
```diff
@@ -0,0 +1,743 @@
...
+      c = c.cwiseProduct(W_[site % symperiod_][confindex_[v(site)]]);
+    }
+    return c;
+  };
+
+  T LogVal(const Eigen::VectorXd &v) override {
+    return std::log(mps_contraction(v, 0, N_).sum());
+  };
+
+  T LogVal(const Eigen::VectorXd &v, const LookupType &lt) override {
```
gcarleo: here the compiler complains about the variable ```v``` being unused. The best trick to remove the warning is to do
```
T LogVal(const Eigen::VectorXd & /* v */, const LookupType &lt) ... 
```

stavros11:
Hi Giuseppe, thanks a lot for the comments. I removed the SBS temporarily because of some compiling problems. The rest seem to compile now but still fail the test.
Indeed code duplication is a major issue of my code. Ideally I would like to have one MPS machine and set the diagonal with a boolean option. I've been trying to do that, but there are several functions that are different in the two implementations (still likely possible, though). This will also get rid of the abstract_mps, which I have been using for the SBS calculations.

gcarleo:
Hi Stavros, no worries, the code duplication can be addressed in a second stage... My suggestion was to use a template parameter, but I haven't thought too much about the technical details on how to do that in practice. Anyway, I think that the tests are failing because the random parameters used to check the derivatives are too large (they have unit variance). Thus the derivatives are the limit of numerical accuracy, especially for large networks/large bond dimensions. The solution should be to reduce the variance of the the initialized random parameters in the tests... Maybe you can try to do that and see if things are OK? Thanks!

stavros11:
Travis seems ok with sigma=0.1 in the derivative test. Is there any reason that tests have larger numbers? NNs are typically initialized with small numbers too.

Regarding duplication, I have a version which needs to define only a few functions twice, but it is a bit complicated (using virtual functions and CRTP) and is not completed yet.

gcarleo:
Great! I think that the reason of the large numbers is simply because these functions are highly nonlinear and the output can easily become very large when you have several weights of order 1 (for the MPS it is a slightly different story, but I guess it's the same problem when you have a large number of products of matrices). 

You might want to remove a few remaining warnings during compile time still popping out, otherwise I will have a more careful look at the code in the next days and provide you with more comments!

stavros11:
Sorry, my previous message was not well-written. I just meant if there is any reason that tests have default sigma=1? (since in physical situations we typically use smaller initialization). I would also expect the matrix product to explode with large numbers (not sure if canonical form could fix that).

The warnings were probably because of some redefinitions of functions in AbstractMPS. I removed this to have a cleaner code for a first review. Code between periodic and diagonal is still duplicated. If you prefer I can post a less duplicated version, however it might be better to check the current state first, to make sure that basic calculations etc. are okay (these will be the same in any implementation anyway).

The reason for the redefinitions is the `confindex_` mapping from Hilbert states (`double`) to state indices (`int`). I am not sure if [in the previous discussions](https://github.com/netket/netket/pull/57#issuecomment-417106487) you finally found a solution better than the `map` that I use here (or the `unordered_map`). The idea is that if we use the MPS machines for SBS calculations, it is more efficient to do this mapping once for every site in the SBS class and not in each MPS, since one site can belong to multiple strings. Thus I had to redefine almost every MPS function to take `std::vector<int> &v` instead of `VectorXd &v`. I guess we can decide later if we would like to include SBS and how.

NetKet/Machine/mps_diagonal.hpp
```diff
@@ -0,0 +1,471 @@
...
+  int Npar() const override { return npar_; }
+
+  VectorType GetParameters() override {
+    int k = 0;
+    VectorType pars(npar_);
+
+    for (int site = 0; site < symperiod_; site++) {
+      for (int spin = 0; spin < d_; spin++) {
+        for (int i = 0; i < D_; i++) {
+          pars(k) = W_[site][spin](i);
```
gcarleo: Eigen has also a (reasonably mature) [Tensor type](http://eigen.tuxfamily.org/index.php?title=Tensor_support), maybe it might be convenient to work with that to store the W?

NetKet/Machine/mps_diagonal.hpp
```diff
@@ -0,0 +1,471 @@
...
+        }
+      }
+    }
+  }
+
+  void InitRandomPars(int seed, double sigma) override {
+    VectorType pars(npar_);
+
+    netket::RandomGaussian(pars, seed, sigma);
+    SetParametersIdentity(pars);
```
gcarleo: Why is this necessary? Can't we initialize the matrices with random unitary matrices for example?

NetKet/Machine/mps_diagonal.hpp
```diff
@@ -0,0 +1,471 @@
...
+    int site = tochange[sorted_ind[0]];
+
+    // Update left (site++)
+    if (site == 0) {
+      lt.V(0) = W_[0][confindex_[newconf[sorted_ind[0]]]];
+    } else {
+      lt.V(2 * site) =
+          lt.V(2 * (site - 1))
+              .cwiseProduct(
+                  W_[site % symperiod_][confindex_[newconf[sorted_ind[0]]]]);
```
gcarleo: I think that an elegant and compact way to avoid code duplication is to do something like 
```c++


template<bool B> void MulCwiseOrFull(MatrixType & m1,MatrixType& m2){
  m1*=m2; 
}
template<> void MulCwiseOrFull<true>(MatrixType & m1,MatrixType& m2){
  m1.cwiseProduct(m2); 
}

template<bool IsDiagMps> class Mps{

..... 

void UpdateLookup(...){
  Multiply<IsDiagMps>(M1,M2);
}


}; 
````

File: Test/Machine/machine_input_tests.hpp [hunk 5]
```diff
@@ -8,7 +8,7 @@ std::vector<netket::json> GetMachineInputs() {
   std::vector<netket::json> input_tests;
   netket::json pars;
 
-  // Ising 1d
+  /** // Ising 1d
```
gcarleo: Can you also add a Tutorial file for the MPS on some model? Say AKLT, Ising or Heisenberg... ? That would be nice!

stavros11:
Thanks for the comments. Fixing duplication was much simpler than I thought (having `W_` as `VectorType` in diagonal was complicating things).
Regarding initialization, the idea was to have the matrices close to identity in the beginning. Random initialization works for small systems but fails the tests so it might be less stable. Unitaries seem to work, but I need to improve the way that I am sampling them a bit.

gcarleo:
Hi, compilation was failing because you cannot specialize template functions within a class. I have fixed it, I will push it in a second.

gcarleo:
My remaining concern is mostly about the functions taking Eigen vectors as arguments. 
We should stick to these guidelines [here](https://eigen.tuxfamily.org/dox/TopicFunctionTakingEigenTypes.html), since now I have the impression that the prod and trace methods are far from being optimized !

gcarleo:
Alright, with the last changes I have done to the MPS class I believe that the prod and trace functions are near optimal (maybe one can still avoid some temporaries though with further thinking...). 
The overall speed looks pretty decent to me! Sure, it would be great to have the reduction trees to have faster look-up tables updates... I have an old implementation in one of my codes, it's a bit ugly though, let me know if you would like to have a look at that. Otherwise we can leave this issue for future pull requests, no problem.

stavros11:
Thanks a lot for fixing this. Strangely the first problematic code was compiling and running tests and tutorials in my local pc (which is why I pulled :( ). I did not know the details regarding Eigen, though, thanks!

I guess the tutorial speed is decent because D is quite small. I was planning to try implementing the trees soon. The code could be useful, but I can also talk to people here if there is a problem, so no worries.

stavros11:
Short update: I have a tree look up which seems to work for Ising/Heisenberg but fails the LogValDiff test. I also think the implementation can be improved for the cases where the binary representation of N has many 1s. Unfortunately, this and next week are a bit busy with other things but I will get back to this soon.

gcarleo:
Thank you @stavros11 ! Did you manage to solve the problems with the testing you had then? 
Do you see some speed up with this tree implementation?

gcarleo:
It seems one can quite comfortably do bond dimension 48, which was the maximum doable at the time of the [first MPS sampling paper](https://arxiv.org/pdf/0708.2232.pdf). Does it compare well in terms of speed to the other codes you have available in your group?

stavros11:
Yes, I did it from scratch based on a different idea and tests work now. 
I don't know any concrete way to run benchmarks, apart from the obvious "run and time". Doing this, I indeed find improvement when D is large, eg. four iterations of 1D Heisenberg with N=80, D=50 took 12 mins with the current and 24 mins with the previous code (in my local pc). For small systems I do not see any difference, the older might actually be faster there.
It seems that in the paper they do some additional steps (eg. diagonalization after every update to enforce invariance in spin inversion) which could make it slower but help to converge in less steps. 
My implementation is not based on other code, so it might be possible to optimize it further. But I will try to find something similar in the group to compare speed.

gcarleo:
Great, thanks Stavros, I think we are close to merging this. We will also need some documentation on this new machine though. This is added including files in the repository for the website. Let me know if you need help to do that.

stavros11:
I wrote [some preliminary documentation](https://github.com/stavros11/Variational-Monte-Carlo/blob/master/cpp/mps_periodic_documentation.md) following the files that are related to machines in the website repository. Is this the required file?

stavros11:
I wrote [some preliminary documentation](https://github.com/stavros11/Variational-Monte-Carlo/blob/master/cpp/mps_periodic_documentation.md) following the files related to machines in the website repository. Is this the only required file?

gcarleo:
Thanks @stavros11 , that's exactly the kind of documentation we need. Can you open a pull request on here https://github.com/netket/netket-website to include the documentation? Thanks for your efforts!

gcarleo:
Thank you @stavros11 , that's solid work. I am now merging this one. Since we are now working on the python bindings (I will add python bindings also for your new machine), the documentation will need some adjustments, but it shouldn't be too much work.

PR #75, opened 2018-10-12
Add exact sampler, mostly for testing purposes on small systems

gcarleo:


kchoo1118:
This is good.
I think in the same spirit we should have something to do exact gradient descent and computing observable exactly?

PR #77, opened 2018-10-29
Adding cholesky decomposition as possible choice for exact solver

fabienalet:
In the Stochastic Reconfiguration part of the variational Monte Carlo, this PR adds the possibility to use Cholesky decomposition in lieu of QR by adding UseCholesky = True in the input parameters.
In my tests, this considerably speeds up computations when using the exact solver.
Currently UseCholesky = False by default, one could set it to True after more tests if you agree.

Ideally, one would like to speed up the iterative solver part too, but I did not manage to do so yet.

### DIFF
--- a/NetKet/GroundState/variational_montecarlo.hpp
+++ b/NetKet/GroundState/variational_montecarlo.hpp
@@ -82,6 +82,8 @@ class VariationalMonteCarlo {
 
   bool dosr_;
 
+  bool use_cholesky_;
+
   int nsamples_;
   int nsamples_node_;
   int ninitsamples_;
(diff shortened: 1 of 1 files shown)


gcarleo:
Great, thanks @fabienalet! 

Indeed it's a small change but it can make quite a huge difference when the SR times dominate over sampling times! (I see a lot of difference in the Ising models for example, where sampling is very fast). I have now set the option true by default, I think there is no good reason to stick to the other version, since we typically work with positive definite matrices S. 
Have you also tried using directly the rectangular matrix ? 
[Here](https://eigen.tuxfamily.org/dox/group__DenseDecompositionBenchmark.html) they have some interesting benchmark.

fabienalet:
Re using rectangular matrix: yes this is indeed what we should do.
As a first step, we should try to use [this](https://eigen.tuxfamily.org/dox/classEigen_1_1LeastSquaresConjugateGradient.html) for the iterative part, but I did not succeed yet in implementing it correctly.
Ideally, one could also use an external solver (by experience, performances are massively improved)

gcarleo:
Yes the problem with the iterative solver for the rectangular part is that MPI communications are a bit harder to implement in that case... but it should be possible to do something reasonably efficient. The other issue with these implementations is that they are not thought for parallel applications in the first place, thus all nodes execute the same code, and once in a while they can reach different stopping conditions because of floating point errors... it's a mess. Maybe it would be easier to implement a simple MPI-based conjugate-gradient ourself, or rely on a third-party parallel solver

PR #78, opened 2018-10-31
Acceptance and Ising2d 

gcarleo:
Add acceptance to the output of variaitonal monte carlo. 
Add 2d Ising model tutorial.

### DIFF
--- a/NetKet/GroundState/variational_montecarlo.hpp
+++ b/NetKet/GroundState/variational_montecarlo.hpp
@@ -376,6 +376,8 @@ class VariationalMonteCarlo {
     // Note: This has to be called in all MPI processes, because converting the
     // ObsManager to JSON performs a MPI reduction.
     auto obs_data = json(obsmanager_);
+    obs_data["Acceptance"] = sampler_.Acceptance();
+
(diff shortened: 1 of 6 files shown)


NetKet/GroundState/variational_montecarlo.hpp
```diff
@@ -376,6 +376,8 @@ class VariationalMonteCarlo {
     // Note: This has to be called in all MPI processes, because converting the
     // ObsManager to JSON performs a MPI reduction.
     auto obs_data = json(obsmanager_);
+    obs_data["Acceptance"] = sampler_.Acceptance();
```
femtobit: Is it necessary to call `Acceptance` on all MPI processes? If not, we could move it to the block below the `if` so it only gets called for rank 0.

gcarleo: Yes, the acceptance is also computed as an average over different MPI processes, thus there is a reduction operation also in that line...

femtobit: Okay, I see.

PR #80, opened 2018-11-02
WIP: Python Bindings 

gcarleo:
This is a work-in-progress PR for Python Bindings, using [pybind11](https://github.com/pybind/pybind11), and addresses issue #8 

It is possible to compile NetKet and install the python library just doing 

```bash
pip3 install netket/  
```
This will automatically call CMake, and also register the python library. 

At the moment, this PR provides bindings for `Graph`, `Hilbert`, `Hamiltonian`, `SparseMatrixWrapper` classes. Eventually, all major classes will have Python bindings. 

In `Tutorials/PyNetKet` there is an example usage. For example one can do things like 

```python
import pynetket as nk
from mpi4py import MPI

g=nk.Graph("Hypercube",L=10,Dimension=1)
print(g.Distances(1))
print(g.IsBipartite())
print(g.IsConnected())
```

The beauty of pybind11 is that all methods are exported using this piece of code
```c++
PYBIND11_MODULE(pynetket, m) {
  py::class_<Graph>(m, "Graph")
      .def(py::init<std::string, py::kwargs>())
      .def("Nsites", &Graph::Nsites)
      .def("AdjacencyList", &Graph::AdjacencyList)
      .def("SymmetryTable", &Graph::SymmetryTable)
      .def("EdgeColors", &Graph::EdgeColors)
      .def("IsBipartite", &Graph::IsBipartite)
      .def("IsConnected", &Graph::IsConnected)
      .def("Distances", &Graph::Distances)
      .def("AllDistances", &Graph::AllDistances);
}
```

The rest is just some changes in the constructors of the c++ classes to take `**kwargs` in addition to Json objects. 

Still to do (help/suggestions welcome):

- Pybind11 is added as a git subtree, instead of using the current strategy of using `ExternalProject_Add` in the CMake. I wasn't able to make the latter work, and decided to use a much easier solution based on subtrees. However, if there is a way to make it work as an external project that would be nice. 

- Unit tests are missing for the python part. What's a good strategy? 

- Provide bindings for Hamiltonian, Machine, Sampler...etc 

- What to do with random number generators? Can we bind them? 

Important remarks: 
- This PR changes the APIs for `Hilbert` and `Hamiltonian`. In particular, it is not possible to deduce the `Hilbert` from `Hamiltonian` anymore, and `Hilbert` must be specified in the input file. This change is necessary to avoid unnecessary complications in the maintenance. It also makes sense to have to specify explicitly the Hilbert space. 
- All glue classes now use `std::unique_ptr`, removing the previous use of `std::share_ptr`, see also #42


twesterhout:
First of all, sorry for being non-responsive for quite some time -- I got somewhat overwhelmed by other projects.
@gcarleo I've created a PR showing how you could use `ExternalProjectAdd` with pybind11. We still copy pybind's `FindPythonLibsNet.cmake` file, but it's not project specific and getting it right yourself is quite tricky, so I hope it's okay.

I like the way you've approached the `nlohmann::json` vs `pybind11::kwargs` problem. However, a byproduct of it is that now `netket` (executable) itself depends on pybind11. It's unfortunate and also quite unnecessary since `netket`'s code doesn't really need python. So my suggestion would be to create a single `Utils/kwargs.hpp`:
```cpp
template <class Parameters>
bool FieldExists(Parameters const& pars, const std::string& field);

template <class Parameters>
void CheckFieldExists(const Parameters& pars, const std::string& field,
                      const std::string& context = "");

namespace detail {
template <class Value, class Parameters>
struct FieldValImpl;
} // namespace detail

template <class Value, class Parameters>
Value FieldVal(const Parameters& pars, const std::string& field,
               const std::string& context = "") {
  return detail::FieldValImpl<Value, Parameters>{}(pars, field, context);
}

template <class Parameters>
auto FieldVal(const Parameters& pars, const std::string& field,
              const std::string& context = "") -> typename detail::FieldValImpl<void, Parameters> {
  return detail::FieldValImpl<Value, Parameters>{}(pars, field, context);
}

template <class Parameters>
void FieldArray(const Parameters& pars, const std::string& field,
                std::vector<int>& arr, const std::string& context = "");

namespace detail {
template <class Value, class Parameters>
struct FieldOrDefaultValImpl;
} // namespace detail

template <class Value, class Parameters>
Value FieldOrDefaultVal(const Parameters& pars, const std::string& field,
                        Value defval)
{
  return detail::FieldOrDefaultValImpl<Value, Parameters>{}(pars, field,
                                                            std::move(defval));
}
```
Now, all files with classes which should be constructed from `nlohmann::json` and/or `pybind::kwargs` can include `kwargs.hpp` file. `python_helper.hpp` and `json_helper.hpp` files then define the declared functions and are only included in `netket.cc` and `pynetket.cc` respectively. Then `netket` doesn't depend on json and `pynetket` doesn't depend on pybind11.
I've started playing around with the implementation of this idea, but would like to get some feedback before submitting a PR.

P.S.: If you're wondering why I've declared the `detail::*Impl` structs, the reason is that there's currently no partial template specialisation for functions (and to my knowledge there won't be any in the near future). But don't worry, using structs in this way has zero performance overhead and is standard practice in (advanced) C++ code.

gcarleo:
Absolutely @twesterhout , I like the solution you are proposing. I agree that it would be a good idea to factor out the dependency on python. Indeed we can maybe even try to put include guards for conditional compiling, even though that might be unnecessary.  I am now finishing the work on the `Hamiltonian` bindings, hopefully these shouldn't collide with your PR, if you have it already ready for pushing. Otherwise I will be done with this by tonight, I guess.  

The thing that would be really nice is to construct NetKet derived classes via a single NetKet specific parameter object (say `netket::Parameters`). Then we should have a mechanism to convert back and forth from JSON to `Parameters` and back and forth from `kwargs` and `Parameters`. 

```c++
class Hypercube: public AbstractGraph{
public:
   //Have a single constructor here 
   Hypercube(const Parameters& pars){
   }
};
```

And then do somewhere 
```c++
Parameters Kwargs2Parameters(const pybind11::kwargs &){
  ... 
}

Parameters Json2Parameters(const nlohmann::json &){
  ... 
}
```

Basically the `Parameters` I have in mind is very similar to the (truly excellent) `nlohmann::json` type (a dict with arbitrary types), and we don't need to reinvent the wheel. In principle one could just do `using netket::Parameters=nlohmann::json`, however getting direct type conversions between `kwargs` and `nlhohmann::json` seems tricky to implement (at least for me).

twesterhout:
[Here's](https://github.com/gcarleo/netket/pull/2) an implementation of the idea briefly explained above.

While implementing the above PR I've noticed that this PR's approach to handling arguments does not scale. I.e. one'd like to write
```py
hilbert = Hilbert(...) # Constructing netket::Hilbert using normal kwargs
graph = Graph(...) # Constructing netket::Graph kwargs
hamiltonian = Hamiltonian(..., hilbert=hilbert, graph=graph) # Trouble!
```
So we also want overloads of constructors of types such as `netket::Hamiltonian` accepting hilbert spaces and graphs rather than plain json/dict objects. If we do that, we can then just ask `pybind11` to convert the arguments to correct types, i.e. no need to dirty your hands with `pybind11::kwargs`.

The linked PR works and even has some nice features that `FieldVal(pars, "name")` returns a `json` or `pybind11::dict` object rather than failing to compile. However, I'm afraid that we're doing something fundamentally wrong.

P.S. This approach with constructors taking "normal" arguments also makes my life a whole lot simpler when implementing benchmarks :)

gcarleo:
Hi @twesterhout , thanks, I appreciate the nice compile-time features you put in `FieldVal`! 

Can you give an example of how you would ask `pybind11` to convert constructor types? Just to see if that would scale to other constructors we have. While doing preliminary work on binding `Machine`, I realized it is a mess with all the constructors we have. 

This abundance of constructors is mostly there because of how everything was conceived at the beginning (i.e. run everything with JSON input, and do not assemble yourself the classes to build your application). Since we are moving towards a more library-style approach, maybe we should also reconsider all these convolved constructors in favor of simpler "normal" arguments, as you mention. 
On the other hand we have to balance the reliance on `pybind11`, having a way to quickly change to another library if needed.

twesterhout:
First of all, I realise now that I haven't really explained what I meant by "normal" arguments. Well done @gcarleo understanding me anyway! [Here's](https://github.com/twesterhout/netket/blob/speed-up-compilation/NetKet/Graph/hypercube.hpp#L54) the proposed transformation applied to `Graph/hypercube.hpp` (it's done in the context of me playing around with separate compilation, but it doesn't matter). Notice that there are no JSON-related constructors. Conversion is done in a [free function](https://github.com/twesterhout/netket/blob/speed-up-compilation/NetKet/Graph/hypercube.hpp#L93).


> Can you give an example of how you would ask pybind11 to convert constructor types? Just to see if that would scale to other constructors we have. While doing preliminary work on binding Machine, I realized it is a mess with all the constructors we have.

I thought about using this feature: https://pybind11.readthedocs.io/en/stable/basics.html#keyword-args and then relying on [overloading](https://pybind11.readthedocs.io/en/stable/classes.html#overloaded-methods) to allow construction from different types.

gcarleo:
Ok, got you! Yes, it makes perfect sense. In this way we have normal constructors from standard types and we rely on free functions for type conversion. If I understand correctly your proposal about using pybind11 type conversion, this would imply that we lose the ability to have Json constructors? 

Would it be possible to use instead c++ parameter packs instead? Something like 

```c++
template<class ... Types,class Partype> void Hypercube(Types ... args, ParType & parameters){
....
}
```

along [these lines](https://www.bfilipek.com/2016/03/nice-c-factory-implementation-2.html)

gcarleo:
I guess that the solution with parameter packs would solve your original scalability problem, and wouldn't need pybind type conversion anymore? Only problem is that in the binding we need to specialize all the possible template parameters in the pack... but still it's maybe a less intrusive solution than using pybind11 type conversions

twesterhout:
>  If I understand correctly your proposal about using pybind11 type conversion, this would imply that we lose the ability to have Json constructors?

No, we don't lost the any ability. It's just modularity: classes such as `netket::CustomGraph` or `netket::Hypercube` don't need json at all. And I'd even suggest to drop `netket::Graph` altogether! Let me explain how things would work in this case.

```cpp
// Utils/json_helpers.hpp

namespace netket {
    NETKET_EXPORT
    std::unique_ptr<AbstractGraph> make_graph(json const& parameters)
} // namespace netket
```

```cpp
// Utils/json_helpers.cc

namespace netket {
    // I'm responsible for determining which useful fields the `parameters` have and
    // calling the corresponding `Hypercube::Hypercube` overload. I'm **not** responsible
    // for sanity checks and error reporting! That's the job of `Hypercube::Hypercube`.
    std::unique_ptr<Hypercube> make_hypercube(json const& parameters) { ... }

    // I'm responsible for determining which useful fields the `parameters` have and
    // calling the corresponding `CustomGraph::CustomGraph` overload. I'm **not** responsible
    // for sanity checks and error reporting! That's the job of `Hypercube::Hypercube`.
    std::unique_ptr<CustomGraph> make_custom_graph(json const& parameters) { ... }

    NETKET_EXPORT
    std::unique_ptr<AbstractGraph> make_graph(json const& parameters)
    {
        // Logic that is currently in netket::Graph::Graph(json const&)
        // i.e. dispatching on graph name and calling the corresponding make_*
        // function
    }
} // namespace netket
```

  * `netket`: `Utils/json_helpers.hpp` is included only in `GroundState/ground_state.hpp` where we really need construction from json.
  * `pynetket`: `pynetket.cc` exports `AbstractGraph`, `CustomGraph`, and `Hypercube` to python with corresponding constructors. And declares that `CustomGraph` and `Hypercube` inherit from `AbstractGraph`. This way users explicitly create object of `CustomGraph` or `Hypercube` type and are free to pass them to other netket class constructors such as `Hamiltonian`.

gcarleo:
Neat, I really like the idea of removing `Graph` etc. and also exposing the finer classes to python. It is not entirely clear to me how you handle the `make_graph` for the `AbstractGraph` case, since you have to deal with `std::unique_ptr<ConcreteClass>` but it is highly likely that it should work :) 

Also, what is exactly `NETKET_EXPORT`??

twesterhout:
> I guess that the solution with parameter packs would solve your original scalability problem, and wouldn't need pybind type conversion anymore? Only problem is that in the binding we need to specialize all the possible template parameters in the pack... but still it's maybe a less intrusive solution than using pybind11 type conversions

I'm afraid I'm failing to see how this would help. JSON object and kwargs seem to be fundamentally different in the sense that kwargs are *flat*. We don't want to write
```py
GroundState(
    Graph={'Edges': ...},
    Hamiltonian={'Name': 'Ising', 'h': 1},
    Machine={'Name': 'RbmSpin', ...}

)
```
And we can't have the following
```py
GroundState(
    Edges=...,
    Name='Ising',
    Name='RbmSpin' # Ooops!
)
```

What we *do* want is
```py
graph = CustomGraph(edges=...)
machine = RbmSpin(...)
hamiltonian = Ising(h=1)

gs = GroundState(
    graph=graph,
    hamiltonian=hamiltonian,
    machine=machine
)
```

twesterhout:
> Neat, I really like the idea of removing Graph etc. and also exposing the finer classes to python. It is not entirely clear to me how you handle the make_graph for the AbstractGraph case, since you have to deal with std::unique_ptr<ConcreteClass> but it is highly likely that it should work :)

But `AbstractGraph` remains abstract in python, it's impossible to construct. So there should be no problem with it. Or am I misunderstanding your question?

> Also, what is exactly NETKET_EXPORT??

It tells the compiler that a particular function should appear in object code. It's required for separate compilation.

NetKet/Hilbert/hilbert.hpp
```diff
@@ -17,65 +17,69 @@
...
+  template <class Ptype>
+  void InitWithoutGraph(const Ptype &pars) {
+    int size = FieldVal<int>(pars, "Size");
 
-    if (FieldExists(pars["Hilbert"], "Name")) {
-      std::set<std::string> hilberts = {"Spin", "Boson", "Qubit"};
+    json gpars;
+    gpars["Graph"]["Name"] = "Custom";
+    gpars["Graph"]["Size"] = size;
+    Graph graph(gpars);
```
twesterhout: Am I missing something or does this result in a memory leak? Concrete classes inheriting from `abstract_hilbert` usually store a reference to the given `Graph`. This reference becomes dangling when `graph` goes out of scope.

gcarleo:
OK, I agree that what we want is 
```python
graph = CustomGraph(edges=...)
machine = RbmSpin(...)
hamiltonian = Ising(h=1)

gs = GroundState(
    graph=graph,
    hamiltonian=hamiltonian,
    machine=machine
)
```

I am all in for your solution, please provide a working version for `Graph` and then we can move from there. In case I can do the heavy lifting for the rest of the code... 

Two notes:

1. For the moment being, just ignore the complication in the constructor of `Graph`
```c++
else {
      if (FieldExists(pars, "Hilbert")) {
        int size = FieldVal<int>(pars["Hilbert"], "Size", "Graph");
        json parsg;
        parsg["Graph"]["Name"] = "Custom";
        parsg["Graph"]["Size"] = size;
        g_ = netket::make_unique<CustomGraph>(parsg["Graph"]);
      } else {
        std::stringstream s;
        s << "Unknown Graph type";
        throw InvalidInputError(s.str());
      }
    }
```
is unnecessary and too convoluted. 
2. I believe that in the CMakeLists we should have
```
add_dependencies(netket eigen_project pybind11_project)
```
otherwise the current version doesn't work (See also unit tests)

NetKet/Hilbert/hilbert.hpp
gcarleo: Yes, this is a problem, and it is again related to the complex constructor logic we want to get rid of (see also point 1 in the comment I just posted). We can get rid of this using free `MakeGraphAndHilbert` functions or similar

twesterhout:
> I am all in for your solution, please provide a working version for Graph and then we can move from there. In case I can do the heavy lifting for the rest of the code...

@gcarleo Are you sure you want to do that? ;) I have a working (I think) version, but I've spent last few hours fighting small compilation errors in unit tests. That's one of the reasons I'm working on separate compilation. Give me a few more hours and I'll come back with a PR and a small write-up on the changes.

twesterhout:
> Give me a few more hours and I'll come back with a PR and a small write-up on the changes.

@gcarleo [here](https://github.com/gcarleo/netket/pull/3) you are. Commit messages should bring some (not a lot though, sorry) structure to the changes.

gcarleo:
I am closing this PR, since #81 makes it obsolete

PR #81, opened 2018-11-07
NetKet version 2.0 

gcarleo:
# Summary 
This is a work-in-progress PR for NetKet version 2.0, also functioning as a way to gather some discussion [**see end of this document to give your feedback**] . 

## Major changes

1.  NetKet now fully exposes its internal types and classes, thus becoming a full-fledged python library built on a monolithic c++ core 
2. Python Bindings are provided using [pybind11](https://github.com/pybind/pybind11), thus fully addressing issue #8   
3. As a result of this transformation, JSON input is not accepted anymore and the `netket` executable will not exist anymore 
4. Unit tests check c++ classes without relying on JSON input, unit tests on python classes are done similarly 

### Reasons behind these changes
1. The most common current usage pattern of NetKet is not to write directly the JSON file, but use the python scripts we have in the `Tutorials/` folder. That's why moving to a (much more usable and flexible) python library is a desirable goal, which shouldn't have a dramatic impact on the current usage pattern. 
2. With a full-fledged python library it is also much easier to submit multiple jobs with different parameters, and have more flexibility on the output 
3. There is a substantial overhead for maintaining the JSON support. Supporting all the possible ways of putting together the NetKet classes, in a consistent way, does not scale well given the increasing size of this project. 
4. It will be easier to interface to a whole spectrum of Python-only libraries 
5. New generations of students (unfortunately) are not very familiar with c++, and strive to have a pure Python library 

# Compiling and Examples 

It is possible to compile NetKet and install the python library just doing 

```bash
pip3 install netket/  
```
This will automatically call CMake, and also register the python library. 

At the moment, this PR provides bindings for the great majority of `NetKet` classes. 

In `Tutorials/PyNetKet` there is an example usage. For example one can run a full ground-state optimization like this: 

```python
from __future__ import print_function
import netket as nk
from mpi4py import MPI

#Constructing a 1d lattice
g=nk.Hypercube(L=20,ndim=1)

hi=nk.Spin(S=0.5,graph=g)

#Hamiltonian
ha=nk.Ising(h=1.0,hilbert=hi)

#Machine
ma=nk.RbmSpin(hilbert=hi,alpha=1)
ma.InitRandomPars(seed=1234,sigma=0.01)

#Sampler
sa=nk.MetropolisLocal(machine=ma)

#Optimizer
op=nk.Sgd(learning_rate=0.1)

#Variational Monte Carlo
gs=nk.Vmc(hamiltonian=ha,sampler=sa,
          optimizer=op,nsamples=1000,
          niter_opt=300,output_file='test',
          diag_shift=0.1)
gs.Run()


```

### Still to do (help/suggestions welcome)

- Provide bindings for `ImaginaryTimePropagation`, `TimeEvolutionDriver` etc. We need to define a convenient container class for those. This is needed in order to nicely handle observables, for example as done in `VariationalMonteCarlo` through the method `AddObservable` 

- New unit tests are needed, since JSON input is not supported anymore. We need tests for both the c++ internal classes and the python libraries. 

- Provide a better installer 

- Update documentation 

- Update python tutorials 

- Remove classes and methods marked with `TODO remove`

### Other remarks

- Version 2.0 removes all "glue" classes, such as `Graph`,`Hamiltonian` etc. The corresponding abstract interfaces remain, with little changes. 

- We introduce the concept of `Operator`, which generalizes `Hamiltonian`,`Observable` etc. `Hamiltonian` and `Observable` disappear. 

- We introduce a `LocalOperator`, and all associated overloaded operations, for example 
```python
hi=nk.Spin(S=0.5,graph=g)
X=[[0,1],[1,0]]
o=nk.LocalOperator(hi,X,[0])*(nk.LocalOperator(hi,X,[1]))
```
defines a quantum operator X(0)*X(1), automatically performing the tensor product. Other operations such as product times a scalar and addition of local operators are also supported.  

- Do we need to rethink the way we write the output results? Now results are written to a text file, as it used to be, in JSON format. Given the python support, we could decide to support more output formats. 

# Action required for FEEDBACK
Please comment/ provide your opinion about these changes, and if you believe they would improve/deteriorate your current/expected usage pattern. We welcome comments both from developers and regular users.


gcarleo:
Feedback appreciated - @kchoo1118 @twesterhout @femtobit @GTorlai @jamesETsmith @emilyjd  @ShHsLin @stavros11 @everthemore @fabienalet @alexwie @theveniaut @artemborin @noamwies @alexandercbooth

gcarleo:
@kchoo1118 , I completed the bindings for layers and feed-forward networks, however I find a very strange segmentation fault for the convolutional layer, in the existing unit tests (the `unit_layer` tests). I left a note in there where I see the seg fault emerging, maybe you can understand better what's going on? 

It seems it should be some JSON-related issue, so in principle we don't care for version 2.0. But it's good to see if that is the case or if it is something more serious. On the other hand, the python version works just fine: 

```python
from __future__ import print_function
import pynetket as nk
from mpi4py import MPI
import scipy.sparse as sparse

L=20

#Constructing a 1d lattice
g=nk.Hypercube(L=L,ndim=1)

hi=nk.Spin(S=0.5,total_sz=0,graph=g)

#Hamiltonian
ha=nk.Heisenberg(hilbert=hi)

#Layers
act=nk.Lncosh()
layers=[nk.Convolutional(graph=g,
                         activation=act,
                         input_channels=1,
                         output_channels=4,
                         distance=4)]

#FFNN Machine
ma=nk.FFNN(hi,layers)
ma.InitRandomPars(seed=1234,sigma=0.1)


#Sampler
sa=nk.MetropolisHamiltonian(machine=ma,hamiltonian=ha)

#Optimizer
op=nk.Sgd(learning_rate=0.01)

#Variational Monte Carlo
gs=nk.Vmc(hamiltonian=ha,sampler=sa,
          optimizer=op,nsamples=1000,
          niter_opt=300,output_file='test',
          diag_shift=0.01)
gs.Run()
```

jamesETsmith:
@gcarleo This looks great! You mentioned a better installer, did you have something in mind? I built/installed it without issue on my cluster and things appear to be running smoothly. If you'd still like help with the Tutorials, I'd be happy to help. I can start with the GraphHamiltonians, Ising, J1J2, and BoseHubbard if you'd like, just let me know. Same goes for the unit tests if you'd like help I can get to work on updating those over the next few days.

gcarleo:
@jamesETsmith thanks, glad you like this! 

Yes, I meant packaging the pip installer, and registering on the server. But I am not en expert of this, we can maybe think about this once we finalize this PR, it's not a top priority!  

It'd be great to have some help on anything including Tutorials and unit testing. I am now stabilizing the python interface, I will be done by the end of this week.  Can you hold on for a few more days (say next monday) to start the changes to the tutorials you mention? 
Also, @twesterhout is going to do some work on this next week too (maybe you might want to coordinate!) 

Thanks a lot, really appreciated!

gcarleo:
@kchoo1118 I found it... my fault, the problem was here 
```c++
else if (pars["Name"] == "Convolutional") {
     Ptype(new Convolutional<T>(graph, activation_, pars));
```

jamesETsmith:
@twesterhout I can start working on the python unit tests, I was thinking of Hamiltonian, Machine, and Hilbert. Does that sound good to you? After those are done, I can help out with whichever we have left. 

@gcarleo I'll start working on those tutorials after we have the unit tests wrapped up. If that works for you. If you would rather have them done sooner, just let me know.

gcarleo:
@jamesETsmith sounds good to me. Just be careful with Machine (especially) since there might be a few methods whose bindings are still missing... if you feel brave, you might try to add them yourself in the super-ugly pynetket.cc file, otherwise you can stick to testing the methods already present in the bindings! Good luck

femtobit:
> I had some trouble with the bindings of the classes written by @femtobit and @alexwie , because of the way Observables are handled in there. Ideally, we would like to have a method AddObservable (to be bound in Python), as much as we have it for VariationalMonteCarlo. I would really appreciate if @femtobit or @alexwie could start looking into this issue!

@gcarleo I will have a look at adding the bindings you mention. I am at a workshop this week, but maybe I will have some time in between sessions. Otherwise, I should get to it next week.

While I have been somewhat busy with other projects and did not get to look at the changes in detail, let me say that I am in general very happy with the new Python interface and I think primarily supporting this interface to interact with netket is the right way moving forward. I also agree that this removes the need for the JSON input feature (in fact, if at some point someone needs support for specifying parameters in a config file, this could be easily added again via the Python interface).

twesterhout:
First of all, @gcarleo, could you, please, change the target branch to something like `v2.0`? We can the immediately merge `gcarleo:v2.0` into `netket:v2.0` such that all the following work on this PR goes through the `netket/netket` repo rather than `gcarleo/netket`. Conversations become visible to everyone etc.

twesterhout:
> @twesterhout I can start working on the python unit tests, I was thinking of Operator, Machine, and Hilbert. Does that sound good to you? After those are done, I can help out with whichever we have left.

I think there's one question we need to resolve first: naming conventions.

Consider e.g. the `Nsites` member function that many classes in NetKet implement. Although many C++-people (me being among them) might disagree with Google's Style Guide, it's a perfectly valid style guide used in huge code bases, so no issue here. Python, however, is a very different programming language and it has a pretty universal style guide called [PEP8](https://www.python.org/dev/peps/pep-0008/). I think it's a good idea to use it. Here's the part relevant for the `Nsites` member function:
> Use the function naming rules: lowercase with words separated by underscores as necessary to improve readability.

In other words, our C++-y `Nsites` should probably become something like `number_sites`, `num_sites` or `n_sites` in Python. And then there's a question whether number of sites should actually be a read-only property rather than a member function (I'm voting for read-only property).

Similar concerns hold for other member functions and keyword arguments. AFAIK, in Python, people would rather write `Hypercube(length=4)` or `Hypercube(len=4)` than `Hypercube(L=4)`.

If we don't want to re-write the tests and tutorial later, it's probably a good idea to discuss the names right now.

P.S. backward-compatibility with JSON makes no sense since JSON input is not accepted anymore and docs have to be updated anyway.

gcarleo:
@twesterhout , I just create a new branch for v2.0. Let's move the discussion there

PR #82, opened 2018-11-12
NetKet V2.0b1

gcarleo:
# Summary for version 2.0  
This PR introduces the first beta version for  NetKet 2.0. 

## Major changes in version 2.0 

1.  NetKet now fully exposes its internal types and classes, thus becoming a full-fledged python library built on a monolithic c++ core 
2. Python Bindings are provided using [pybind11](https://github.com/pybind/pybind11), thus fully addressing issue #8   
3. As a result of this transformation, JSON input is not accepted anymore and the `netket` executable does not exist anymore 
4. A great deal of flexibility in the usage of NetKet comes with these changes, removing the majority of constraints coming from the JSON-based input and usage pattern 
5. Bindings are provided for the great majority of `NetKet` classes, with the exception of some internals, that don't need python exposure.  

### Reasons behind these changes
1. The most common current usage pattern of NetKet is not to write directly the JSON file, but use the python scripts we have in the `Tutorials/` folder. That's why moving to a python library is a desirable goal, improving the user experience and opening the way to a number of applications previously hard or impossible to perform  
2. With a full-fledged python library it is also much easier to submit multiple jobs with different parameters, and have more flexibility on the output 
3. There is a substantial overhead for maintaining the JSON support. Supporting all the possible ways of putting together the NetKet classes, in a consistent way, does not scale well given the increasing size of this project. 
4. It becomes easier to interface to a whole spectrum of Python-only libraries, including advanced visualization tools and state-of-the-art machine learning libraries (pytorch, tensorflow etc) 
5. New generations of students (unfortunately?) are not very familiar with c++, and strive to have a pure Python library 

# Installing, Tutorials, Examples 
To install the latest development version do 
```shell
pip install . 
```

The directory `/Tutorials/` contains pynotebook tutorials. More will be added. 
Several example codes showcasing a specific application are contained in `/Examples`.

### Other remarks

- Version 2.0 removes all "glue" classes, such as `Graph`,`Hamiltonian` etc. The corresponding abstract interfaces remain, with little changes. 

- We introduce the concept of `Operator`, which generalizes `Hamiltonian`,`Observable` etc. `Hamiltonian` and `Observable` disappear.

- We introduce a `LocalOperator`, and all associated overloaded operations, for example 
```python
hi=nk.Spin(s=0.5,graph=g)
X=[[0,1],[1,0]]
o=nk.LocalOperator(hi,X,[0])*(nk.LocalOperator(hi,X,[1]))
```
defines a quantum operator X(0)*X(1), automatically performing the tensor product. Other operations such as product times a scalar and addition of local operators are also supported.


gcarleo:
@femtobit thank you for looking into this ! 

@twesterhout, concerning your comments:  

> In other words, our C++-y Nsites should probably become something like number_sites, num_sites or n_sites in Python. And then there's a question whether number of sites should actually be a read-only property rather than a member function (I'm voting for read-only property).
> 
> Similar concerns hold for other member functions and keyword arguments. AFAIK, in Python, people would rather write Hypercube(length=4) or Hypercube(len=4) than Hypercube(L=4).

I believe you are absolutely right. We should stick to the PEP8 style. Indeed what I did is an incomplete pythonization of the names, see
```python
hi=nk.Spin(S=0.5,total_sz=0,graph=g)
```
for example. 

I have put it in the summary up here that we should stick to this more consistently.  
We should also consider enforcing this automatically. For the moment, if any of you starts working on this soon, please also try to use consistent names for python variables. The best thing would be to push new commits changing name variables as soon as they become available.

jamesETsmith:
@gcarleo It looks like there aren't currently any bindings for `Hilbert::RandomVals()`, should I create these or skip the test involving them for now?

gcarleo:
> @gcarleo It looks like there aren't currently any bindings for Hilbert::RandomVals(), should I create these or skip the test involving them for now?
@jamesETsmith , indeed it would be good to add those bindings. The problem was that in turn this requires to bind the c++ random generator as well since
```c++
void RandomVals(Eigen::VectorXd &state,
                              netket::default_random_engine &rgen)
```
where `netket::default_random_engine` is defined in `Utils/random_utils.hpp` 

Unles pybind11 already converts from numpy random generator to the c++ random generator, but I don't think so

gcarleo:
Ok @jamesETsmith  let me quickly add that

gcarleo:
@jamesETsmith , it's done. You can now do
```python
#Random states
rg=nk.RandomEngine(seed=1234)
conf=np.zeros(h.Size())
h.RandomVals(conf,rg)
```

I also added a few more methods to Machine, but I am not binding yet anything related to the LookUp  type. For the moment you can skip the tests involving that...

jamesETsmith:
@gcarleo That was fast, thanks!

gcarleo:
@jamesETsmith , since you are there, you might also start changing the name of the python functions (as declared in PyNetket.cc) to follow these rules https://www.python.org/dev/peps/pep-0008/#function-and-variable-names ... :) 

Basically 
1. Class names should normally use the CapWords (CamelCase) convention we already have 
2. But, function names should be lowercase, with words separated by underscores as necessary to improve readability
3. and variable names follow the same convention as function names

jamesETsmith:
@gcarleo Sounds good, I'll start changing the wrapped function names (and their arguments) over to conform with the PEP-8 style.

femtobit:
Since we are already renaming things, I would suggest also splitting the Python bindings into submodules., since everything directly under `netket` will get more and more confusing over time (or even already now).

The grouping could be mostly done along the lines of the existing subfolders in the C++ code (e.g., `netket.operator.Hamiltonian`, `netket.dynamics.TimeStepper`, or `netket.machine.Rbm`)

For conciseness, we could also discuss adding some convenience functions at the top level, e.g.,  something like
```
import netket as nk

graph = Hypercube(16, 1)
hamiltonian, hilbert = nk.create_system('Ising', graph, J=1)
```

gcarleo:
Agreed for separating things into modules. I already put a TODO comment about this in the monstrous  pynetket.cc file :) 
We could define those convenience functions maybe later, I would prioritize modularization now

jamesETsmith:
Should we partition before we write the Tutorials and Python unit tests so we don't have to back track and rewrite parts of those files?

gcarleo:
> Should we partition before we write the Tutorials and Python unit tests so we don't have to back track and rewrite parts of those files?
@jamesETsmith , yes. It's just that I am not sure how to do that in pybind11, I was looking into the documentation at some point and didn't find the relevant info

gcarleo:
OK, @jamesETsmith , @femtobit . I found a way which also saves some typing. 
See `NetKet/Hilbert/pyhilbert.hpp` and notice that in `NetKet/pynetket.cc` there is a line 
```c++
  AddHilbertModule(m);
```

In the hilbert headers I also define a convenience macro that allows to spare some work after and it is also more maintenable. 

```c++
#define ADDHILBERTMETHODS(name)                \
                                               \
  .def("is_discrete", &name::IsDiscrete)       \
      .def("local_size", &name::LocalSize)     \
      .def("size", &name::Size)                \
      .def("local_states", &name::LocalStates) \
      .def("random_vals", &name ::RandomVals)  \
      .def("update_conf", &name::UpdateConf)   \
      .def("get_graph", &name::GetGraph)
```

Are we all set now to propagate these changes to the other classes? I am really off for the next few days now :)

jamesETsmith:
@gcarleo Thanks for figuring this out! I can start extending these changes to the other classes.

femtobit:
I have pushed a few commits which add bindings for the imaginary time propagation and update the corresponding tutorial.
The interface can probably still be polished a bit, but the basic functionality is there.

twesterhout:
Can we perhaps stop pushing to this branch? Let's create PRs instead. This uncontrollable evolution is kind of scary :)

femtobit:
> Can we perhaps stop pushing to this branch? Let's create PRs instead. This uncontrollable evolution is kind of scary :)

@twesterhout For moving code around and adding the Python submodules, I'd say it wouldn't make too much sense to create new PRs at this point.

For more fundamental changes, however, I agree that additional PRs targeting this branch would help making things easier to follow and discuss.

twesterhout:
> I can have a go at improving the graph-related part today in the evening/night as it acts as a guinea pig anyway :)

#85 is the first change I propose. I appreciate to some may find the code with lambdas inside lambdas etc. somewhat tricky to read, but it's quite concise and gets the job done. By following #84 we get rid of the huge and buggy `Init()` function (buggy because it doesn't handle corner cases correctly). So the C++ core is simplified. By using iterators we also (possibly) avoid some copies which is always nice.

Feedback highly appreciated.

gcarleo:
Hey folks, I was sleeping! 
Just one quick thing: I see that now none of the tests is compiling/working. What happened exactly? 
I agree with @twesterhout that PRs ARE the way of proceeding, unless it's for simples changes like moving around names/classes!

gcarleo:
> @gcarleo Yes, very much I agree with that. Note that I did not use any of these in the constructors. The functions using kwargs are isolated from the rest of the C++ code and meant as helpers to make something like nk.create_timestepper(dim, 'Dopri54', rel_tol=1e-12) possible (since different time steppers take different kinds of arguments).

@femtobit  alright, I think it's actually not a bad idea to have a different class per timestepper, but let's discuss this when at least things are COMPILING again here :)

gcarleo:
@femtobit , it seems that around commit a266d11 the tests compilation was broken

gcarleo:
It seems there is still some issue, sorry 
```
/Users/travis/build/netket/netket/NetKet/Operator/MatrixWrapper/abstract_matrix_wrapper.hpp:50:44: error: no member named 'adjoint' in 'netket::Hamiltonian'
    const std::complex<double> var = state.adjoint() * state2;
```
and related

NetKet/Graph/custom_graph.hpp
```diff
@@ -155,12 +284,41 @@ class CustomGraph : public AbstractGraph {
         }
       }
     }
+  }
+
+  /// Checks that for each edge (i, j): 0 <= i <= j and returns max(j) + 1, i.e.
+  /// the number of nodes
+  int CheckEdges() const {
```
gcarleo: Comment on the fact that this also returns the number of sites

NetKet/Graph/custom_graph.hpp
```diff
@@ -44,73 +67,175 @@ class CustomGraph : public AbstractGraph {
...
+          std::vector<std::vector<int>>(),
+      std::vector<std::vector<int>> edgecolors =
+          std::vector<std::vector<int>>(),
+      bool isbipartite = false)
+      : adjlist_(adjl), isbipartite_(isbipartite) {
+    Init(size, edges, automorphisms, edgecolors);
+  }
+#endif
+
+  CustomGraph(std::vector<Edge> edges, ColorMap colors = ColorMap(),
```
gcarleo: It seems we cannot construct this anymore just passing CustomGraph(size=10) say. 
The only unit tests in python we have are failing. `Test/Graph/test_graph.py`.

gcarleo: We need to enable these tests at least, otherwise it's going to be a disaster in the following with all these changes

jamesETsmith: I have partially complete python tests for Hilbert and Operator, do you want me to push those now and then finish them up as we go along?

gcarleo:
Yes please we need more python tests to contain code anarchy :) do you know
how to enable the python tests in Travis though? I think you already did
that for the other PR, right?

jamesETsmith:
Correct, I appended `Test/CMakeLists.txt` with the following: 

```cmake
file(GLOB PY_TEST_SOURCES
  nktools/*.py
  )

# Make tests for all python files
foreach(_file ${PY_TEST_SOURCES})
  # Strip filename
  get_filename_component(_file_basename ${_file} NAME_WE)
  string(REGEX REPLACE "unit-(.+)" "python-test-\\1" testcase ${_file_basename})

  # Add tests
  add_test(
    NAME ${testcase}
    COMMAND python ${_file}
    )

endforeach()
```

I can make similar changes when I push the tests I mentioned above in a few minutes after a bit of cleanup.

gcarleo:
Thank you @jamesETsmith that's excellent. I think in our case we just need 
`
pytest Test/
`
(and maybe make sure pytest is installed on travis machines...)

jamesETsmith:
@gcarleo, you're right, and this makes things a lot cleaner!

NetKet/Graph/custom_graph.hpp
twesterhout: @gcarleo you mean the reason for it? Quite simple really: if you're only given a list of edges how do you figure out how many nodes the graph contains? The answer is by going through the edges. So why do it twice when you can obtain the number of sites for free from `CheckEdges`?

NetKet/Graph/custom_graph.hpp
twesterhout: @gcarleo Indeed we can't. I don't think such graphs make much sense though or do they?

gcarleo:
You want to be able to define a graph without structure. So basically just
nodes and no edges... We need that for example if you have a custom
hamiltonian which is not naturally defined on any graph... Maybe you can
still accommodate size using a single constructor.lets think about this

jamesETsmith:
@gcarleo As a heads up, I opted to call pytest on each individual file because I found that lumping them all together made it harder to distinguish between them when we run them through ctest.

gcarleo:
@jamesETsmith I guess we need now to change Travis to install netket as pip install netket/ or something...

jamesETsmith:
@gcarleo you're right, I'll add that to the travis file. Should we stop running the C++ unit tests and just stick to the Python ones?

gcarleo:
yes please! it seems that c++ tests are broken at the moment. Eventually all the c++ tests we have will be replaced by corresponding python tests, so this is not a problem if we have a full coverage relatively soon…

gcarleo:
@jamesETsmith you forgot to remove `mkdir -p build && cd build` I think !

femtobit:
>  yes please! it seems that c++ tests are broken at the moment. Eventually all the c++ tests we have will be replaced by corresponding python tests, so this is not a problem if we have a full coverage relatively soon…

In the meantime, I fixed the compilation of the C++ just in case we still need them.

@twesterhout The graph tests fail now (I also commented out a part that no longer compiles). Could you check whether there is a real error or whether the tests cases are just outdated after your changes?

Also, GCC warns about the default arguments of your lambdas in `pygraph.hpp`. As far as I understand, these are [disallowed in C++11](https://stackoverflow.com/a/6281715) (though that changed with C++14).

gcarleo:
I don't like that we don't have a sanitizer check anymore, maybe we should change the CMake options in the setup.py to have options for debugging

gcarleo:
But otherwise, @jamesETsmith @femtobit , it's great to have green lights in Travis again! Thanks! We just need to systematically increase coverage.. little by little we'll get this thing done

twesterhout:
> The graph tests fail now (I also commented out a part that no longer compiles). Could you check whether there is a real error or whether the test cases are just outdated after your changes?

I haven't had a look at C++ tests yet, but will do so... Thanks for the heads up!

> Also, GCC warns about the default arguments of your lambdas in pygraph.hpp. As far as I understand, these are disallowed in C++11 (though that changed with C++14).

Right you are! I'll fix it.

jamesETsmith:
I just pushed the bindings for the HilbertIndex class. I couldn't find a way to bind the MaxStates attribute (max_states in python) to the HilbertIndex class where we could access it without needlessly creating an instance of the class in python (which is what we do in C++). For now, I've bound it to netket module. 

https://github.com/netket/netket/blob/195ac27f69381e1517a4053535d22c3dcb696d8d/NetKet/pynetket.cc#L607

This is probably not best practice and it would be better if we kept things as consistent as possible between C++ and Python, so I'd welcome any ideas you all have about it.

gcarleo:
@jamesETsmith , that was a bit trick, it turns out you need to do
```c++
.def_readonly_static("max_states", &HilbertIndex::MaxStates);
```
(and also declare it before calling the pybind classes...)

gcarleo:
@femtobit , please get the clang-tidy pluging for your editor!

Tutorials/ImaginaryTime/ising1d_imag.py
```diff
@@ -28,13 +28,13 @@
...
 hamiltonian = nk.operator.Ising(hilbert, h=1.0)
-mat = nk.wrap_operator(hamiltonian)
+mat = nk.operator.wrap_as_matrix(hamiltonian)
 
 # create time stepper
 stepper = nk.dynamics.create_timestepper(mat.dimension, rel_tol=1e-10, abs_tol=1e-10)
 
 # prepare output
-output = nk.JsonOutputWriter('test.log', 'test.wf')
+output = nk.output.JsonOutputWriter('test.log', 'test.wf')
```
gcarleo: Thanks for creating these new modules, it makes sense. I think we should also start putting together ideas for how we would like the output now, i.e. if json output is still OK or not.

gcarleo:
The new tests revealed already some problems with the `Machine` bindings. Once in a while I saw happening on tests on the symmetric RBM 
```
E   MemoryError: std::bad_alloc
```

gcarleo:
OK, we are now back to the standard PR policy we have for the master branch. Thanks for the fun!

Tutorials/ImaginaryTime/ising1d_imag.py
femtobit: @gcarleo Agreed. Let's create an issue to discuss this there.

femtobit:
> @femtobit , please get the clang-tidy plugin for your editor!

I think I ran `clang-format -style=google` on all files I modified, except for one oversight in `pyoperator.hpp`, if that is what you mean. Or did I miss something else?

gcarleo:
> I think I ran clang-format -style=google on all files I modified, except for one oversight in pyoperator.hpp, if that is what you mean. Or did I miss something else?

Yes, it's hard to remember to apply clang-format on all files, that's why I prefer to use a plugin that does that automatically every time I save the c++ files. I was also looking at having clang-tidy run automatically after each push, but didn't find a good solution at the time

NetKet/Graph/hypercube.hpp
```diff
@@ -46,33 +50,268 @@ class Hypercube : public AbstractGraph {
...
+                                                      int const n_dim, bool pbc,
+                                                      int const n_sites) {
+    if (!pbc) return {};
+    assert(L >= 2 && n_dim >= 1 &&
+           "Bug! This should be the case by construction.");
+    std::vector<std::vector<int>> symm_table;
+    symm_table.reserve(static_cast<std::size_t>(n_dim * (L - 1)));
+    std::vector<int> coord(static_cast<std::size_t>(n_dim));
+    for (auto dim = 0; dim < n_dim; ++dim) {
+      for (auto rotate_by = 1; rotate_by < L; ++rotate_by) {
```
gcarleo: @twesterhout , here you should include the identity, otherwise this gives wrong results for symmetric machines.
I.e. ```rotate_by=0;...``` 
I am correcting this now

gcarleo: In fact overall this generates wrong/less symmetries, especially in higher dimension. I am reverting to the old version

fabienalet:
Hi everyone,
thanks for the great work on v2.0. I hope I will find some time soon to test all this and provide feedback.

> New generations of students (unfortunately) are not very familiar with c++, and strive to have a pure Python library

I'm (unfortunately) not belonging to this category and rather have the following point of worry (since I'm not used to pybind11): would any contribution to the c++ core require a python binding, and python test units? If yes, a small edit in [CONTRIBUTING.md](https://github.com/netket/netket/blob/master/CONTRIBUTING.md) pointing towards the best way to do this would be appreciated.

gcarleo:
Hi @fabienalet , indeed hopefully at some point we will have a stable 2.0 ready to try out, and it would be invaluable to have some testing there! 

> I'm (unfortunately) not belonging to this category and rather have the following point of worry (since I'm not used to pybind11): would any contribution to the c++ core require a python binding, and python test units? If yes, a small edit in CONTRIBUTING.md pointing towards the best way to do this would be appreciated.

Just think of python as an easier way to construct c++ objects :) 
But anyway, not necessarily all c++ functions will have bindings in python. Most low-levels things will not be bound in python, and for those we will still require c++ testing. `Machine`, `Sampler` (including your samplers!) etc are higher level interfaces and are exposed to python. For those it makes sense to have unit tests directly in python, because they also check the underlying c++ functionality. 
But it's a good point, we will need to update the CONTRIBUTING file with all this information. Doing unit tests in python is relatively easy actually, it's just a bunch of `assert(...)`.

gcarleo:
Hi all, this has been another busy week of work on the version 2.0. Thanks all for your contributions! 

Here is a status update on the main remaining tasks before we have a first release candidate version (thus a stable user interface):

1. Separate remaining python modules (for `Optimizer`, `Sampler` etc) 
2. Write python unit tests for the uncovered classes, mostly translating from python 
3. Rewrite all the tutorials using the python library  
4. Provide a nice wrapper class for the exact diagonalization 

Any volunteer to take some of these for the next week? :)

jamesETsmith:
I can rewrite the tutorials and, depending on how long they take, I'll try to help out with some of the other TODOS.

gcarleo:
Great, thanks @jamesETsmith !

gcarleo:
I have create a project [here ](https://github.com/netket/netket/projects/2) so that we can more easily track the progress and see the tasks people are working on! 

If you start working on an open task please immediately do a PR referencing the open issue (this can just be a "clone" PR, with little or no changes to start with), so that we know who has started working on what!

gcarleo:
Hi @fabienalet 

> (...) netket/External/pybind11/include/pybind11/detail/common.h(473): error: user-defined literal operator not found
template <size_t... IPrev, size_t I, bool B, bool... Bs> struct select_indices_impl<index_sequence<IPrev...>, I, B, Bs...>

Is this with the latest master version? This looks like the compiler not using c++11 standard?

fabienalet:
@gcarleo 
> Is this with the latest master version? This looks like the compiler not using c++11 standard?

Yes, latest master. Well both compilers I use certainly do (and pass the CMake test on this). I've even added -std=c++11 in CPPFLAGS and that did not help.

In fact, when I don't ask for lapack (i.e. removing -DUSE_LAPACK=ON), it goes through and compiles fine ! I don't know if this is due to the lapack (MKL in my case) or the environment I used, but this could be a problem for other users wishing to use optimized lapack with netket... 
Hopefully there is a simple way to fix this.

fabienalet:
No worries, I was just trying to follow up the incredible pace!...

gcarleo:
> No worries, I was just trying to follow up the incredible pace!... 
Great! @fabienalet , maybe you can try to figure out what the problem is with Lapack support from Eigen. It's a bit strange that it won't compile... Otherwise I'll look into this once we get closer to release

gcarleo:
@jamesETsmith @ooreilly could you also please update `Docs/README.md` with the correct procedure to follow to generate the docs?

jamesETsmith:
@gcarleo  do you mind if I push directly to this branch?

gcarleo:
> @gcarleo do you mind if I push directly to this branch?

@jamesETsmith this is a protected branch, you can't push directly, sorry :)

gcarleo:
@femtobit it seems that the [tutorial](netket/Tutorials/PyNetKet/observable.py) on observables does not print the custom observable, do you know what happened?

femtobit:
> @femtobit it seems that the [tutorial](netket/Tutorials/PyNetKet/observable.py) on observables does not print the custom observable, do you know what happened?

Ah, good catch. Yes, I see what's wrong, just a missing call to `ComputeObservables`. I will make a PR in a moment (and I should also add a test that would catch this problem in the future).

gcarleo:
Hi all, I think we are approaching the end of our journey into version 2.0

This has been a time of intense work on the code, and I would like to thank you all for your efforts. 
I think that, overall, we can be more than happy of what this version has become, and that we are now in good shape for a public release of the first beta version. 

Here is the plan to make this as smooth as possible:  
1. Merge this PR into master (hopefully later today) 

2. At that point I would recommend that (if possible) all of us try to install it in a clean environment (see here https://packaging.python.org/guides/installing-using-pip-and-virtualenv/) just doing 
```shell
pip install netket 
```
I have already registered the latest version (2.0b1) on the pip server, so you should already be able to test its installation using the command above.  

3. If everything goes smoothly, the new website with the new Docs and Tutorials will go live (maybe around Monday or Tuesday), and version 2.0b1 will be officially public. 

4. At that point we will work on possible issues and if necessary release 2.0bX versions, otherwise go for 2.0. We just need a few more Tutorials, since at the moment we only have two (we have plenty of Examples, but we also need full Tutorials!)  

Thanks all! 

( @kchoo1118 @femtobit @jamesETsmith @twesterhout @GTorlai @ooreilly @ShHsLin @everthemore @martamau @fabienalet @stavros11 @alexwie @emilyjd @theveniaut @cmendl )

gcarleo:
One more approval to go @jamesETsmith @femtobit @twesterhout

fabienalet:
Hi all,
thanks @gcarleo for all this!
I'd be happy helping on tutorials as I mentioned earlier, however I'm fighting hard to get a pip installation working on my Mac with conda. 
Is there anyone successful in installing correctly+using this version of netket within conda ?
Any tips appreciated, thanks!

jamesETsmith:
@gcarleo thanks for leading the charge on this! Great work everyone!

gcarleo:
@fabienalet , it is unfortunately a known issue we haven't fixed yet. Conda installs don't work on Mac Os, most likely because of some MPI shared library problem. Actually, could you please open an issue reporting exactly your problem, so that we can focus on this? 

As an alternative, you can install netket without using Conda, and creating a clean environment as the one I was suggesting. Does this work? https://packaging.python.org/guides/installing-using-pip-and-virtualenv/

gcarleo:
@jamesETsmith could you please check the pip install netket on a clean Mac Os enviroment (no Conda) ? thanks

jamesETsmith:
@gcarleo, sorry I didn't check this more carefully earlier.  `pip install .` works on on both my mac (using the native python) and on  the redhat linux cluster my group uses, but on each configuration `pip install netket` crashes . For both systems I created a virtualenv, as you suggested. Both display a very similar problem: 

On my mac:
```bash
(env) ~ $ /usr/bin/python -m pip install netket --user
Collecting netket
  Using cached https://files.pythonhosted.org/packages/65/b1/a7a198d18043076bdbc65323d28f28db1ccc0e3ae1cff411e963b4d08b0d/netket-2.0b1.tar.gz
Requirement already satisfied: cmake in ./Library/Python/2.7/lib/python/site-packages (from netket)
Requirement already satisfied: numpy in /Library/Python/2.7/site-packages (from netket)
Building wheels for collected packages: netket
  Running setup.py bdist_wheel for netket ... error
  Complete output from command /usr/bin/python -u -c "import setuptools, tokenize;__file__='/private/var/folders/rw/dzv8pd9n2ng7sdj7b8cf9xjw0000gn/T/pip-build-Na30dG/netket/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" bdist_wheel -d /var/folders/rw/dzv8pd9n2ng7sdj7b8cf9xjw0000gn/T/tmpAFX4GQpip-wheel- --python-tag cp27:
  running bdist_wheel
  running build
  running build_ext
  warning: CMakeBuild: CMake Error: The source directory "/private/var/folders/rw/dzv8pd9n2ng7sdj7b8cf9xjw0000gn/T/pip-build-Na30dG/netket" does not appear to contain CMakeLists.txt.
...
...
```

```bash
(env) ~ $ pip install netket
Collecting netket
  Using cached https://files.pythonhosted.org/packages/65/b1/a7a198d18043076bdbc65323d28f28db1ccc0e3ae1cff411e963b4d08b0d/netket-2.0b1.tar.gz
Requirement already satisfied: cmake in ./env/lib/python3.5/site-packages (from netket) (3.13.3)
Requirement already satisfied: numpy in ./env/lib/python3.5/site-packages (from netket) (1.16.1)
Building wheels for collected packages: netket
  Building wheel for netket (setup.py) ... error
  Complete output from command /home/jasm3285/env/bin/python -u -c "import setuptools, tokenize;__file__='/rc_scratch/jasm3285/pip-install-gdmd39mi/netket/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" bdist_wheel -d /rc_scratch/jasm3285/pip-wheel-66auz2g3 --python-tag cp35:
  running bdist_wheel
  running build
  running build_ext
  warning: CMakeBuild: CMake Error: The source directory "/rc_scratch/jasm3285/pip-install-gdmd39mi/netket" does not appear to contain CMakeLists.txt.
  Specify --help for usage, or press the help button on the CMake GUI.
  
...
...
```

Using the latter as an example, when I try to find the directory `/rc_scratch/jasm3285/pip-install-gdmd39mi/netket` it doesn't exit.

femtobit:
Thanks everyone and agreed, let's merge this and finish the work on `master`.

femtobit:
Maybe we can discuss the PyPI issues in a separate issue? In any case, this does not seem to be Mac specific. On my linux machine (Ubuntu 18.04) I get the same error.

gcarleo:
Hi , thanks for checking this out. We have a problem when registering the package on the server... the install works just fine locally but this issue with pypi is most likely due to some limitation in the way our setup.py is done. 

Agreed, we can merge this for the moment, since this is a (critical) but unrelated issue

gcarleo:
@fabienalet , @kchoo1118 was anyway telling me that Conda install should work also the mac, provided that you download this latest version from github and do 
```
pip install .
``` 
not pip install netket, for which we still are experiencing issues. Let me know if that works.

fabienalet:
> As an alternative, you can install netket without using Conda, and creating a clean environment as the one I was suggesting. Does this work? https://packaging.python.org/guides/installing-using-pip-and-virtualenv/

Yes, it works this way, thanks!

Now checking  ````pip install .```` within a fresh conda env ...

fabienalet:
@gcarleo @kchoo1118
I didn't get ```pip install .``` to work on a conda environment.
I also get the 
```Fatal Python error: PyThreadState_Get: no current thread```
error, which perhaps is the same thing as pointed out [here]( https://github.com/netket/netket/pull/90#issuecomment-440447341) by @gcarleo

twesterhout:
Glad to see this in master. Thanks everyone!

Issue #84, opened 2018-11-13
Why does each graph store an adjacency list?


twesterhout:
Looking at current master, the only meaningful use case for the adjacency list is DFS. DFS is in turn only used during initialisation, and seeing as it's O(N), it's no big deal to create a temporary adjacency list. That way creation of an adjacency list becomes an internal thing we can test and trust; we don't have to accept it as an argument which simplifies the interface (i.e. no more choosing between `edges` and `adjacency_list`) as well as the code. All of that without touching the hot path.

gcarleo:
Can I close this?

PR #85, opened 2018-11-13
Proposed way to handle graph construction illustrated on CustomGraph

twesterhout:


NetKet/Graph/custom_graph.hpp
```diff
@@ -55,7 +81,41 @@ class CustomGraph : public AbstractGraph {
...
+              std::vector<std::vector<int>> automorphisms =
+                  std::vector<std::vector<int>>(),
+              bool isbipartite = false)
+      : edges_{std::move(edges)},
+        eclist_{std::move(colors)},
+        automorphisms_{std::move(automorphisms)},
+        isbipartite_{isbipartite} {
+    nsites_ = CheckEdges();
+    if (nsites_ == 0) {
+      throw std::invalid_argument{"Empty graphs are not supported."};
```
femtobit: If you throw `netket::InvalidInputError` it will be caught by the code in `netket.cc`.

twesterhout: There's no `netket.cc` anymore -- we're moving to Python. That's why I use `std::invalid_argument` which is correctly converted to `ValueError` by pybind11.

gcarleo:
True, but check if throwing in python has issues with Mpi etc.

NetKet/Graph/custom_graph.hpp
twesterhout: > True, but check if throwing in python has issues with Mpi etc.

@gcarleo could you elaborate? I'm afraid I don't get your point.

gcarleo:
The reason why we had custom exceptions was to handle different mpi
processes. Damian worked on that so he knows better why that was necessary.
I'm just saying that we should how throwing through mpi4py works

femtobit:
> The reason why we had custom exceptions was to handle different mpi
processes. Damian worked on that so he knows better why that was necessary.

The difference is largely cosmetic, we could just as well catch `std::invalid_argument`. We just should agree on the exception type.

If an exception is thrown in cod called from Python now, it will raise a Python exception. Probably that should be caught in the Python code to properly terminate the MPI processes, but I am not sure yet about the best way to do this.

twesterhout:
But cases when exceptions are thrown due to invalid input are user's responsibility! Code that actually does something related to MPI (e.g. Optimizers) is a whole different story and there we do need to be careful. Totally unrelated to this PR though...

NetKet/Graph/pygraph.hpp
```diff
@@ -39,46 +39,176 @@ namespace netket {
...
+/// Given a Python iterable object constructs the edge list and (optionally)
+/// the colour map for the soon to be graph. `callback` is then called in one of
+/// the following ways:
+/// * `callback(edges, colour_map)` if the iterable contained elements of type
+/// `(int, int, int)`.
+/// * `callback(edges)` if the iterable contained elements of type `(int, int)`.
+///
+// TODO(twesterhout): We should probably split this into smaller functions...
+template <class Function>
+auto WithEdges(py::iterable xs, Function&& callback)
```
gcarleo: We have to decide if we have simple (i.e. pybind11 free) constructors or not. Personally I am in favor of NOT having python-based constructors, see also comment to @femtobit . Can we make this using std objects instead?

gcarleo: I see, you put it in `pygraph.hpp` to trick me :) 
Fine, as long as this dependence on pybind in contained in the py* files we can probably do this

twesterhout: Constructors are pybind11-free, are they not? `WithEdges` is a _conversion_ function: it converts a Python iterable to a `std::vector<Edge>` while avoiding going through `std::vector<std::tuple<int, int>>`.

twesterhout: > I see, you put it in `pygraph.hpp` to trick me :)
> Fine, as long as this dependence on pybind in contained in the py* files we can probably do this

Sorry, didn't notice this. Ignore my previous comment then :) Can I mark this as resolved then?

gcarleo: Yes

NetKet/Graph/custom_graph.hpp
```diff
@@ -25,13 +25,37 @@
...
 */
 class CustomGraph : public AbstractGraph {
+ public:
+  using AbstractGraph::Edge;
+  using AbstractGraph::ColorMap;
+
+ private:
   // adjacency list
+  // TODO(twesterhout): To be removed
   std::vector<std::vector<int>> adjlist_;
```
gcarleo: So you want to work only with edges and get the adjlist only if requested? That is fine to me, in fact the two ways of storing the graph are completely equivalent

NetKet/Graph/custom_graph.hpp
femtobit: @twesterhout We can [convert](https://pybind11.readthedocs.io/en/stable/advanced/exceptions.html#registering-custom-translators) `InvalidInputError` to a Python `ValueError`, which we probably should do.
I do not have particularly strong feelings about whether to use custom exceptions or the `std` ones, but it should be consistent throughout the project.

NetKet/Graph/custom_graph.hpp
twesterhout: It's easier to run sanity checks on edges, they require less memory, and (if we want to) we can very easily expose them to Python without copying (I'll show this in the next PR once this one is merged; exposing adjacency lists requires 2D iterators and I'm too lazy to implement one for such a simple task). (Ref. #84)

twesterhout: But from the maths perspective, yes, they are equivalent.

NetKet/Graph/custom_graph.hpp
twesterhout: @femtobit indeed we can.
Using `std` exceptions here is the first step in getting rid of NetKet's custom exceptions entirely since they add no additional info and are non-conformant (i.e. throwing them is most likely UB, but I'm not sure about the UB part).

femtobit: At least the C++ Core guidelines seem to disagree [(see E.14)](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#e14-use-purpose-designed-user-defined-types-as-exceptions-not-built-in-types). (Looking at this, I think it would be better for our custom type to inherit from `std::logic_error`, just as `std::invalid_argument` does, however.) Though I guess this is mostly a stylistic question for us, except of course for your point regarding potential UB.
I'd be interested to hear why you suspect throwing them to be UB and how this should be fixed.

gcarleo: Guys, let's just use some ```netket::whatever_exception``` and define ```using netket::whatever_exception=whatever_we_will_decide```. The natural place do this is in the `Utils/exceptions.hpp`

NetKet/Graph/custom_graph.hpp
gcarleo: I am even fine with re-computing adjacency lists when asked by some GetAdjacency method. This is not performance critic, let's stick to simple implementations

NetKet/Graph/custom_graph.hpp
twesterhout: > At least the C++ Core guidelines seem to disagree (see E.14)

@femtobit touché! The reason it's not really applicable here is that all exceptions are converted to Python analogues anyway... So let's stick to @gcarleo's suggestion.

> I'd be interested to hear why you suspect throwing them to be UB and how this should be fixed.

As per [18.8.2] (taken from C++17 draft I have at hand, but I'm pretty sure this didn't change since C++11):
> Each standard library class T that derives from class exception shall have a publicly accessible copy constructor and a publicly accessible copy assignment operator that do not exit with an exception.

NetKet's exceptions store a `std::string`. `std::string`'s copy constructor and assignment operators allocate memory and can thus throw. That's why exceptions in `std` are usually implemented using COW (copy-on-write) strings.

jamesETsmith:
I think that the colors should be fine as long as we give the `EdgeColorsFromAdj()` one created from some `GetAdjacency()` function.

twesterhout:
> exceptions should be handled by a netket exception (this can be a typedef for std exceptions, if necessary)

@gcarleo done, see the latest commit.

> do colors work or is there some problem?

I see no reason why they wouldn't.

twesterhout:
All in all, can we merge this and I'll move on to updating `Hypercube` and other graph-related methods? Tests are failing anyway :)

gcarleo:
@femtobit , going to merge, is it ok for you too? Exceptions are isolated to the netket definitions

femtobit:
I did not have the time to look into the actual code other than the minor part about the exception, but that point has been addressed.

(And if there is indeed a problem with our exception class, I'd be happy for a pointer in the right direction, so far I do not see it.)

gcarleo:
The reason why tests do not compile is because 
```
/home/travis/build/netket/netket/NetKet/GroundState/imaginary_time.hpp:6:49: fatal error: Dynamics/TimeStepper/time_stepper.hpp: No such file or directory
````
it wouldn't be a bad idea to fix this at some point :)

femtobit:
> it wouldn't be a bad idea to fix this at some point :)

Oh, my bad. I will fix this.

PR #89, opened 2018-11-14
Fix automorphisms for Hypercube and initialization bug in RbmSpinSymm

gcarleo:
Also change python names for machine and add test for log derivatives

### DIFF
--- a/NetKet/Graph/hypercube.hpp
+++ b/NetKet/Graph/hypercube.hpp
@@ -16,8 +16,8 @@
 #define NETKET_HYPERCUBE_HPP
 
 #include <mpi.h>
-#include <array>
 #include <algorithm>
+#include <array>
 #include <cassert>
 #include <map>
 #include <unordered_map>
@@ -38,7 +38,7 @@ class Hypercube : public AbstractGraph {
                      //  which the graph is invariant
   ColorMap colors_;  //< Edge to color mapping
 
-#if 1 // TODO(twesterhout): This is to be removed
+#if 1  // TODO(twesterhout): This is to be removed
   // contains sites coordinates
   std::vector<std::vector<int>> sites_;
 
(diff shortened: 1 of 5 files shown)


NetKet/Graph/hypercube.hpp
```diff
@@ -218,30 +218,43 @@ class Hypercube : public AbstractGraph {
...
-        do {
-          auto &i = coord[dim];
-          auto const old_i = i;
-          i = (i + rotate_by) % L;
-          permutation.push_back(Coord2Site(coord, L));
-          i = old_i;
-        } while (next_variation(coord.rbegin(), coord.rend(), L - 1));
-
-        symm_table.emplace_back(std::move(permutation));
+    if (!pbc) {
```
twesterhout: Indeed, I forgot this case!

NetKet/Graph/hypercube.hpp
```diff
@@ -218,30 +218,43 @@ class Hypercube : public AbstractGraph {
   static std::vector<std::vector<int>> BuildSymmTable(int const L,
                                                       int const n_dim, bool pbc,
                                                       int const n_sites) {
-    if (!pbc) return {};
-    assert(L >= 2 && n_dim >= 1 &&
-           "Bug! This should be the case by construction.");
-    std::vector<std::vector<int>> symm_table;
```
twesterhout: But what's wrong with the rest of the implementation?

gcarleo: I am not sure, but your implementation gives different permtables for square or cubic lattices. I think you are not considering all the possible translation vectors, but I am not sure

twesterhout:
@gcarleo ah, I see what you mean... I also understand where the mistake is, but let's just merge this, because the whole reason for an alternative solution was simplicity and I'm not sure it'll remain just as simple if I fix the bug. Good job noticing this!

As for the Machine part, I'm not sure I like the exposed interface, but I'll complain about that later :)

PR #90, opened 2018-11-15
WIP: Updating setup.py

twesterhout:


gcarleo:
Can you make this PR to the branch v2.0? thanks!

twesterhout:
> Can you make this PR to the branch v2.0? thanks!

Yes, that and around a 100 other fixes :) I'll ping when this PR is ready for review.

twesterhout:
I'm still busy with Python 2 support, but I think I'm done with the "big" changes. So @gcarleo @femtobit @jamesETsmith what do you think?

* Users should be able to do
  ```
  pip install .
  ```
  for default arguments and
  ```
  pip install . --install-option="--cmake-args=-DCMAKE_CXX_COMPILER=my_compiler -DCMAKE_BUILD_TYPE=Debug ..."
  ```
  for custom config options.
* For developers
  ```
  python setup.py build --cmake-args=...
  export LD_LIBRARY_PATH=...
  ```
  so that repeated calls to `build` reuse the available CMake cache.

P.S.: I failed to get setuptools' `develop` to play ball, so `pip install -e . --install-option="--cmake-args=..."` won't work. You can still do `pip install -e .` though.

gcarleo:
Thanks @twesterhout , this looks tricky indeed. Glad you managed to get default cmake options to work! Let's hope that Travis cooperates a bit for python 2 ..

twesterhout:
Okay, so the reason Python 2 doesn't work is because our `.travis.yml` is a mess :) We need to use Python's virtual environment correctly... I'll have a go at fixing it tomorrow, but I have no experience in setting up Python projects on Travis, so if you have any tips or want to take over, do tell.

gcarleo:
@twesterhout , maybe the easiest is to look at how this is handled in the pybind11 repo? They must have a decent travis.yml file in there

twesterhout:
@gcarleo I think I'm getting somewhere with this PR. I'm just struggling with OS X as all computers I have access to run Linux or Windows. So help with `brew` nightmare is highly appreciated (if you know how to fix the errors, feel free to submit a PR to twesterhout/netket targeting this branch).

.travis.yml
```diff
@@ -2,121 +2,264 @@
...
-
-
-cache:
-  apt: true
-  directories:
-    - $HOME/Library/Caches/Homebrew
+    - os: osx
+      osx_image: xcode10.1
+      language: generic
+      env: PYTHON=3.6.0 CXX=clang++
```
gcarleo: At least one of the runs should happen with the sanitizer on, can you put it back here?

setup.py
```diff
@@ -1,60 +1,261 @@
...
+import sys
 
+from distutils import log
 from setuptools import setup, Extension
-from setuptools.command.build_ext import build_ext
-from distutils.version import LooseVersion
+from setuptools.command.build_ext import build_ext as build_ext_orig
+
+# NOTE(twesterhout): distutils and setuptools use a pretty strange way of
+# handling command line arguments, where some arguments are propagated to
```
gcarleo: Yes I realized that...! It is much easier to just pass environment variables, but I agree that it's ugly. It's good if you found a solution here

setup.py
```diff
@@ -65,6 +266,9 @@ def build_extension(self, ext):
     author_email='netket@netket.org',
     license='Apache',
     ext_modules=[CMakeExtension('netket')],
-    cmdclass=dict(build_ext=CMakeBuild),
+    cmdclass={'build_ext': build_ext},
     zip_safe=False,
+    install_requires=[
+        'mpi4py'
```
gcarleo: Good point

.travis.yml
twesterhout: I'd prefer to have OS X build actually working first... But then sure, I'm open to adding however many flags you'd like.

gcarleo:
Notice that the failure of CI here should be solved with #111 , so please let's merge that one too

twesterhout:
> Notice that the failure of CI here should be solved with #111 , so please let's merge that one too

@gcarleo are you sure? Okay, then I'll clean code up a bit and we can merge this (% all the extra flags of course).

twesterhout:
@gcarleo What do you think about [this build](https://travis-ci.org/netket/netket/builds/457485871)? It looks like everything compiles, links and installs fine.

gcarleo:
@twesterhout 

> What do you think about this build? It looks like everything compiles, links and installs fine.

It looks like the two Mac builds are failing there though

twesterhout:
> It looks like the two Mac builds are failing there though

@gcarleo that's what I'm talking about! If you look carefully at the log, you'll notice that the build fails during the execution of `pytest`. Again, as I have no OS X system available locally to investigate the problem more thoroughly, I suspect that it's just due to one of the tests failing. If you experience similar problems locally and can get me more info, then we can try to fix it. Otherwise I'll just clean up the code here and let's hope for the best.

gcarleo:
Can’t you pass —verbose to pytest ?

gcarleo:
there is an abort trap. But it's weird that it doesn't happen also on Linux. 
There was a problem causing abort trap on tests if compiled in debug mode, that I fixed with the PR we just merged. But if you compiled with debug mode also on Linux, you should have had an abort trap also there... weird

gcarleo:
Can you merge from the upstream changes? Just to make sure the tests are not playing a role anymore

femtobit:
@twesterhout Thanks for all the work on `setup.py`!

Before we merge this PR into 'v2.0' (and once no one else needs push to this branch, of course), could you squash the "Go number *n* with Travis" and "Try installing (*n*)"  and similar into one commit each with `git rebase -i`?

I think this would make the git history a lot more readable, since one does not have to wade though about 40 commits concerning setup.py and Travis when looking at the recent changes. Especially considering the `v2.0` PR is already somewhat extensive. :slightly_smiling_face:

twesterhout:
@femtobit you're absolutely right! I was thinking about closing this PR and creating a fresh one with just a few commits. There aren't that many changes from the LOC perspective.

gcarleo:
@twesterhout it's a mess. 
Abort trap 6 is a sort of seg fault error. It's unclear where it comes from and if it's due to some compiling/installing problem or if it comes from the tests themselves. Maybe you can disable a few tests and try again, but this is really painful...

twesterhout:
> Abort trap 6 is a sort of seg fault error. It's unclear where it comes from and if it's due to some compiling/installing problem or if it comes from the tests themselves. Maybe you can disable a few tests and try again, but this is really painful...

@gcarleo are you experiencing it locally as well?

gcarleo:
@twesterhout , no... I don't have these issues on my mac, and all the other builds on mac via Travis were working OK

gcarleo:
Also, there is something weird going on, because a few builds have `USE_LPACK=ON` in the CMake options, but we now that these builds should fail. However this doesn't happen, suggesting that your CMake flags are not actually passed to CMake ?

gcarleo:
Indeed @twesterhout , NETKET_CMAKE_FLAGS are not actually passed to CMake... I tested that locally

gcarleo:
Honestly the approach 
```python 
self.debug = check_env_flag('DEBUG')	
self.sanitizer = check_env_flag('SANITIZER')	
```
that I used before is not THAT bad (actually I got "inspired" by the pytorch repository)

twesterhout:
> Also, there is something weird going on, because a few builds have USE_LPACK=ON in the CMake options, but we now that these builds should fail. However this doesn't happen, suggesting that your CMake flags are not actually passed to CMake ?

True, only a few special arguments were passed (like CXX and CXXFLAGS). This was done on purpose.

> Indeed @twesterhout , NETKET_CMAKE_FLAGS are not actually passed to CMake... I tested that locally

@gcarleo please, have a look at the latest log. `NETKET_CMAKE_FLAGS` are clearly being handled. It works locally as well.

By the way, we have some new info regarding the failures (see Python 2 build on OS X [here](https://travis-ci.org/netket/netket/jobs/457663676)):
```
$ python -c 'import netket'
Fatal Python error: Interpreter not initialized (version mismatch?)
/Users/travis/.travis/job_stages: line 104: 96167 Abort trap: 6           python -c 'import netket'
The command "python -c 'import netket'" exited with 134.
```

gcarleo:
Looks like you are experiencing this https://github.com/caffe2/caffe2/issues/854#issuecomment-364538485

twesterhout:
> Looks like you are experiencing this caffe2/caffe2#854 (comment)

@gcarleo wow! thanks, testing it out :)

twesterhout:
@gcarleo I think we're done here. The last failure is because of `NETKET_Sanitizer=ON` which is not the topic of this PR. So if there are no more comments, I'll create a new clean version of this PR which we can immediately merge.

gcarleo:
@twesterhout OK, glad that solved the Mac Os issue. But what's wrong now with the Sanitizer?! 
I think that the sanitizer for some reason works only under Mac Os, I already had this issue before. You might just want to move the sanitizer on flag for Mac Os environments. Last change than we merge the new PR, OK?

twesterhout:
> But what's wrong now with the Sanitizer?!
I think that the sanitizer for some reason works only under Mac Os, I already had this issue before. You might just want to move the sanitizer on flag for Mac Os environments. Last change than we merge the new PR, OK?

No, I don't think that's it. The reason is that we now have a shared library rather than an executable so we have to correctly link with e.g. `libasan.so` and, what's worse, manage to make it the first library loaded. Locally this can simply be achieved by setting `LD_PRELOAD` to the path of the ASan library, but doing it on Travis is a problem. A bit of CMake magic should do it, but I don't (yet) know how :) So for now I'd much rather disable `NETKET_Sanitizer` in Travis and open a new issue marked "help wanted".

P.S.: Just to be sure, I will add another OS X build with `NETKET_Sanitizer`... It might just magically work, it's OS X after all ;)

jamesETsmith:
Sorry, I'm a little late to the discussion here. This `Abort trap: 6` error has come up when I've tried to build on my local machine, which is a Mac. I've tried adding the `--enable-framework` flag, but this hasn't solved my local issue. Did it fix the problem here?

gcarleo:
You mean that doing locally on your Mac pip install doesn't work/leads to
an abort error?

jamesETsmith:
@gcarleo Correct, I've searched through some of the other errors on stackoverflow/GitHub issues and most people have the issue when they have conflicting versions of python, mine is anaconda. I've tried getting rid of some unused python versions from brew and anaconda. I haven't tried uninstalling it and reinstalling yet because that's a bit of a pain, but if it comes down to it I can try that.

gcarleo:
@twesterhout any progress on this? It seems there is some problem with this setup on Mac, @jamesETsmith ?

twesterhout:
> @twesterhout any progress on this? It seems there is some problem with this setup on Mac, @jamesETsmith ?

Not really, I've been a little busy with other things. AFAIK there shouldn't be any problems with OS X... But if there are, could you please provide a bit more info? Otherwise I think I'll just clean this up and we can merge it.

jamesETsmith:
@gcaleo I'm still having trouble using NetKet on my Mac laptop. I've gone back and made sure that the python I'm using was built with the `--enable-frameworks` flag and still no luck. I believe this is more of a conda issue than a NetKet one. It might make sense to open a separate issue about it

jamesETsmith:
If I use the native version of python from Mac it works fine, so this is definitely a Conda problem.

twesterhout:
I'll see if I can finish this up in the weekend.

PR #91, opened 2018-11-15
Add tests in Hilbert, Operator, remove some obsolete tests

gcarleo:


Test/Operator/test_operator.py
```diff
@@ -34,7 +34,7 @@
 
 def test_produce_elements_in_hilbert():
     for name, ha in operators.items():
-        hi = ha.GetHilbert()
+        hi = ha.get_hilbert()
```
gcarleo: It is still a mistery to me why here get_hilbert works without seg_faulting, whereas in machine.get_hilbert breaks down. Both return references and not pointers.... @jamesETsmith @femtobit @twesterhout

twesterhout: @gcarleo you might want to study the sections on return value policies: https://pybind11.readthedocs.io/en/stable/advanced/functions.html#return-value-policies. They explain stuff way better than I ever could.

gcarleo:
Still seg_faulting :) It's clear that we need to move to pointers or something more reliable...

jamesETsmith:
Is the seg_faulting happening after the changes from #92 ?

gcarleo:
Yes, still happening. The best/safer thing is to switch from references to
smart pointers. I am in the middle of doing that...

twesterhout:
> The best/safer thing is to switch from references to smart pointers. I am in the middle of doing that...

@gcarleo could you, please, share your design first? Before making huge changes to the code base...

gcarleo:
@twesterhout , we keep a smart pointer as a private variable instead of a reference
```c++
const std::shared_ptr<AbstractHilbert> h_;
```
and we have a method 
```c++
const std::shared_ptr<AbstractHilbert> GetHilbert()const {
   return h_;
}
```

The only problem is that these changes propagate in `Sampler` and `Operator`, but this seems inevitable to me, since it's hard to keep working with references and having safe python objects. 
If you have a way of working with references instead of course I'd be happy to hear about it!

gcarleo:
The only thing I really don't like about this is that I need to move from `.` to `->` for calls to hilbert methods. I hate `->`.

twesterhout:
I most certainly object to `const std::shared_ptr<AbstractHilbert> h_`. It's bad practice to use `const` with member data.

gcarleo:
Why?

twesterhout:
@gcarleo as for the "better" solution, I'd suggest passing arguments when we actually need them. If you start using `shared_ptr`s then you can't use references anywhere. And allocating `shared_ptr`s on the stack is a huge pain...

gcarleo:
`Hilbert` is needed by machines, by `Sampler` etc. There is no way to do without it without making things cumbersome... Would you prefer unique pointers ?

gcarleo:
@twesterhout https://google.github.io/styleguide/cppguide.html#Reference_Arguments 
```
there are some instances where using const T* is preferable to const T& for input parameters. For example:

You want to pass in a null pointer.
The function saves a pointer or reference to the input.
Remember that most of the time input parameters are going to be specified as const T&. Using const T* instead communicates to the reader that the input is somehow treated differently. So if you choose const T* rather than const T&, do so for a concrete reason; otherwise it will likely confuse readers by making them look for an explanation that doesn't exist.
```
I think we do have a concrete reason

twesterhout:
> Hilbert is needed by machines, by Sampler etc. There is no way to do without it without making things cumbersome... Would you prefer unique pointers ?

@gcarleo unique pointers mean unique ownership which is not what you want here. Shared pointers mean, you guessed it, shared ownership. What I'm trying to say is that here we have neither! :) Does `Sampler` need `Hilbert` to generate new states? No problem, let's pass it as an argument. Does `Machine` need to know the number of sites? No problem, ask `Hilbert` and pass it as an argument etc.

gcarleo:
> Does Sampler need Hilbert to generate new states? No problem, let's pass it as an argument. Does Machine need to know the number of sites? No problem, ask Hilbert and pass it as an argument etc.

Yes but the question is how do we pass it as an argument? reference or (smart) pointer? :)

twesterhout:
> Yes but the question is how do we pass it as an argument? reference or (smart) pointer? :)

As a const reference, of course. It's used only during one function call, so no lifetime-related issues here.

gcarleo:
The other scenario why we want to store a reference to `Hilbert` is when we want to know on what Hilbert a given `Machine` instance has been generated. That's basically why we need `get_hilbert`. Of course we can do without this, but this leaves too much room for unecessary errors from the user. Also, if I understand correctly your proposal, you want to pass a Hilbert every time I need to generate a new sample in `Sampler`? I am not sure this is very nice !

twesterhout:
> The other scenario why we want to store a reference to Hilbert is when we want to know on what Hilbert a given Machine instance has been generated. That's basically why we need get_hilbert. Of course we can do without this, but this leaves too much room for unecessary errors from the user. Also, if I understand correctly your proposal, you want to pass a Hilbert every time I need to generate a new sample in Sampler? I am not sure this is very nice !

It makes little sense to expose functions creating single samples to Python (unless it's purely for testing, of course). Because if I have go Python->C++->Python inside my Monte Carlo loop, I'll be better of writing pure Python :)

twesterhout:
We want the whole Monte Carlo loop running in C++ and to expose a high level function to Python that calculates the observables  corresponding to the given operators.

twesterhout:
> In any case, your proposal would imply changing all the sampler methods passing a Hilbert const reference... ! But the situation will be very similar also for other objects. 

True.

> Do you want to pass (in c++) Machine, Hilbert, Graph etc every time you call a single method in Sampler? I mean, that's a possibility...

@gcarleo I only ever want to call a single method in sampler -- "run Monte Carlo loop and give me my observables back" :) Why'd anyone need anything more low level? But if someone does, I'd say it makes sense to use C++ directly.
So to answer your question: yes, I think, I do.

Or am I missing something significant here?

gcarleo:
> @gcarleo I only ever want to call a single method in sampler -- "run Monte Carlo loop and give me my observables back" :) Why'd anyone need anything more low level? But if someone does, I'd say it makes sense to use C++ directly.
So to answer your question: yes, I think, I do.

> Or am I missing something significant here?

I understand what you are saying, but if we don't store a reference to the input, then also in c++ we need (for the low-level, c++ calls) give a hilbert,wf etc reference every time we want a sample + we have to make sure we give the good ones we used to construct the sampler with. Isn't it much easier to just store the input instead?

twesterhout:
> I understand what you are saying, but if we don't store a reference to the input, then also in c++ we need (for the low-level, c++ calls) give a hilbert,wf etc reference every time we want a sample + we have to make sure we give the good ones we used to construct the sampler with. 

If you write modern C++ rather than C, then no, you don't :)

> Isn't it much easier to just store the input instead?

That's a bit subjective... The answer is no for me, personally, but I can't say anything about other active contributors.

By the way, about shared pointers and why I prefer not to use them everywhere, have you seen this talk https://channel9.msdn.com/Events/GoingNative/2013/Cpp-Seasoning (I hope that's the right one)?

jamesETsmith:
> If you write modern C++ rather than C, then no, you don't :)

@twesterhout could you elaborate? Maybe with a pseodocode example?

gcarleo:
How do I know what Hilbert has been used to construct a given instance of a Machine?

twesterhout:
> How do I know what Hilbert has been used to construct a given instance of a Machine?

Do I really care? What I probably do care about is how many sites there are in the system according to `hilbert` and whether this number is compatible with `machine`. Also, implementing comparison operators for *Hilbert classes is too much work for too little gain (and no, you can't simply compare addresses).

gcarleo:
I care though, because I don't want to do useless checks every time I call a critical function... !!! 

Also, it might well be modern c++, but I personally advocate the use of simple designs. That design is not simple...  I mean, honestly I am not sure it is good practice to pass all the arguments of the constructor all the times you have to invoke a method of your class. 
This is also an issue if you want to pass an instantiation  of your class to another class method. To be consistent, at that point it wouldn't be enough to pass just a pointer or a reference to your class' object, but you'd need to pack in also all the arguments you used to construct it!!! OMG.

twesterhout:
> I care though, because I don't want to do useless checks every time I call a critical function... !!!

Until a week ago NetKet was mostly used as a standalone executable. You only had one `hilbert` object, one `machine`, one `sampler` etc. Do you expect the workflow to change so significantly that it becomes difficult to keep track, which wave function lives in which Hilbert space? If you do, then sure, let's implement comparison operators for *Hilbert classes.

> I mean, honestly I am not sure it is good practice to pass all the arguments of the constructor all the times you have to invoke a method of your class.
This is also an issue if you want to pass an instantiation of your class to another class method. To be consistent, at that point it wouldn't be enough to pass just a pointer or a reference to your class' object, but you'd need to pack in also all the arguments you used to construct it!!! OMG.

I suspect you slightly misunderstood me. We have a class `Sampler` which stores some configuration parameters like the number of steps to skip during thermolisation, which random number generator to use etc. But it doesn't store the the wave function that will be sampled. It also doesn't store the Hilbert space in which this wave function lives. As a user I don't create this class every time I want to compute the energy of the system. I create it once, just like I create my Hamiltonian once.

Now, this `Sampler` has a method called `sample` or `__call__`. Given the actual wave function that describes the probability distribution and the Hilbert space that describes the possible states, it returns a Markov chain (e.g. a Python iterable or a C++ iterator). I can now iterate through my Markov chain and sample all the observables I want (that's also very similar to the way things are currently done in variational_montecarlo.hpp where the method `Sample()` generates the Markov chain, except that there the whole chain is kept in memory whereas here states are computed on demand).

As for the implementation details, we don't really want to write the whole logic of a sampler in one function. For that, we create a small wrapper class `detail::Sampler`, that stores references to `Sampler`, `Machine`, `Hilbert` etc. It gets the configuration parameters like thermolisation length from the `Sampler` and uses `Machine` and `Hilbert` in much the same way they are currently used in various NetKet samplers. Creation of such a class is a bunch of pointer copies which is very fast. The logic inside the class itself is simple, because it uses references and doesn't care about lifetime-related issues. *If* later on, you have some class `Foo`, where you already store a `shared_ptr<Machine>`, `shared_ptr<Hilbert>` etc. you can safely add `Sampler` and `detail::Sampler` as member variables and it'll be perfectly safe.

Does that make it a bit clearer?

gcarleo:
OK, I'll need to think more about that. 

For the moment I just propose the least action solution, which is simply not to expose the method `GetHilbert` and similar to python. In testing we can pack machine, and the hilbert used to construct it. In that case it is not a big deal...

NetKet/Hilbert/pyhilbert.hpp
```diff
@@ -42,37 +42,37 @@ constexpr int HilbertIndex::MaxStates;
       .def("size", &name::Size)                \
       .def("local_states", &name::LocalStates) \
       .def("random_vals", &name ::RandomVals)  \
-      .def("update_conf", &name::UpdateConf)   \
-      .def("get_graph", &name::GetGraph)
+      .def("update_conf", &name::UpdateConf)
```
gcarleo: I just removed 'get_graph' and 'get_hilbert'

NetKet/Hilbert/pyhilbert.hpp
```diff
@@ -42,37 +42,37 @@ constexpr int HilbertIndex::MaxStates;
...
 
 void AddHilbertModule(py::module &m) {
   auto subm = m.def_submodule("hilbert");
 
-  py::class_<AbstractHilbert>(subm, "Hilbert")
+  py::class_<AbstractHilbert, std::shared_ptr<AbstractHilbert>>(subm, "Hilbert")
```
gcarleo: I am not sure it is necessary to use a shared_ptr handler here

Test/Machine/test_machine.py
```diff
@@ -1,96 +1,111 @@
...
 
 # Hilbert space of spins from given graph
-hi1 = nk.hilbert.Spin(s=0.5, graph=g)
-
-machines["RbmSpin 1d Hypercube spin"] = nk.machine.RbmSpin(hilbert=hi1, alpha=1)
+hi = nk.hilbert.Spin(s=0.5, graph=g)
 
-machines["RbmSpinSymm 1d Hypercube spin"] = nk.machine.RbmSpinSymm(
-    hilbert=hi1, alpha=2)
+machines["RbmSpin 1d Hypercube spin"] = [nk.machine.RbmSpin(
```
gcarleo: Here I pack together machines and respective hilberts. Maybe this can be done in a nicer way in python

Test/Machine/test_machine.py
```diff
@@ -1,96 +1,111 @@
...
+    nk.machine.JastrowSymm(hilbert=hi), hi]
 
-# FFNN Machine
-machines["FFFN 1d Hypercube spin"] = nk.machine.FFNN(hi2, layers)
 
-machines["MPS Diagonal 1d spin"] = nk.machine.MPSPeriodicDiagonal(
-    hi2, bond_dim=8)
-machines["MPS 1d spin"] = nk.machine.MPSPeriodic(hi2, bond_dim=8)
+# TODO FFNN give a seg fault here for some still obscure reason
+# # Layers
```
gcarleo: Here there still is a segmentation fault... it's not clear why, if I had to bet this is related to the layers pointers

Test/Machine/test_machine.py
jamesETsmith: I think this is a fine way to do it.

gcarleo:
@twesterhout what do you think about [this approach](http://www.modernescpp.com/index.php/c-core-guidelines-type-erasure-with-templates) ? 

Basically I like it because we can return an arbitrary `Hilbert` both by reference OR by value, if we want. 

I took the liberty to rewrite it with our names:
 
```c++
#include <iostream>
#include <memory>
#include <string>
#include <vector>

class Hilbert {
 public:
  template <typename T>
  Hilbert(const T& hilb)
      : hilb_ptr(std::make_shared<ModelHilbert<T>>(std::move(hilb))) {}

  std::string GetName() const { return hilb_ptr->GetName(); }

  struct ConceptHilbert {
    virtual ~ConceptHilbert() {}
    virtual std::string GetName() const = 0;
  };

  template <typename T>
  struct ModelHilbert : ConceptHilbert {
    ModelHilbert(const T& t) : hilbert_(t) {}
    std::string GetName() const override { return hilbert_.GetName(); }

   private:
    T hilbert_;
  };

  std::shared_ptr<const ConceptHilbert> hilb_ptr;
};

struct Spin {
  std::string GetName() const { return "Spin"; }
};

struct Boson {
  std::string GetName() const { return "Boson"; }
};

class Machine {
  Hilbert hilb_;

 public:
  Machine(Hilbert hilb) : hilb_(hilb) {}

  void Run() { std::cout << "Machine using " << hilb_.GetName() << std::endl; }
  Hilbert GetHilbert() { return hilb_; }
};

int main() {
  std::cout << std::endl;

  Spin spin;
  Boson boson;

  Machine ma(spin);
  ma.Run();

  // Retrieve a Hilbert object by value
  Hilbert hi = ma.GetHilbert();
  std::cout << hi.GetName() << std::endl;

  // Vector of concrete types
  std::vector<Hilbert> hilberts{spin, boson};
  for (auto hilb : hilberts) {
    std::cout << hilb.GetName() << std::endl;
  }
}
```

gcarleo:
@twesterhout , I am now watching the video by Sean Parent you linked and realized that it is exactly the same approach he discussed there! I like this a lot

twesterhout:
I'm seriously against the first code snippet, because it's a bit buggy. As for the second version -- yes, that's textbook implementation of type erasure. And as much as I like type erasure as a concept I'm afraid that's not what we need here. You can see it from the fact that you're using `unique_ptr` to store the object. It implies ownership. You don't want to own a copy of the Hilbert space, you just want access to it.

So if after considering my proposal with passing arguments by reference and only when we need them you decide not to use it, then using plain old `shared_ptr`is the second best solution. Especially when, apparently, we can trick `shared_ptr` into using custom deleters:
```cpp
Hilbert hilbert{...}; // Allocated on the stack
Machine machine{..., std::shared_ptr<Hilbert>{&hilbert, [](void*){}}};
```
That way we can still allocate objects on the stack when we know that `machine` will not go out of scope later than `hilbert`.

femtobit:
I'd suggest not making things too complicated and use `shared_ptr`s for the Hilbert space for now. This requires only small changes to the current design, which I consider a good thing as there are already many other changes for version 2.0.

The use of `shared_ptr`s should not have a noticeable effect on performance as long as we are not creating or passing them around in a hot loop.

The approach of passing the Hilbert space and other objects to only when they are needed does have some favorable properties, but we can still discuss making these more involved changes at a later time.

gcarleo:
@twesterhout I see, but I actually meant to have a copy of `Hilbert`, since those are very small objects, passed only in constructors. The same goes for the types we use to construct more complicated objects (especially now that you removed some heavy members in `Graph`). So in principle I would say that passing a copy is really not an issue. 

@femtobit, agreed. If we don't want to go for type erasure now, we can try with shared pointers, or just not expose get_hilbert and do without shared pointers. 


If we take this direction, there is really **only one case** at the moment where we still need shared pointers. That is for the `Layer` class. The problem is that a vector of Layers has to be passed to the `FFNN` machine. Not surprisingly, this is also the only machine at the moment causing a segmentation fault in the tests.... I think that working on `Layer` is a good guinea pig. I would really like to pass a vector of generic `Layer` without using shared pointers though... that's why I was looking into type erasure solutions.

gcarleo:
In any case, using type erased objects is actually not a big change in the code base. We already have `abstract_xxx.hpp` and `xxx.hpp`. We basically just need to add the `Model` bit in the abstract file, and do some affordable changed to `xxx.hpp`. Then the rest is just about changing constructors from taking `AbtractXxx` to just `Xxx`, not a big deal. Just saying...

gcarleo:
Alright guys, maybe all of this is just not necessary. 
I found the origin of the seg fault. Basically this is OK: 
```python
act = nk.activation.Lncosh()
layers = [
    nk.layer.FullyConnected(
        input_size=g.n_sites,
        output_size=40,
        activation=act)
]
```
but this is not 
```python
layers = [
    nk.layer.FullyConnected(
        input_size=g.n_sites,
        output_size=40,
        activation=nk.activation.Lncosh())
]
```

it seems that it doesn't like the fact that the activation is not explicitly stored as a pointer somewhere... My headaches are coming back. 

Let me know if you know how to solve this, otherwise for the moment let's just merge this PR!

twesterhout:
> it seems that it doesn't like the fact that the activation is not explicitly stored as a pointer somewhere... My headaches are coming back.

@gcarleo well, of course it still segfaults :) Sorry, but what did you expect? You're [passing a reference](https://github.com/netket/netket/blob/146651923b1d0d9cd6ae1bc16e4af133f8de4a10/NetKet/Machine/pylayer.hpp#L50) to an activation function to the fully-connected layer, and it [stores a reference](https://github.com/netket/netket/blob/c8ba87136b77312ba00375233039923d17ef5835/NetKet/Machine/fullconn_layer.hpp#L36). Exactly the pattern we've been discussing with machines and Hilbert spaces.

gcarleo:
I am merging this now, and keep issue #97 as a reminder that seemingly something is wrong with `Jastrow` and `JastrowSymm`

PR #92, opened 2018-11-15
Fix segmentation fault in machine test

femtobit:
The reason for the segfault seems to have been that the Machine
instances only store references to the Hilbert instances.
Since these were created from Python and stores in the variable `hi`,
they were deleted when `hi` was assigned another Hilbert instance,
in the test code and the machines then contained dangling references.

The return value policy for `Machine.get_hilbert` is `py::return_value_policy::reference`, which I think is correct (and this policy should also be specified for `Operator.get_hilbert`).

The part using `Gradient` in one of the tests is still commented out because `Gradient` does not seem to be exported right now (unless I missed something).

### DIFF
--- a/Test/Machine/test_machine.py
+++ b/Test/Machine/test_machine.py
@@ -10,17 +10,17 @@ machines = {}
 g = nk.graph.Hypercube(length=20, ndim=1)
 
 # Hilbert space of spins from given graph
-hi = nk.hilbert.Spin(s=0.5, graph=g)
+hi1 = nk.hilbert.Spin(s=0.5, graph=g)
 
-machines["RbmSpin 1d Hypercube spin"] = nk.machine.RbmSpin(hilbert=hi, alpha=1)
+machines["RbmSpin 1d Hypercube spin"] = nk.machine.RbmSpin(hilbert=hi1, alpha=1)
 
 machines["RbmSpinSymm 1d Hypercube spin"] = nk.machine.RbmSpinSymm(
-    hilbert=hi, alpha=2)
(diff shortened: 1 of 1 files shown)


femtobit:
If this is indeed the source of the error, we should discuss whether the current approach of having operators and machines store references to the Hilbert objects is a good idea, given that these are created from Python. Having to take care of the lifetime of the Hilbert objects in Python seems counter-intuitive to me.

Maybe we should store an `std::shared_ptr<AbstractHilbert>` in the machine and operator classes? Shared pointers have a runtime performance overhead, but that should not be relevant for the performance-critical parts of our code, as the cost is only incurred when [creating, copying, or destroying the shared pointers](https://stackoverflow.com/questions/22295665/how-much-is-the-overhead-of-smart-pointers-compared-to-normal-pointers-in-c).

twesterhout:
Do machines actually need access to an `AbstractHilbert` object? I've skimmed through a few machines and it seems that the stored references are never used for anything except for returning them from the `GetHilbert()` function. Can we maybe get around with never storing references to `AbstractHilbert`? The idea of reference counting on both Python and C++ side gives me the creeps... But if not, then what you're proposing @femtobit (i.e. using `std::shared_ptr`) is probably the way to go.

gcarleo:
Thank a lot @femtobit , I was really puzzled by this! 
Basically in the current machines we generally need the AbstractHilbert object only in the contructor, but there might be newer machines that need it also for other purposes. 
Using a shared pointer seems to me the only safe solution? Ideally we would like to just pass a copy, since these are small objects, but we can't... unless there's a workaround

PR #93, opened 2018-11-15
WIP: inserting wavelets in RBMSpin

louity:
Hey Giuseppe,
I started the wavelets integration.
The things I'm unsure about it the vector matrix product `W_wavelets_ * Wavelets_.transpose()` or `(W_wavelets_ * W_).row(sf)` where `W_wavelets_` is the vector of  the complex weights or length `nw_` `Wavelets_` is the matrix of the wavelets of shape `nv_, nw_`.

### DIFF
--- a/NetKet/Machine/rbm_spin.hpp
+++ b/NetKet/Machine/rbm_spin.hpp
@@ -37,11 +37,18 @@ class RbmSpin : public AbstractMachine<T> {
   // number of hidden units
   int nh_;
 
+  // number of wavelets
+  int nw_;
+
   // number of parameters
   int npar_;
 
   // weights
   MatrixType W_;
+
+  // predefined wavelet and wavelet weights
+  MatrixType Wavelets_;
+  VectorType W_wavelets_;
 
   // visible units bias
   VectorType a_;
@@ -57,6 +64,9 @@ class RbmSpin : public AbstractMachine<T> {
   bool usea_;
   bool useb_;
 
+  // flag to know wether wavelets are used or not
+  bool usewavelets_;
+
   const Hilbert &hilbert_;
 
  public:
@@ -70,7 +80,22 @@ class RbmSpin : public AbstractMachine<T> {
   }
 
   void Init() {
-    W_.resize(nv_, nh_);
+    if (usewavelets_) {
+      W_wavelets_.resize(nw_)
+
+      Wavelets_.resize(nv_, nw_)
+      // initialize wavelets using gabor wavelets
+      for (int i = 0; i < nv_; i++) {
+        for (int j = 0; j < nw_; j++) {
+            // NOTE: there I need to know if I'm in 1D or 2D
+            Wavelets_(i, j) = 0.;
+            k++;
+        }
+      }
+    } else {
+      W_.resize(nv_, nh_);
+    }
+
     a_.resize(nv_);
     b_.resize(nh_);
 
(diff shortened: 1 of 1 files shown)


NetKet/Machine/rbm_spin.hpp
```diff
@@ -37,12 +37,19 @@ class RbmSpin : public AbstractMachine<T> {
   // number of hidden units
```
gcarleo: I think it's better to create a separate class `RbmWaveletSpin` or something similar

gcarleo: The problem is that we are now in the middle of rewriting most of the `Machine` constructors and we won't support json constructors anymore in version 2.0
I think that the best would be for you to have a look at the branch v2.0 instead of the master, so that you can be sure that the changes you are doing will be compatible with the new version. The interface for machines is relatively stable.

NetKet/Machine/rbm_spin.hpp
```diff
@@ -70,7 +80,22 @@ class RbmSpin : public AbstractMachine<T> {
...
-    W_.resize(nv_, nh_);
+    if (usewavelets_) {
+      W_wavelets_.resize(nw_)
+
+      Wavelets_.resize(nv_, nw_)
+      // initialize wavelets using gabor wavelets
+      for (int i = 0; i < nv_; i++) {
+        for (int j = 0; j < nw_; j++) {
+            // NOTE: there I need to know if I'm in 1D or 2D
+            Wavelets_(i, j) = 0.;
```
gcarleo: I think that the solution is to create a new wavelet class that takes a `Hypercube` object as a constructor. 
`Hypercube basically contains all information about the the number of sites, geometry, periodic boundary conditions etc.

NetKet/Machine/rbm_spin.hpp
```diff
@@ -121,7 +157,12 @@ class RbmSpin : public AbstractMachine<T> {
       lt.V(0).resize(b_.size());
     }
 
-    lt.V(0) = (W_.transpose() * v + b_);
+    if (usewavelets_) {
+      // NOTE: I don't know if the Vector matrix mul is corrrect and if I understood correctly the convolution
+      lt.V(0) = (W_wavelets_ * Wavelets_.transpose() * v + b_);
```
gcarleo: I am not sure either, it depends on what Wavelets_ are exactly here.

NetKet/Machine/rbm_spin.hpp
louity: Ok good to know I'll check out to this branch and do a separate class.

NetKet/Machine/rbm_spin.hpp
louity: Ok I found it https://github.com/netket/netket/blob/v2.0/NetKet/Graph/hypercube.hpp . Is there an example of an existing class that would look like `Wavelets`I could inspire myself from ?

NetKet/Machine/rbm_spin.hpp
louity: `W_wavelets_` is a vector of length `nw_` (= number of wavelets) and represents the multiplicative weights in front of the wavelets.  `Wavelets_` is a matrix of shape `(nv_, nw_)` so `Wavelets_.transpose()` should be a matrix of shape  `(nw_, nv_)`. In the expression  `W_wavelets_ * Wavelets_.transpose()`, the `*` has to be a pointwise product, i.e.  the row `w, 0 <= w < nw_' of the matrix  `Wavelets_.transpose()` is multiplied by `W_wavelets_(w)`. Thus `W_wavelets_ * Wavelets_.transpose()` is a matrix of shape `(nw_, nv_)`. This seems coherent to me since in my understanding, the number of wavelets `nw_` plays the same role a the number of hidden units `nh_` in the convolutional RBM.

PR #100, opened 2018-11-17
Create separate modules for remaining classes

### LINKED ISSUE #96: Create separate python modules for Sampler etc
Move out the remaining bindings still in `pynetket.cc`

gcarleo:
And move Lookup into utils.

fixes #96


gcarleo:
Cool!

PR #104, opened 2018-11-18
Add python tests for sampler and modernize custom sampler, fixes #95

### LINKED ISSUE #95: Add python unit tests for Machine

gcarleo:
fixes #95


Test/Sampler/test_sampler.py
```diff
@@ -0,0 +1,154 @@
...
+import networkx as nx
+import numpy as np
+import pytest
+from mpi4py import MPI
+from pytest import approx
+
+samplers = {}
+
+
+# TESTS FOR SPIN HILBERT
```
gcarleo: @twesterhout these tests apparently give a seg fault on linux machines ONLY. Is this trivial as well?!? ! Anyway, these memory management issues are a mess.

twesterhout: Okay, this one is slightly more complicated, because I had to look through a lot of code, but _yes_, that's again the exact same thing! :) See the other comment for the explanation

Test/Sampler/test_sampler.py
```diff
@@ -0,0 +1,154 @@
...
+sa = nk.sampler.MetropolisLocalPt(machine=ma, n_replicas=4)
+samplers["MetropolisLocalPt Boson"] = [sa, hi, ma]
+
+ma = nk.machine.RbmMultiVal(hilbert=hi, alpha=1)
+ma.init_random_parameters(seed=1234, sigma=0.2)
+sa = nk.sampler.MetropolisLocal(machine=ma)
+samplers["MetropolisLocal Boson MultiVal"] = [sa, hi, ma]
+
+hi = nk.hilbert.Spin(s=0.5, graph=g)
+g = nk.graph.Hypercube(length=6, ndim=1)
```
twesterhout: @gcarleo you construct `hi` with the graph stored in `g` and now you assign a new graph to `g`. There are no more references to the old one and GC can thus collect it (which it does, or if not, you can force it to by doing `import gc; gc.collect()`). You then try to construct a machine which internally asks `hi` for the graph which is now a dangling reference -- and you get a segmentation fault.

P.S.: a word of advice, if I may, try compiling the C++ core with address sanitizer from Clang or GCC, they do a pretty good job at detecting such problems.

gcarleo: Let's see if your theory is correct... :)

gcarleo: Alright your theory is correct, chapeau @twesterhout ! 
Fine, but this reference situation is too error prone, the average user here would get crazy as much as I am getting crazy with these little problems... we need to find a decent solution to pass by value. 

And I agree for the sanitizer, that's why the new python setup file is much welcome ;)

twesterhout: Maybe I'm seriously misreading it, but I though we agreed (at least for now) to follow @femtobit's suggestion and use `std::shared_ptr`s. I kind of assumed that's what you were working on @gcarleo...

gcarleo: But would that solve the problem entirely? That's the question

gcarleo: In the layers case, I am passing a vector of shared pointers to the FFNN class, and I still had a seg fault, if you remember

twesterhout: > But would that solve the problem entirely? That's the question

Yeah, it really should, if done correctly.

> In the layers case, I am passing a vector of shared pointers to the FFNN class, and I still had a seg fault, if you remember

I haven't looked at the code carefully, but I really hope it falls under the "not (yet) implemented correctly" clause :)

gcarleo:
> It's great when within a single commit, 11 files are changed considerably... 

Aha, I can read some light touch of irony there. But you know, the important is the quality of the commit ;) 

I can try to do less changes per commit, let's see if I can...

gcarleo:
By the way, @fabienalet , I just wanted to show you the changes I did to custom sampler. Basically nothing too structural, but some changes were needed to better integrate with local operator. I also added a check that the transition operators should be off-diagonal, otherwise nothing moves!

fabienalet:
Hi @gcarleo, I'll check this in detail tomorrow, but I'm sure everything's OK. Thanks

fabienalet:
Yes, I checked and everything's OK.

PR #105, opened 2018-11-18
Change references of python-bound classes to std::shared_ptr

### LINKED ISSUE #95: Add python unit tests for Machine

### LINKED ISSUE #97: Python bindings for Jastrow machines randomly failing
I realized in #91 that once in a while the python tests for the `Jastrow` machines fail. This issue seems related to some memory problem, but I still don't understand if it is on the c++ side or python

gcarleo:
To help avoiding hard to track down memory management problems in python. 

fixes #95 and fixes #97


NetKet/GroundState/imaginary_time.hpp
```diff
@@ -58,7 +58,8 @@ class ImaginaryTimeDriver {
       ComputeObservables(initial_state);
       auto obs_data = json(obsmanager_);
       output_.WriteLog(step, obs_data, t);
-      output_.WriteState(step, initial_state);
+      // TODO enable this again
+      // output_.WriteState(step, initial_state);
```
gcarleo: This is the offending line

NetKet/Output/json_output_writer.hpp
```diff
@@ -111,13 +111,14 @@ class JsonOutputWriter {
   // The first overload works for classes inheriting from AbstractMachine, the
   // second one for Eigen matrices.
   template <typename T>
-  void SaveState_Impl(std::ofstream& stream, const AbstractMachine<T>& state) {
-    state.Save(stream);
+  void SaveState_Impl(std::ofstream& stream,
+                      std::shared_ptr<AbstractMachine<T>> state) {
```
femtobit: I do not see why it would make sense to pass a shared pointer here. As above, nothing in `SaveState` retains the reference so it is safe to just pass a const reference to the machine.

The only reason to pass a shared pointer to a function (instead of just the underlying raw pointer or, equivalently, a reference) is if the functions requires shared ownership of the object. The output writer does not require this.

NetKet/Output/json_output_writer.hpp
```diff
@@ -96,7 +96,7 @@ class JsonOutputWriter {
    * frequency.
    */
   template <class State>
-  void WriteState(int iteration, const State& state) {
+  void WriteState(int iteration, State state) {
```
femtobit: Why copy the state? Nothing in WriteState retains the reference so it is perfectly fine to just use a const reference.

NetKet/Output/json_output_writer.hpp
```diff
@@ -111,13 +111,14 @@ class JsonOutputWriter {
...
-    state.Save(stream);
+  void SaveState_Impl(std::ofstream& stream,
+                      std::shared_ptr<AbstractMachine<T>> state) {
+    state->Save(stream);
   }
 
   template <typename T, int S1, int S2>
   void SaveState_Impl(std::ofstream& stream,
-                      const Eigen::Matrix<T, S1, S2>& state) {
+                      Eigen::Ref<Eigen::Matrix<T, S1, S2>> state) {
```
femtobit: I think this should be a `Ref<const Eigen::Matrix...>` since the function is not supposed to modify the matrix.

NetKet/GroundState/imaginary_time.hpp
femtobit: See my comments on the changes in `JsonOutputWriter`. Reverting them should fix this.

NetKet/Output/json_output_writer.hpp
femtobit: If this is changed back to take a const reference, you will have to replace `psi_` by `*psi_` when it is called with a machine from VMC:
https://github.com/netket/netket/blob/e90628faed38c6f22ee6691588ae8b84840f210d/NetKet/GroundState/variational_montecarlo.hpp#L401

NetKet/Output/json_output_writer.hpp
gcarleo: Machine are handled and passed as shared pointers throughout the code, to avoid memory management problems in python. I am not sure I can pass a reference to the state instead of a pointer there because this is called from `VariationalMonteCarlo` where machine is stored as a shared pointer

NetKet/Output/json_output_writer.hpp
gcarleo: That's the point, psi_ here is a shared pointer, I can't revert it to a reference

gcarleo: Sorry I misunderstood your comment. I can change that

gcarleo: Should be good now

gcarleo:
@femtobit are you OK with the latest changes I made? Basically I am not touching the JsonOutput anymore

gcarleo:
OK great, thanks !

PR #108, opened 2018-11-19
add activations as layer

### LINKED ISSUE #98: Create new layers for activations only
and remove the `Activation` class

kchoo1118:
The info messages are not yet fixed
closes #98

### DIFF
--- a/NetKet/Machine/activation_layer.hpp
+++ b/NetKet/Machine/activation_layer.hpp
@@ -1,0 +1,159 @@
+// Copyright 2018 The Simons Foundation, Inc. - All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef NETKET_ACTIVATIONLAYER_HH
+#define NETKET_ACTIVATIONLAYER_HH
+
+#include <Eigen/Dense>
+#include <complex>
+#include <fstream>
+#include <random>
+#include <vector>
+#include "Utils/all_utils.hpp"
+#include "Utils/lookup.hpp"
+#include "abstract_layer.hpp"
+
+namespace netket {
+
+template <typename T, typename A>
+class Activation : public AbstractLayer<T> {
+  using VectorType = typename AbstractLayer<T>::VectorType;
+  using MatrixType = typename AbstractLayer<T>::MatrixType;
+  using VectorRefType = typename AbstractLayer<T>::VectorRefType;
+  using VectorConstRefType = typename AbstractLayer<T>::VectorConstRefType;
+
+  A activation_;  // activation
+  int size_;      // size_ = input size = output size
+ public:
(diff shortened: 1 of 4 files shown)


NetKet/Machine/activation_layer.hpp
```diff
@@ -0,0 +1,159 @@
...
+                    const VectorType & /*output*/,
+                    std::vector<int> &output_changes,
+                    VectorType &new_output) override {
+    const int num_of_changes = tochange.size();
+    if (num_of_changes > 0) {
+      output_changes = tochange;
+      new_output.resize(num_of_changes);
+      for (int i = 0; i < num_of_changes; ++i) {
+        theta[0](tochange[i]) = newconf[i];
+      }
```
gcarleo: Is this a "light-cone" like implementation that you didn't have for the old activations?

NetKet/Machine/pylayer.hpp
```diff
@@ -65,6 +65,22 @@ void AddLayerModule(py::module &m) {
                                                                    "SumOutput")
         .def(py::init<int>(), py::arg("input_size")) ADDLAYERMETHODS(LayerType);
   }
+  {
```
gcarleo: I believe you also would need to change the python unit tests, especially `Test/Machine/test_machine.py`, since in there I was using the old activation!

gcarleo: No wait, how is it possible that the tests are passing?!

NetKet/Machine/activations.hpp
```diff
@@ -132,6 +134,37 @@ class Tanh : public AbstractActivation {
   }
 };
 
+class Relu : public AbstractActivation {
+ private:
```
gcarleo: These private statements are redundant

NetKet/Machine/pylayer.hpp
```diff
@@ -65,6 +65,22 @@ void AddLayerModule(py::module &m) {
...
+        .def(py::init<int>(), py::arg("input_size")) ADDLAYERMETHODS(LayerType);
+  }
+  {
+    using LayerType = Activation<MachineType, Relu>;
+    py::class_<LayerType, AbLayerType, std::shared_ptr<LayerType>>(subm, "Relu")
+        .def(py::init<int>(), py::arg("input_size")) ADDLAYERMETHODS(LayerType);
+  }
 }
 
 }  // namespace netket
```
gcarleo: I see, the thing is that you didn't remove the activation module `pyactivation.hpp`. Is it because you want to do changes to the Conv and FullyConn layers so that they do not take activation as an input parameter anymore?

NetKet/Machine/activation_layer.hpp
kchoo1118: yea but honestly at the moment the light cones are not handled very well. Now only the first layer will take advantage of light cones

NetKet/Machine/pylayer.hpp
kchoo1118: I didnt want to break any of the old stuff. I wanted to do a minimal additive change first.
But sure we can decide now if want to no longer have the activation within the layer.
This would require some adjustments to the lookup

gcarleo: OK, but eventually the idea would be to separate convolutions and fully connected layers from activations I guess... like in Keras 
Do you need adjustements in the quantity stored in the look up or more complicated changes?

kchoo1118: no it would a simplification actually

gcarleo: Yeah I figured this would simplify things , that's cool

gcarleo: Ok let's merge this and then we can deal with the other changes to the layers later, including the convnet bug

PR #109, opened 2018-11-19
Automatically calling MPI_Init and MPI_Finalize

twesterhout:


NetKet/pynetket.cc
```diff
@@ -37,6 +37,48 @@ using AbSamplerType = AbstractSampler<AbMachineType>;
...
+#if !defined(NDEBUG)
+      std::cerr << "MPI successfully finalized by NetKet." << std::endl;
+#endif
+    }
+  }
+
+ private:
+  bool have_initialized_;
+};
+static MPIInitializer _do_not_use_me_dummy_{};
```
gcarleo: It's not completely trivial to me that this is called at import time of the module

jamesETsmith:
Pybind11 has [a way to import modules](https://pybind11.readthedocs.io/en/stable/advanced/embedding.html?highlight=import#importing-modules) could we use that?

If that doesn't work we could execute python code from inside `pynetket.cc`, maybe something like this:

```cpp
#include <pybind11/embed.h>
namespace py = pybind11;

int main() {
    py::scoped_interpreter guard{};

    py::exec(R"(
        from mpi4py import MPI
    )");
}
```

For more details, the docs are [here](https://pybind11.readthedocs.io/en/stable/advanced/embedding.html?highlight=import#executing-python-code).

twesterhout:
> So far this does not seem to work for me. Maybe an issue with how setup.py handles linking to MPI?

@femtobit  if you compile everything in Debug mode (currently hardcoded in CMakeLists.txt), then the expected results are:

```bash
$ python -c 'import netket' # Initialises and finalises MPI
MPI successfully initialized by NetKet.
MPI successfully finalized by NetKet.
$ python -c 'from mpi4py import MPI; import netket' # Does nothing because MPI is handled by mpi4py
$
```
I'm afraid I'm not familiar with IPython at all... So I also have no idea what `opal_init` should do, but importing `netket` should work regardless of whether MPI has or has not been initialised. If you then try to construct a machine without initialising MPI first -- sure, it'll crash, but not before.

I'm against using `mpi4py` for initialisation of MPI, because it adds an unnecessary dependency.

NetKet/pynetket.cc
twesterhout: It's called at loading time of the shared library. At least it should according to the C++ standard and that is what I'm observing in practice. Whether it's trivial or not is a bit subjective, but I also find rules regarding internal/external linking + static initialisation very confusing.

femtobit:
@twesterhout This is unrelated to IPython (the `opal_init` and related functions belong to OpenMPI, by the way).

Indeed, importing `netket` without previously importing `MPI` from `mpi4py` fails on my machine when using `python -c 'import netket'`. Letting `mpi4py` do the initialization and then importing netket works (also with a Tutorial example).

```
(base) damian@neon:~/repos/research/netket$ python -c 'import netket'
[neon:19716] mca_base_component_repository_open: unable to open mca_patcher_overwrite: /usr/lib/x86_64-linux-gnu/openmpi/lib/openmpi/mca_patcher_overwrite.so: undefined symbol: mca_patcher_base_patch_t_class (ignored)
[neon:19716] mca_base_component_repository_open: unable to open mca_shmem_mmap: /usr/lib/x86_64-linux-gnu/openmpi/lib/openmpi/mca_shmem_mmap.so: undefined symbol: opal_show_help (ignored)
[neon:19716] mca_base_component_repository_open: unable to open mca_shmem_sysv: /usr/lib/x86_64-linux-gnu/openmpi/lib/openmpi/mca_shmem_sysv.so: undefined symbol: opal_show_help (ignored)
[neon:19716] mca_base_component_repository_open: unable to open mca_shmem_posix: /usr/lib/x86_64-linux-gnu/openmpi/lib/openmpi/mca_shmem_posix.so: undefined symbol: opal_shmem_base_framework (ignored)
--------------------------------------------------------------------------
It looks like opal_init failed for some reason; your parallel process is
likely to abort.  There are many reasons that a parallel process can
fail during opal_init; some of which are due to configuration or
environment problems.  This failure appears to be an internal failure;
here's some additional information (which may only be relevant to an
Open MPI developer):

  opal_shmem_base_select failed
  --> Returned value -1 instead of OPAL_SUCCESS
--------------------------------------------------------------------------
--------------------------------------------------------------------------
It looks like orte_init failed for some reason; your parallel process is
likely to abort.  There are many reasons that a parallel process can
fail during orte_init; some of which are due to configuration or
environment problems.  This failure appears to be an internal failure;
here's some additional information (which may only be relevant to an
Open MPI developer):

  opal_init failed
  --> Returned value Error (-1) instead of ORTE_SUCCESS
--------------------------------------------------------------------------
--------------------------------------------------------------------------
It looks like MPI_INIT failed for some reason; your parallel process is
likely to abort.  There are many reasons that a parallel process can
fail during MPI_INIT; some of which are due to configuration or environment
problems.  This failure appears to be an internal failure; here's some
additional information (which may only be relevant to an Open MPI
developer):

  ompi_mpi_init: ompi_rte_init failed
  --> Returned "Error" (-1) instead of "Success" (0)
--------------------------------------------------------------------------
*** An error occurred in MPI_Init
*** on a NULL communicator
*** MPI_ERRORS_ARE_FATAL (processes in this communicator will now abort,
***    and potentially your MPI job)
[neon:19716] Local abort before MPI_INIT completed completed successfully, but am not able to aggregate error messages, and not able to guarantee that all other processes were killed!
(base) damian@neon:~/repos/research/netket$ python -c 'from mpi4py import MPI; import netket'
(base) damian@neon:~/repos/research/netket$ python Tutorials/ImaginaryTime/ising1d_imag.py 
# Transverse-Field Ising model created 
# h = 1
# J = 1
(base) damian@neon:~/repos/research/netket$
```

> I'm against using mpi4py for initialisation of MPI, because it adds an unnecessary dependency.

I think depending on `mpi4py` is fine. We depend on MPI from the C++ side anyway and the Python module already requires working `numpy` and `scipy` installations (otherwise wrapped functions returning Eigen matrices will fail). Adding `mpi4py` as another dependency should not be a problem. @gcarleo What do you think?

twesterhout:
@femtobit are you absolutely sure that you're importing the latest version? I.e. not the one previously built with the code from a different branch? Sorry if this sounds stupid to you... but I don't quite understand how your system is different from Travis where it kind of works. Also, do you perhaps have multiple different MPI implementations on your platform?

So all in all, could you, please, post as much info as possible for me to at least have a chance at reproducing the failure?

P.S.: I've now tested it with Python 2 as well -- all works... There's simply no way the `std::cerr` statements are not executed...

twesterhout:
@femtobit another update: I've now tested with OpenMPI (3.1.3) too. Previous tests ran on MPICH (3.3). Failed with neither Python 2.7.15 nor with Python 3.7.1...

Also, looking more carefully at the output you posted I realise that `MPI_Init` _is_ being called, but fails. The only reason I can think of why this could be the case is if the MPI's implementation doesn't support `nullptr`s being passed to `MPI_Init` (which all conforming implementations of MPI >1.1 should support...). Could you perhaps try something like this:
```cpp
int argc = 1;
char* argv[] = {"dummy"};
```
and then replace `MPI_Init(nullptr, nullptr)` by `MPI_Init(&argc, &argv)`. Alternatively just create a `main.c` file with a call to `MPI_Init(nullptr, nullptr)`.

gcarleo:
@femtobit I agree that it's fine to depend on mpi4py, it's just that if we use that only to call MPI_Init and MPI_Finalize, maybe @twesterhout is right that we can do that ourselves... But sure, if this causes infinite pain it's better just to import mpi4py

gcarleo:
ok @twesterhout , it seems that the issue found by @jamesETsmith is related to conda on mac os only, and not this specific PR. Yes please, go ahead with cleaning and then we merge, thanks!

.travis.yml
```diff
@@ -118,5 +118,6 @@ script:
 
   # compile and execute Python unit tests
   - pip install . --user
+  - PYTHONPATH="$PWD" python Test/Machine/test_machine.py
```
femtobit: This line is probably a leftover from testing?

CMakeLists.txt
```diff
@@ -19,6 +19,7 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/CMakeModules/)
 
+set(CMAKE_BUILD_TYPE "Debug")
```
femtobit: This probably also shouldn't be in the final version, right?

NetKet/pynetket.cc
gcarleo: Just as a matter of readability, I'd maybe move the implementation of `MPIInitializer` somewhere else, maybe in `Utils`

twesterhout: `static MPIInitializer _do_not_use_me_dummy_{};` should remain in the `.cpp` file, but the rest can definitely be moved. Will do.

.travis.yml
twesterhout: Right you are!

CMakeLists.txt
twesterhout: Correct again!

twesterhout:
Tests do pass locally for me on Linux. My guess is that the failures have more to do with `.travis.yml` rather than the C++ code. But to know for sure I suggest we merge #90 first.

gcarleo:
This looks like exactly the same error we had in #109 on Linux... puzzling, most likely a Travis issue, but it's a bit strange, since the tests start and then for some reason die. I hope this is not related to some hidden seg fault

cmendl:
Maybe related to this discussion: I'm experiencing the following error when running a tutorial (Ubuntu Linux 18.04 LTS, netket v2.0 branch):
```
christian@VirtualSchorsch:~/netket/Tutorials/Ising1d$ python3 ising1d.py 
Attempting to use an MPI routine before initializing MPICH
```

And importing mpi4py before netket results in a segfault:
```
christian@VirtualSchorsch:~/netket/Tutorials/Ising1d$ python3 -c 'from mpi4py import MPI; import netket; g = netket.graph.Hypercube(4, 1); h = netket.hilbert.Spin(g, 0.5); m = netket.machine.RbmSpin(h, 10)'
[VirtualSchorsch:05656] *** Process received signal ***
[VirtualSchorsch:05656] Signal: Segmentation fault (11)
[VirtualSchorsch:05656] Signal code: Address not mapped (1)
[VirtualSchorsch:05656] Failing at address: 0x44000098
[VirtualSchorsch:05656] [ 0] /lib/x86_64-linux-gnu/libc.so.6(+0x3ef20)[0x7fdee26f8f20]
[VirtualSchorsch:05656] [ 1] /usr/lib/x86_64-linux-gnu/libmpi.so.20(MPI_Comm_rank+0x42)[0x7fdee0eaec42]
[VirtualSchorsch:05656] [ 2] /home/christian/netket/netket.cpython-36m-x86_64-linux-gnu.so(+0x16de2)[0x7fdecf604de2]
[VirtualSchorsch:05656] [ 3] /home/christian/netket/netket.cpython-36m-x86_64-linux-gnu.so(+0x61e50)[0x7fdecf64fe50]
[VirtualSchorsch:05656] [ 4] /home/christian/netket/netket.cpython-36m-x86_64-linux-gnu.so(+0xb1ce0)[0x7fdecf69fce0]
[VirtualSchorsch:05656] [ 5] /home/christian/netket/netket.cpython-36m-x86_64-linux-gnu.so(+0xb2899)[0x7fdecf6a0899]
[VirtualSchorsch:05656] [ 6] python3(_PyCFunction_FastCallDict+0x35c)[0x565d5c]
[VirtualSchorsch:05656] [ 7] python3[0x5914e1]
[VirtualSchorsch:05656] [ 8] python3[0x54b985]
[VirtualSchorsch:05656] [ 9] python3[0x555421]
[VirtualSchorsch:05656] [10] python3(_PyObject_FastCallKeywords+0x19c)[0x5a730c]
[VirtualSchorsch:05656] [11] python3[0x503073]
[VirtualSchorsch:05656] [12] python3(_PyEval_EvalFrameDefault+0x449)[0x506859]
[VirtualSchorsch:05656] [13] python3[0x504c28]
[VirtualSchorsch:05656] [14] python3(PyRun_StringFlags+0xaf)[0x634f9f]
[VirtualSchorsch:05656] [15] python3(PyRun_SimpleStringFlags+0x3d)[0x63840d]
[VirtualSchorsch:05656] [16] python3(Py_Main+0x33c)[0x638f1c]
[VirtualSchorsch:05656] [17] python3(main+0xe0)[0x4a6f10]
[VirtualSchorsch:05656] [18] /lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xe7)[0x7fdee26dbb97]
[VirtualSchorsch:05656] [19] python3(_start+0x2a)[0x5afa0a]
[VirtualSchorsch:05656] *** End of error message ***
Segmentation fault (core dumped)
```

gcarleo:
@cmendl , sorry I misinterpreted your first error

> christian@VirtualSchorsch:~/netket/Tutorials/Ising1d$ python3 ising1d.py 
Attempting to use an MPI routine before initializing MPICH

This shouldn't happen, since the tutorial imports mpi4py, unless you modified it. Can you please check? 

If you import mpi4py in the tutorial, but you still get that error then it definitely means that mpi4py has been installed with a different MPI library than the one used to install netket

cmendl:
@gcarleo Thanks for the clarification, and pointing out the different MPI libraries: reinstalling mpi4py using pip (instead of the package manager) fixed the issue.

gcarleo:
Is anybody looking into this at the moment? We are approaching 2.0 release, and it is quite important in my opinion to remove the ugly dependency from mpi4py , at least for simple usage patterns... Unless there are objections, I am going to open a PR with the approach I was proposing at some point.

twesterhout:
> Is anybody looking into this at the moment? We are approaching 2.0 release, and it is quite important in my opinion to remove the ugly dependency from mpi4py , at least for simple usage patterns... Unless there are objections, I am going to open a PR with the approach I was proposing at some point.

@gcarleo I'm still waiting for feedback :) It seems like errors people encountered with this PR were related to installations of MPI (i.e. linking against wrong libraries). But then noone actually approved this PR...

How about I try building it against latest `v2.0` and if Travis keeps complaining, we try your approach (hopefully Travis likes it better)? The changes here seem so obvious and simple, that I have no idea what could go wrong.

gcarleo:
But automated testing was failing here, that's why it was never approved :) Anyway, yes, please try it with the latest v2.0, I think there is chance that the MPI installation should work better on Travis now?

twesterhout:
> But automated testing was failing here, that's why it was never approved :)

@gcarleo how about now? :)

gcarleo:
Great! this is much better now @twesterhout. If by any chance you have some spare time to remove all the unnecessary mpi4py from Test/, Tutorials/ , Examples/ it would be great... should be straightforward to do with a replace all in project files. The only place where mpi4py is actually nedeed is `Examples/PyNetKet/ground_state_iter.py` . If you can't it's fine, I can do that later once we merge this.

Issue #48, opened 2018-07-22
Built-in function IsBipartite 


gcarleo:
The current implementation of `CustomGraph` relies on external input to determine whether a user-defined Graph is bipartite or not, however it is fairly simple to implement a built-in check. 
For example see https://www.geeksforgeeks.org/bipartite-graph/

gcarleo:
#161 contains a generic implementation of bipartite, we should make sure it is ported in `CustomGraph` too

PR #110, opened 2018-11-19
WIP: Finishing up Python support for graphs

### LINKED ISSUE #101: Python support for custom graphs
The python bindings need to be finished/polished a bit

twesterhout:
Should fix #101.

### DIFF
--- a/NetKet/Graph/abstract_graph.hpp
+++ b/NetKet/Graph/abstract_graph.hpp
@@ -34,16 +34,27 @@ namespace netket {
 class AbstractGraph {
  public:
   /**
-  Member function returning the number of sites (nodes) in the graph.
-  @return Number of sites (nodes) in the graph.
+  Custom type for unordered_map<array<int,2>, int> w/ a custom hash function
   */
-  virtual int Nsites() const = 0;
+  using Edge = std::array<int, 2>;
+  using ColorMap = std::unordered_map<Edge, int, netket::ArrayHasher>;
 
   /**
   Member function returning the number of sites (nodes) in the graph.
   @return Number of sites (nodes) in the graph.
   */
-  virtual int Size() const = 0;
+  virtual int Nsites() const noexcept = 0;
+
+  /**
+  Member function returning the number of sites (nodes) in the graph.
+  @return Number of sites (nodes) in the graph.
+  */
+  virtual int Size() const noexcept = 0;
+
+  /**
+  Returns the graph edges.
+  */
+  virtual std::vector<Edge> const &Edges() const noexcept = 0;
 
   /**
   Member function returning the adjacency list of the graph.
@@ -58,16 +69,10 @@ class AbstractGraph {
   virtual std::vector<std::vector<int>> SymmetryTable() const = 0;
 
   /**
-  Custom type for unordered_map<array<int,2>, int> w/ a custom hash function
-  */
-  using Edge = std::array<int, 2>;
-  using ColorMap = std::unordered_map<Edge, int, netket::ArrayHasher>;
-
-  /**
   Member function returning edge colors of the graph.
   @return ec[i][j] is the color of the edge between nodes i and j.
   */
-  virtual const ColorMap &EdgeColors() const = 0;
+  virtual const ColorMap &EdgeColors() const noexcept = 0;
 
   // Edge Colors from users specified map
   void EdgeColorsFromList(const std::vector<std::vector<int>> &colorlist,
@@ -91,14 +96,14 @@ class AbstractGraph {
   Member function returning true if the graph is bipartite.
   @return true if lattice is bipartite.
   */
-  virtual bool IsBipartite() const = 0;
+  virtual bool IsBipartite() const noexcept = 0;
 
   /**
    * Checks whether the graph is connected, i.e., there exists a path between
    * every pair of nodes.
    * @return true, if the graph is connected
    */
-  virtual bool IsConnected() const = 0;
+  virtual bool IsConnected() const noexcept = 0;
 
   /**
    * Perform a breadth-first search (BFS) through the graph, calling
--- a/NetKet/Graph/custom_graph.hpp
+++ b/NetKet/Graph/custom_graph.hpp
@@ -310,20 +310,22 @@ class CustomGraph : public AbstractGraph {
     return automorphisms_;
   }
 
-  int Nsites() const override { return nsites_; }
+  int Nsites() const noexcept override { return nsites_; }
 
-  int Size() const override { return nsites_; }
+  int Size() const noexcept override { return nsites_; }
+
+  std::vector<Edge> const &Edges() const noexcept override { return edges_; }
 
   std::vector<std::vector<int>> AdjacencyList() const override {
     return adjlist_;
   }
 
-  bool IsBipartite() const override { return isbipartite_; }
+  bool IsBipartite() const noexcept override { return isbipartite_; }
 
-  bool IsConnected() const override { return is_connected_; }
+  bool IsConnected() const noexcept override { return is_connected_; }
 
   // Returns map of the edge and its respective color
-  const ColorMap &EdgeColors() const override { return eclist_; }
+  const ColorMap &EdgeColors() const noexcept override { return eclist_; }
 
  private:
   bool ComputeConnected() const {
--- a/NetKet/Graph/hypercube.hpp
+++ b/NetKet/Graph/hypercube.hpp
@@ -15,7 +15,6 @@
 #ifndef NETKET_HYPERCUBE_HPP
 #define NETKET_HYPERCUBE_HPP
 
-#include <mpi.h>
 #include <algorithm>
 #include <array>
 #include <cassert>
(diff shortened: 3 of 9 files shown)


NetKet/Graph/hypercube.hpp
```diff
@@ -136,59 +125,93 @@ class Hypercube : public AbstractGraph {
...
   std::vector<std::vector<int>> SymmetryTable() const override {
     return symm_table_;
   }
 
+  /// Given a site's coordinate as a `Ndim()`-dimensional vector, returns the
+  /// site's index. The mapping is unique, but unspecified. I.e. the fact that
+  /// sometimes `Coord2Site({1, 2, 0, 0, 1}) == X` and `Coord2Site({1, 2, 0, 0,
+  /// 2}) == X+1` should not be relied on.
+  int Coord2Site(std::vector<int> const &coord) const {
+    auto const print_coord = [&coord](std::ostream &os) -> std::ostream & {
```
gcarleo: Is this print for debugging?

NetKet/Graph/hypercube.hpp
```diff
@@ -136,59 +125,93 @@ class Hypercube : public AbstractGraph {
...
 
   std::vector<std::vector<int>> SymmetryTable() const override {
     return symm_table_;
   }
 
+  /// Given a site's coordinate as a `Ndim()`-dimensional vector, returns the
+  /// site's index. The mapping is unique, but unspecified. I.e. the fact that
+  /// sometimes `Coord2Site({1, 2, 0, 0, 1}) == X` and `Coord2Site({1, 2, 0, 0,
+  /// 2}) == X+1` should not be relied on.
+  int Coord2Site(std::vector<int> const &coord) const {
```
gcarleo: We would also need the inverse mapping, from site to coordinates. This was previously returned by the member `Sites`

gcarleo:
@twesterhout I fixed it, it was a problem with python 2 sorting

gcarleo:
Interestingly enough, it seems our unit tests found a bug in NetworkX, LOL. 
According to networkx this graph is disconnected ... 
```
[[0, 1], [0, 3], [1, 2], [1, 3], [3, 4]]
```

gcarleo:
I am merging this later today unless you have objections

twesterhout:
> I am merging this later today unless you have objections

@gcarleo Yes, I do. I'll get back to you asap.

Test/Graph/test_graph.py
```diff
@@ -73,3 +82,6 @@ def test_computes_distances():
             for i in range(graph.n_sites):
                 for j in range(graph.n_sites):
                     assert d1[i][j] == d[i][j]
+
+
+test_edges_are_correct()
```
twesterhout: We probably shoudn't call `test_*` functions explicitly. `pytest` runs them automatically.

NetKet/Graph/abstract_graph.hpp
```diff
@@ -54,7 +54,7 @@ class AbstractGraph {
   /**
   Returns the graph edges.
   */
-  virtual std::vector<Edge> const &Edges() const noexcept = 0;
+  virtual std::vector<Edge> Edges() const noexcept = 0;
```
twesterhout: Why? I made `Edges()` return a reference on purpose.

NetKet/Graph/custom_graph.hpp
```diff
@@ -47,34 +44,13 @@ class CustomGraph : public AbstractGraph {
 
   std::vector<std::vector<int>> automorphisms_;
 
-  bool isbipartite_;
```
twesterhout: Why? This info is used for optimisation.

twesterhout:
> It's OK for me, but I a did a few changes, @twesterhout tell me if I can merge this with my changes

I think I agree with everything before 2675dc3, but definitely not with the latest two commits...

Test/Graph/test_graph.py
gcarleo: Indeed that's a leftover

NetKet/Graph/abstract_graph.hpp
gcarleo: Yes but how is this reference handled in python? What if the graph goes out of scope?

gcarleo:
> I think I agree with everything before 2675dc3, but definitely not with the latest two commits...

Alright, just revert to the commit you agree with. I thought you weren't active on this branch anymore and pushed changed on here. My fault. 

I would keep graph methods as read_only properties though, unless there are good reasons not to do so.

NetKet/Graph/custom_graph.hpp
gcarleo: Then you need to store also adjacency matrices etc, if you don't want to compute them every time you call the corresponding method. Ideally you need a flag that stores the state of the computation and possibly avoids recomputing everything from scratch if already computed.

twesterhout:
> Interestingly enough, it seems our unit tests found a bug in NetworkX, LOL.
According to networkx this graph is disconnected ...

Nope, it's my bug :) Sorry. I missed the fact that `networkx.dense_gnm_random_graph` returns a *directed* graph. In that case `[[0, 1], [0, 3], [1, 2], [1, 3], [3, 4]]` is indeed disconnected.

```python
>>> import networkx
>>> networkx.is_connected(networkx.from_edgelist([[0, 1], [0, 3], [1, 2], [1, 3], [3, 4]]))
True
```

twesterhout:
> @twesterhout , did you comment something about the networkx problem?
The issue there is that one of the random graph has a completely disconnected node, that does not show up in the list of edges (I think this is a "conceptual" bug in networkx i.m.h.) since we construct our graph from their list of edges, the extra disconnected node is not even picked up by our graph

Yes, but what I wrote turned out to be wrong, so I deleted it to not confuse anyone. That's a very clear explanation of the problem, by the way, thanks! I think we should just come up with a simple condition to filter out such graphs.

twesterhout:
> I would keep graph methods as read_only properties though, unless there are good reasons not to do so.

@gcarleo about that. I certainly agree that `n_sites`, `edges`, `is_connected`, and `is_bipartite` (if we decide to leave it) should be properties. Not so sure about `adjacency_list` and `distances` though, as they recompute the result on every call so it feels natural to declare them as functions. What do you think?

NetKet/Graph/custom_graph.hpp
twesterhout: It's not about recomputing. `CustomGraph` now always assumes that it's not bipartite. I suggest we leave the argument there for now and add implementing two-coloring to the TODO list. [this](https://github.com/netket/netket/blob/9570034e38d98846229015bb3c5676e84cb5ff61/NetKet/Hamiltonian/heisenberg.hpp#L59) has a *huge* impact on convergence.

NetKet/Graph/abstract_graph.hpp
twesterhout: The way I had it implemented :) Using iterables. One can always use `list` it Python to construct a list from the edges and the graph can be safely garbage-collected. Also, `edges` uses `keep_alive`, so stuff like this works:
```python
>>> import gc
>>> edges = g.edges()
>>> g = None
>>> gc.collect()
23
>>> list(edges)
[[0, 1], [1, 2], [2, 3], [0, 3]]
```

NetKet/Graph/custom_graph.hpp
gcarleo: There is already an open issue about that #48 , that's why I just set up a warning. We can integrate this at a later stage. And by the way, the impact on the convergence is only if you don't do things properly :)

gcarleo:
> @gcarleo about that. I certainly agree that n_sites, edges, is_connected, and is_bipartite (if we decide to leave it) should be properties. Not so sure about adjacency_list and distances though, as they recompute the result on every call so it feels natural to declare them as functions. What do you think?

The thing is that the graph is now to all purposes a const object. Thus you cannot modify its structure dynamically (i.e. attaching more nodes etc). In this sense, all properties are constant and never change during its life. That's why I was suggesting maybe to put a flag `adjacency_list_computed_` that basically prompts a computation of the adjacency list only if set up on false. Once computed, this is stored internally and returns as a const every time we call `GetAdjacencyList()` or something. 
Maybe this can be done in a nicer way using `std::optional` ...

NetKet/Graph/custom_graph.hpp
twesterhout: > There is already an open issue about that #48

I see.

> And by the way, the impact on the convergence is only if you don't do things properly :)

Out of curiosity, what do you mean by "not doing thing properly"?

twesterhout:
Yeah, this can very easily be accomplished using `std::optional`, but what's the point? I mean, `adjacency_list` function is there for testing and nothing more. If I need to do something with graphs, I'll just do it in networkx, no need to reinvent the wheel. On the C++ side, `AdjacencyList` function is only used during initialisation, so there's little to gain by caching its result.

Anyway, should I add an extra commit making `adjacency_list` and `distances` properties? Or do you want to merge this PR in its current state (i.e. before I break something :)).

NetKet/Graph/pygraph.hpp
```diff
@@ -182,19 +182,20 @@ void AddGraphModule(py::module& m) {
               Returns the number of vertices in the graph.
            )EOF")
       .def_property_readonly("edges",
```
gcarleo: Can we find a compromise here? Maybe you can define another method "edges_iterator" that returns the iterator and edges to return the actual list? NetworkX for `edges` does return a simple list, as far as I can see, and not an iterator, for example

twesterhout: I understand that a bit more functional style of programming that I bring may seem too new and even too involved at times, so I won't push it unless I feel it's really important (like in the discussion about Monte Carlo we had about a week ago). So I'll just make `edges` return a list, but do note that `networkx`'s `edges` returns an `EdgeView` rather than a list (at least in version 2.2 on Python 3).

gcarleo: OK thanks, let's merge

Issue #94, opened 2018-11-17
Rewrite tutorials using the Python library


gcarleo:

gcarleo:
Duplicate issue #157 , closing this one

PR #112, opened 2018-11-20
removed activations from fully connected layer 

kchoo1118:
1) removed activations from fully connected layer 
2) lookups were simplified
3) info messages modified

Todo:
1) conv net issue
2) additional layer information

### DIFF
--- a/NetKet/Machine/activation_layer.hpp
+++ b/NetKet/Machine/activation_layer.hpp
@@ -35,6 +35,9 @@ class Activation : public AbstractLayer<T> {
 
   A activation_;  // activation
   int size_;      // size_ = input size = output size
+
+  std::string name_;
+
  public:
   using StateType = typename AbstractLayer<T>::StateType;
   using LookupType = typename AbstractLayer<T>::LookupType;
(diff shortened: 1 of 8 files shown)


NetKet/Machine/activation_layer.hpp
```diff
@@ -75,30 +76,25 @@ class Activation : public AbstractLayer<T> {
 
   void InitLookup(const VectorType &v, LookupType &lt,
                   VectorType &output) override {
-    lt.resize(1);
-    lt[0].resize(size_);
+    lt.resize(0);
```
gcarleo: Do you need to resize it at all?

Test/Machine/test_machine.py
```diff
@@ -31,8 +31,7 @@
 layers = [
     nk.layer.FullyConnected(
         input_size=g.n_sites,
-        output_size=40,
-        activation=nk.activation.Lncosh())
+        output_size=40)
```
gcarleo: Shouldn't you add an activation layer here?

kchoo1118:
also the tutorial is outdated i think

Test/Machine/test_machine.py
gcarleo: yes, but maybe in the next PR we should add checks for all the implemented layers, just to make sure it's all OK!

gcarleo:
> also the tutorial is outdated i think 

yes, that's issue #94 , we will fix that once we finish stabilizing the interface of the various classes. At the moment we are missing just a couple of classes

PR #115, opened 2018-11-21
Remove lookups from the layers

kchoo1118:


kchoo1118:
The conv is not fixed yet.
I think maybe I will just change it to a hypercube only conv net

NetKet/Machine/abstract_layer.hpp
```diff
@@ -79,18 +79,6 @@ class AbstractLayer {
   @param sigma is the variance of the gaussian.
   */
   virtual void InitRandomPars(int seed, double sigma) = 0;
-  /**
-  Member function to initialise the lookup for the later. "Summary: This
```
gcarleo: Why remove the comment? :)

NetKet/Machine/abstract_layer.hpp
```diff
@@ -79,18 +79,6 @@ class AbstractLayer {
...
-  function uses the input vector v to compute the lookup lt and the output. This
-  function also resizes the lookup elements to the correct size."
-  @param v is the input vector.
-  @param theta is a vector in the lookuptable storing the output of the
-  layer before the nonlinear function is applied.
-  @param output is a vector in the lookuptable storing the previous output of
-  the current layer.
-  */
-  virtual void InitLookup(const VectorType &v, LookupType &lt,
```
gcarleo: Finally gone!

NetKet/Machine/abstract_layer.hpp
```diff
@@ -114,7 +102,7 @@ class AbstractLayer {
   */
   virtual void UpdateLookup(const VectorType &input,
```
gcarleo: Maybe we can just call these UpdateOutput, just to remove any reference to Lookup tables ?

NetKet/Machine/abstract_layer.hpp
```diff
@@ -185,7 +157,6 @@ class AbstractLayer {
   */
   virtual void Backprop(const VectorType &prev_layer_output,
                         const VectorType &this_layer_output,
-                        const LookupType &this_layer_theta,
```
gcarleo: Great

NetKet/Machine/sum_output.hpp
```diff
@@ -84,30 +84,20 @@ class SumOutput : public AbstractLayer<T> {
 
   void SetParameters(VectorConstRefType /*pars*/, int /*start_idx*/) override {}
```
gcarleo: I never liked these SetParameters and GetParameters with two arguments... Can't we just use an Eigen:Map as you have done in other parts? https://eigen.tuxfamily.org/dox/group__TutorialReshapeSlicing.html

gcarleo:
@kchoo1118  the easiest is to use https://eigen.tuxfamily.org/dox/group__TutorialBlockOperations.html 

There is the very nice 
```c++
vector.segment(i,n);
```

NetKet/Machine/abstract_layer.hpp
kchoo1118: sure.

NetKet/Machine/sum_output.hpp
kchoo1118: sure that is possible.

PR #117, opened 2018-11-21
Completing python unit tests for hamiltonians.

### LINKED ISSUE #116: Finish up Hamiltonian Python unit tests

jamesETsmith:
This PR resolves #116.


jamesETsmith:
I'll add the python bindings for custom hamiltonians as well unless someone has an objection.

gcarleo:
No it's OK!

jamesETsmith:
Sorry, do you mean you want me to add the bindings or skip them?

gcarleo:
I meant add them, sorry

gcarleo:
No wait, custom hamiltonian is just an Operator now! We actually just need
to remove custom hamiltonian and move its tests to operator

jamesETsmith:
Ah yes, I forgot about the LocalOperator class, I will just use that

gcarleo:
@jamesETsmith let me explain better.
Now the logic is: 
1. You construct a graph (this can just be a bunch of nodes, or nodes with edges)
2. You construct a hilbert object defined on this graph
3. You construct an `Operator` from the `Hilbert` object

In turn this implies that we don't have anymore the `Hamiltonian`, since that's a special case of an `Operator`. This also means that we need to rethink the way we construct your `GraphHamiltonian`. 
This should become a `GraphOperator`, and again be constructed just from a hilbert object, as done from the other operators. You can deduce from the graph from the hilbert object calling `hilbert_.GetGraph()`.  This means that in the `GraphOperator` constructor there is no more need to pass edges colors etc! You just need to pass the edge operator (the matrix) and everything should be handled by the constructor. Basically you will build `LocalOperator` from the matrix and the edges you have deduced from the graph.

I hope this is clear!

jamesETsmith:
Thanks for the explanation. I've been playing with the Local Operator and I've been running into an error when I try to construct a custom hamiltonian from several local operators. For example

```python
sx = [[0, 1], [1, 0]]
sy = np.array([[0, 1j], [-1j, 0]]).tolist()
sz = [[1, 0], [0, -1]]

g = nk.graph.CustomGraph(edges=edges)
hi = nk.hilbert.CustomHilbert(local_states=[1, -1], graph=g)

sx_hat = nk.operator.LocalOperator(hi, [sx] * 3, [[0], [1], [5]])
sy_hat = nk.operator.LocalOperator(hi, [sy] * 4, [[2], [3], [4], [9]])
szsz_hat = nk.operator.LocalOperator(hi, sz, [0]) * nk.operator.LocalOperator(
    hi, sz, [1])
szsz_hat += nk.operator.LocalOperator(hi, sz, [4]) * nk.operator.LocalOperator(
    hi, sz, [5])
szsz_hat += nk.operator.LocalOperator(hi, sz, [6]) * nk.operator.LocalOperator(
    hi, sz, [8])
szsz_hat += nk.operator.LocalOperator(hi, sz, [7]) * nk.operator.LocalOperator(
    hi, sz, [0])

operators["Custom Hamiltonian"] = sx_hat + sy_hat
```

Gives a segmentation fault error when I call the function `get_conn()` on `operators["Custom Hamiltonian"]`, has anyone else seen this before? If I only set the `operators["Custom Hamiltonian"] = sx_hat` then I don't see the seg fault.

jamesETsmith:
@gcarleo I will switch the graph hamiltonian over to a graph operator, after I get this seg fault issue figured out.

gcarleo:
@jamesETsmith thanks for pointing this out, indeed the operator `+=` should be OK, but I also experienced problems with the operator `+`, I think there is a memory problem, but I hadn't the time to solve it yet. In any case, maybe you can do everything without using the `+` operator?

jamesETsmith:
@gcarleo unfortunately that doesn't work either. As an example, the following code gives a seg fault before it reaches the print statement `print("After")`. I will keep tinkering to see if I can figure out what's going on here.

```python
sx = [[0, 1], [1, 0]]
sy = np.array([[0, 1j], [-1j, 0]]).tolist()
sz = [[1, 0], [0, -1]]

g = nk.graph.CustomGraph(edges=edges)
hi = nk.hilbert.CustomHilbert(local_states=[1, -1], graph=g)

sx_hat = nk.operator.LocalOperator(hi, [sx] * 3, [[0], [1], [5]])
sy_hat = nk.operator.LocalOperator(hi, [sy] * 4, [[2], [3], [4], [9]])
szsz_hat = nk.operator.LocalOperator(hi, sz, [0]) * nk.operator.LocalOperator(
    hi, sz, [1])
szsz_hat += nk.operator.LocalOperator(hi, sz, [4]) * nk.operator.LocalOperator(
    hi, sz, [5])
szsz_hat += nk.operator.LocalOperator(hi, sz, [6]) * nk.operator.LocalOperator(
    hi, sz, [8])
szsz_hat += nk.operator.LocalOperator(hi, sz, [7]) * nk.operator.LocalOperator(
    hi, sz, [0])

sx_hat += sy_hat
v = np.ones(10)
print("Before")
szsz_hat.get_conn(v)
print("After")
```

jamesETsmith:
@gcarleo, I've gone back to look at the GraphOperator (formally GraphHamiltonian) and I wanted to ask you some clarifying questions. 

1. Are you looking for the GraphOperator to extend the LocalOperator? Essentially providing an easy way to combine local operators acting on sites and edges?
1. GraphOperator already doesn't take the edge colors as an argument to initialize, that information is extracted from the graph. What is the edge operator you want to pass in the constructor?

jamesETsmith:
Sorry I didn't ask those questions earlier when you first mentioned the graph operator, I should have read it more thoroughly the first time before saying I understood!

gcarleo:
Yes basically `GraphOperator` is very similar to a `LocalOperator`, it's just the way it is constructed that changes... !

jamesETsmith:
Sorry for the delay on this. I had a deadline yesterday. I've replaced the vector of local operators, with a single local operator, that is is modified and updates as we add operators for sites and bonds. 

The initialization of the `operator_` in the initializer list is a bit messy. Would it be worthwhile to add a constructor to LocalOperator where we only give a hilbert space? If we did this, we wouldn't need to pass other empty objects and I think it would be a much nicer/cleaner way to initialize a GraphOperator. 

On a related note, would it make sense to overload `+` and `*` operators for GraphOperator objects like we do for LocalOperator?

gcarleo:
@jamesETsmith no problem for the delay, I understand! 
You work on this is solid I think! And yes, we can maybe define an empty LocalOperator passing only hilbert as a constructor, I agree...

gcarleo:
@jamesETsmith , it seems there are some conflicts with the upstream version. Let me know if you have trouble fixing them, otherwise I can give it a shot

jamesETsmith:
I am running into a bit of trouble, if I can't figure it out by this afternoon, I'll let you know.

jamesETsmith:
@gcarleo The issue I ran into when merging was due (I believe) to the updating `GraphOperator::operator_` as we add new `LocalOperators` to it, e.g. `operator_ = operator_ + LocalOperator(hilbert_, mat, site)`. 

In the newest version of netket v2.0, `LocalOperators` have a `const Hilbert & hilbert_` member variable, rather than a shared pointer version, and this caused some problems when I tried to add new `LocalOperators` to `GraphOperator::operator_`. 

To fix this I overloaded the assignment operator for the local operator so now it checks that the hilbert spaces have the same size and copies all member vabiables except `hilbert_`. Let me know if you're ok with this solution!

NetKet/Operator/graph_operator.hpp
```diff
@@ -102,14 +103,12 @@ class GraphHamiltonian : public AbstractOperator {
     newconfs.clear();
     mel.resize(0);
 
-    for (std::size_t i = 0; i < operators_.size(); i++) {
-      operators_[i].AddConn(v, mel, connectors, newconfs);
-    }
+    operator_.AddConn(v, mel, connectors, newconfs);
```
gcarleo: Can we use FindConn here instead of AddConn? Would that be a problem? I wanted to remove the method AddConn in Operator, since it seems unecessary in the new version!

NetKet/Operator/local_operator.hpp
```diff
@@ -293,6 +297,17 @@ class LocalOperator : public AbstractOperator {
     return LocalOperator(lhs.GetHilbert(), mat, sites);
   }
 
+  void operator=(const LocalOperator &rhs) {
+    assert(rhs.hilbert_.LocalStates().size() == hilbert_.LocalStates().size());
```
gcarleo: OK, but maybe we can also check the other properties are equal? 
Would it work to do 
```c++
assert(rhs.hilbert_==hilbert_);
```

? Maybe not

Tutorials/BoseHubbard1d/bosehubbard1d.py
```diff
@@ -12,67 +12,37 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import netket as nk
+from mpi4py import MPI
 
-from __future__ import print_function
-import json
+# 1D Periodic Lattice
+g = nk.graph.Hypercube(length=12, ndim=1, pbc=True)
```
gcarleo: just check again that this tutorials work maybe? now for example it's `n_dim` not `ndim`... PEP!

Tutorials/BoseHubbard1d/bosehubbard1d.py
```diff
@@ -12,67 +12,37 @@
...
-# here we use Metropolis sampling
-pars['Sampler'] = {
-    'Name': 'MetropolisHamiltonian',
-}
+# Variational Monte Carlo
+vmc = nk.gs.Vmc(
+    hamiltonian=ha,
+    sampler=sa,
+    optimizer=op,
+    nsamples=1000,
```
gcarleo: `n_samples`

NetKet/Operator/local_operator.hpp
gcarleo: Definitely not the last `assert`! We can set up a comparison operator in the abstract `hilbert` to checkout out that the properties are equal... but that's for another PR, no worries.

gcarleo:
@jamesETsmith to answer to your question 

> On a related note, would it make sense to overload + and * operators for GraphOperator objects like we do for LocalOperator?

Yes, if you have time to add these two overloads it would be great.

NetKet/Operator/graph_operator.hpp
jamesETsmith: I removed it and it doesn't seem to cause any problems.

Tutorials/BoseHubbard1d/bosehubbard1d.py
jamesETsmith: I'll change these over!

Tutorials/BoseHubbard1d/bosehubbard1d.py
jamesETsmith: nsamples is the currently supported name, but I can change this over too!

NetKet/Operator/pyoperator.hpp
```diff
@@ -15,22 +15,22 @@
 #ifndef NETKET_PYOPERATOR_HPP
 #define NETKET_PYOPERATOR_HPP
 
+#include "netket.hpp"
```
gcarleo: Better include just the headers used here, i.e. 'Operator/operator.hpp' etc

gcarleo:
Hi @jamesETsmith , can I make a few edits to your PR to address the things I mentioned? I would work on your branch directly, if that's not a problem for you

jamesETsmith:
@gcarleo, I've made most of the changes you requested, but when I run the Tutorials, the energies I get are incorrect. I've been trying to track down the problem before pushing my latest commits. I can push them now and you're welcome to change what you like!

jamesETsmith:
Sorry I will resolve these conflicts. I am pulling from upstream, I promise, just not fast enough!

gcarleo:
Alright no worries, just let me know what tutorials go wrong and I'll have a look

jamesETsmith:
@gcarleo the `GraphOperator/Ising/ising.py` and `GraphOperator/J1J2/j1j2.py` aren't working currently.

gcarleo:
> @gcarleo the GraphOperator/Ising/ising.py and GraphOperator/J1J2/j1j2.py aren't working currently.

OK, but I don't understand why unit tests are not running here in the PR. Maybe some conflict resolution gone wrong? uhm

jamesETsmith:
Hmm, this is puzzling. On this PR, I can't see checks for commit e4c8dfc, but I can see that it's run on Travis.

Tutorials/GraphOperator/Ising/ising.py
```diff
@@ -0,0 +1,66 @@
...
+
+# Restricted Boltzmann Machine
+ma = nk.machine.RbmSpin(hilbert=hi, alpha=1)
+
+# Local Metropolis Sampling
+sa = nk.sampler.MetropolisLocal(machine=ma)
+
+# Gradient Descent
+opt = nk.optimizer.AdaMax()
+
```
gcarleo: you are not initializing the machine to random parameters, most likely this is the problem? In `Tutorials/ground_state.py' this is done

```python
# Machine
ma = nk.machine.RbmSpin(hilbert=hi, alpha=1)
ma.init_random_parameters(seed=1234, sigma=0.01)
```

gcarleo:
@jamesETsmith it may be that if you don't initialize the parameters the energy just doesn't change, because the machine would be initialized to zero weights, which is a pathological case!

Tutorials/GraphOperator/Ising/ising.py
jamesETsmith: Adding it to the `GraphOperator/Ising/ising.py` doesn't fix it. The energies are really low from the first iteration, e.g. -0.5 instead of nearly -20.

gcarleo: Interesting... those shouldn't work either :) uhm, more puzzles

gcarleo:
OK let me give it a try locally

gcarleo:
@jamesETsmith can I push to this repo?

jamesETsmith:
Yeah absolutely

jamesETsmith:
Something is not right with my repo, some of the changes from this PR are there and others are not. Is that from the force push? I have never used the `-f` flag to push until this PR and I haven't had this much trouble with git until now, so I'm getting the feeling that they're related.

gcarleo:
Indeed you don't have the += operator etc in `LocalOperator`, that's what I was finding out... I don't know about forced pushes, it sounds violent to me though... Can you just regress to some older version?

jamesETsmith:
Yeah, I'm going to revert and then I'll make all of the updates in a cleaner way (I hope).

gcarleo:
Is it OK now with the tutorials?

jamesETsmith:
The ising tutorial is working now and I'm updating the J1J2 tutorial now and I'll compare them.

gcarleo:
> The ising tutorial is working now and I'm updating the J1J2 tutorial now and I'll compare them.

Great! 

Just try to see if you can remove `AddConn`, I think it should work just with `FindConn` instead 
```c++
  void FindConn(VectorConstRefType v, std::vector<std::complex<double>> &mel,
                std::vector<std::vector<int>> &connectors,
                std::vector<std::vector<double>> &newconfs) const override {
    operators_.FindConn(v, mel, connectors, newconfs);
  }
```

then we merge... ! Thanks

jamesETsmith:
I missed that one, thanks for catching it!

jamesETsmith:
The `Tutorials/GraphOperator/J1J2/j1j2.py` and `Tutorials/J1J2/j1j2.py` are both working in the new local operator style.

jamesETsmith:
@gcarleo, I think it's finally ready to go!

PR #118, opened 2018-11-22
specialise conv net to hypercube

kchoo1118:


NetKet/Machine/hypercube_conv_layer.hpp
```diff
@@ -35,33 +35,36 @@ namespace netket {
...
-class Convolutional : public AbstractLayer<T> {
+class ConvolutionalHypercube : public AbstractLayer<T> {
   using VectorType = typename AbstractLayer<T>::VectorType;
   using MatrixType = typename AbstractLayer<T>::MatrixType;
   using VectorRefType = typename AbstractLayer<T>::VectorRefType;
   using VectorConstRefType = typename AbstractLayer<T>::VectorConstRefType;
 
   static_assert(!MatrixType::IsRowMajor, "MatrixType must be column-major");
 
-  std::shared_ptr<const AbstractGraph> graph_;
```
gcarleo: I think it's a bit risky if we don't store any graph object in here. 
This Layer should just take a Hypercube object and use its methods, not do everything from scratch! Are there methods missing in Hypercube that you need and can't use yet?

kchoo1118: Hmm but in principle here if use a custom graph, e.g. the j1j2 we can still use this conv net.

kchoo1118: And put stride as well which changes the graph

gcarleo: I agree, but we can't have a machine completely disconneted from the graph used for the simulation! We still use the graph for sampling/defining operators etc... If there are inconsistencies between the graph used in the machine and the one used for all the other tasks, it's going to be a problem! 
What are the methods you need from the graph?

kchoo1118: Well but the graph for each layer is not the same already if u have stride

kchoo1118: So i need to have a new graph for each layer?

gcarleo:
I am confused. What I mean is that in a good design at least the geometry
of the input layer should be consistent with all the other geometries used
in sapling etc... That's why I was asking about the requirements on graph.
What are the requirements?

kchoo1118:
Hmm coordtosite and sitetocoord

kchoo1118:
Maybe also the dimensions and the length of the cube

gcarleo:
OK. So can you just take a hypercube if those methods are implemented? :)

kchoo1118:
But the problem is the layer doesn’t know which layer it is whether it is the input layer or the last layer

kchoo1118:
Well if we just do stride 1 then the graph can be the same

kchoo1118:
but the problem is what if u decide to use a custom graph?

gcarleo:
Indeed maybe each layer should take a separate graph, to be fully
consistent.. But you don't allow for graphs that are diffenrt from
hypercube... Would this solve the issue?

kchoo1118:
no i mean lets say i want to j1j2 on square lattice so i use a custom graph. Then if the conv net only takes hypercube then i cant use the conv net.

kchoo1118:
the thing that is not fully consistent is: convolutions only need a lattice where as what we are specify is a graph with edges

gcarleo:
Then maybe the broken logic is that to do a j12 on a square lattice you
can't use hypercube!

kchoo1118:
yea the point is maybe we need to have a lattice class? 
And then define a graph on the lattice

gcarleo:
Ok... Maybe we need a lattice class. But what are the methods that you need
in a lattice that you don't have in a graph?

gcarleo:
A lattice is a special case of a graph. Basically you construct it from
basis vectors that you translate, with or without pbc. You can ask for site
coordinates too. Would the convnet layer work with a generic lattice or do
you need a hypercube lattice?

kchoo1118:
hmm i guess in principle we can make it work for a generic lattice but at the moment this conv net is only for hypercubes

PR #122, opened 2018-11-29
Save and load methods for machines

### LINKED ISSUE #114: Save and load objects from python
The last main design issue to be solved for v2.0 concerns saving and loading objects from python.

Pybind11 has some [pickling support](https://pybind11.readthedocs.io/en/stable/advanced/classes.html#pickling-support).

However, the design issue to be addressed is how to serialize objects stored internally as pointers.
Basically, each pickable object needs to define a GetState function, returning a python tuple of the arguments needed to construct the object. 

```c++
py::tuple GetState(const Pickleable &p) { 
            return py::make_tuple(p.Field1(),p.Field2(),...);
}
```
However, if the `Pickeable` stores a pointer to some abstract object (say Hilbert), then one obviously cannot do: 
```c++
py::tuple GetState(const Pickleable &p) { 
            auto hilbert= p.GetHilbert(); //NO! 
            return py::make_tuple(p.Field1(),p.Field2(),hilbert);
}
```

Suggestions are welcome.

gcarleo:
Adds save and load to machines. 
Also exposes some methods such as `parameters` as properties.
fixes #114

### DIFF
--- a/NetKet/Machine/abstract_machine.hpp
+++ b/NetKet/Machine/abstract_machine.hpp
@@ -169,7 +169,7 @@ class AbstractMachine {
   virtual void to_json(json &j) const = 0;
   virtual void from_json(const json &j) = 0;
 
-  void Save(std::string filename) const {
+  void Save(const std::string &filename) const {
     std::ofstream filewf(filename);
     Save(filewf);
     filewf.close();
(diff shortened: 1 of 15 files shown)


gcarleo:
Approved by @femtobit

PR #123, opened 2018-12-03
Concrete types

gcarleo:
This PR re-establishes the usage of concrete types for `Machine`, `Graph` etc, removing the interim usage of `std::shared_ptr` that was dictated by the pybind11 interface rather than NetKet design. 

Concrete types are based on `std::variant` (actually a non-standard version backported to c++14). 

This effectively moves NetKet from C++11 to C++14, which seems a reasonable move, given the current development of the compilers. Moving to C++17 is too premature because of sparse support for `std::variant` `std::optional` etc, especially on Mac platforms. 

Using `std::variant` should (maybe) also help a little bit with performance (to be seen). 
Working with concrete types also has many advantages when implementing python pickling. 

I am not entirely satisfied by this implementation because I wasn't able to define copy constructors for the concrete types, because of some obscure variant-related issue. Any help with this is much appreciated. 

The variant-based polymorphism is obtained like this: 

```c++
class Hilbert : public AbstractHilbert {
 public:
  using VariantType = mpark::variant<Spin, Boson, Qubit, CustomHilbert>;

 private:
  VariantType obj_;

 public:
  Hilbert(VariantType obj) : obj_(obj) {}

  bool IsDiscrete() const override {
    return mpark::visit([](auto &&obj) { return obj.IsDiscrete(); }, obj_);
  }
.....
```

Since we are working with objects, it would be useul also to use more thoroughly `std::move` than what I already did.


twesterhout:
> This PR re-establishes the usage of concrete types for Machine, Graph etc, removing the interim usage of std::shared_ptr that was dictated by the pybind11 interface rather than NetKet design.
> 
> Concrete types are based on std::variant (actually a non-standard version backported to c++14).

@gcarleo as you correctly pointed out when I suggested we moved to separate compilation, right now is probably not the time to do it. There are already a lot of changes in v2.0. Also, let's be honest, v2.0 is currently hanging by a thread. We have almost no proper tests... I suggest we stabilize v2.0 first. I'm not sure there's anything to gain in using `variant` (except for adding a whole lot of templates to the code), but I haven't thought about it too deeply yet. Anyway, we can always introduce variant to the C++ code base without breaking the Python tests.



And a few comments for "the future":

> This effectively moves NetKet from C++11 to C++14, which seems a reasonable move, given the current development of the compilers. Moving to C++17 is too premature because of sparse support for std::variant std::optional etc, especially on Mac platforms.

How about [this implementation](https://github.com/martinmoene/variant-lite) of `variant<Ts...>` which supports C++11. We have enough breaking changes in v2.0, so I'd avoid moving to C++14 just yet.

> I am not entirely satisfied by this implementation because I wasn't able to define copy constructors for the concrete types, because of some obscure variant-related issue. Any help with this is much appreciated.

Defining copy and move constructors should be a walk in the park. `variant` does all the heavy-lifting. I can probably help you out here, if we go the `variant` route.

gcarleo:
> @gcarleo as you correctly pointed out when I suggested we moved to separate compilation, right now is probably not the time to do it. There are already a lot of changes in v2.0. Also, let's be honest, v2.0 is currently hanging by a thread. We have almost no proper tests... I suggest we stabilize v2.0 first.

What do you mean we have no proper test?! Actually almost the entire suite of tests has been ported to python! Testing the same functions on the C++ side does not make too much sense, I think. 
Sure, there is a few more tests we need to add, but it's not that dramatic. Actually, the only major thing we haven't stabilized yet is the input/output interface, but @femtobit is going to work on it soon. 

In any case, even if without variant, I would still keep types "concrete", and use the file structure I proposed here (and that actually we already had in v1.x). If we don't like variant in the future we can just internally switch to something else, without touching the interface of the various classes...
If we don't do this change now, it is going to take a lot more work in the future (and believe me, this has already taken a lot of work). 

> I'm not sure there's anything to gain in using variant (except for adding a whole lot of templates to the code), but I haven't thought about it too deeply yet. Anyway, we can always introduce variant to the C++ code base without breaking the Python tests.

This is to be measured, in principle we should avoid some virtual calls and if I understand correctly the compiler can put some stuff on the stack ?

> How about this implementation of variant<Ts...> which supports C++11. We have enough breaking changes in v2.0, so I'd avoid moving to C++14 just yet.

OK, but the other problem is that I am also using generic lambdas (see implementation snippet above). We can maybe do without that, but it's going to be really ugly? Maybe with templates?  
But anyways, what is your argument against C++14? Actually it makes sense to switch to a higher standard for v2.0 and not, say, v2.1. . The idea is that major releases only would change the standard

> Defining copy and move constructors should be a walk in the park. variant does all the heavy-lifting. I can probably help you out here, if we go the variant route.

Great, I am sure this is feasible, it's just that I didn't have more fuel to work on that! It seems that for some reason the copy constructor for `Machine` was deleted, and I couldn't understand why

gcarleo:
Actually, @twesterhout notice that all tests on Travis pass, this means that basically all the platforms we were testing on are already supporting C++14 :)

twesterhout:
> What do you mean we have no proper test?! Actually almost the entire suite of tests has been ported to python!

@gcarleo, okay, "almost no" was an exaggeration, I apologise :) But! We have no tests with invalid input. `test_machine.py`, `test_sampler.py`, `test_operator.py` and maybe other files still use dictionaries, i.e. it's not clear that setting `Hilbert` to `None` won't terminate with a seg fault. There are no tests for `log_val`. There are no tests that `Hilbert` objects actually return the correct local states, stuff like fixed magnetisation etc. There are no tests for `ExactSampler` which is used to test all other samplers. No tests for `Operator`s, `Hamiltonian`s, `GroundState`, `Dynamics`. Many of these are probably in a "work in progress"-state, but still.

> If we don't do this change now, it is going to take a lot more work in the future (and believe me, this has already taken a lot of work).

Oh, I do appreciate the effort! I fail to see how doing this in the future will be _a lot more_ work, but I'll take your word for it for now.

---------
__`<important part>`__

> In any case, even if without variant, I would still keep types "concrete", and use the file structure I proposed here (and that actually we already had in v1.x)

Concrete types are nicer to work with, I won't argue with that. But you're not just introducing concrete types, you're changing the ownership semantics completely. E.g. graphs, Hilbert spaces etc. are now copied all over the place. Could you, please, list concrete reasons why e.g. "sharing" the graph between a `Hilbert` object and a `Machine` is wrong? Why I now need to copy my machine every time I want to sample from it? It least to me it seems obvious that a sampler shouldn't own a machine, but maybe I'm missing something...

__`</important part>`__

--------

> This is to be measured, in principle we should avoid some virtual calls and if I understand correctly the compiler can put some stuff on the stack ?

It's actually way trickier than that :) My point is that performance is definitely not among the reasons to switch to sum types.

> OK, but the other problem is that I am also using generic lambdas (see implementation snippet above). We can maybe do without that, but it's going to be really ugly? Maybe with templates?

Well, yes, C++11 version requires a bit more boilerplate. I.e. `[...] (auto x) { ... x ...; }` becomes
```cpp
struct ... {
    template <class T>
    auto operator(T x) -> ... { ... x ...; }
    ...
};
```
There are also things to be gained, though. You can't have a `constexpr` lambda in C++14.

> But anyways, what is your argument against C++14? Actually it makes sense to switch to a higher standard for v2.0 and not, say, v2.1. . The idea is that major releases only would change the standard

That there's quite little to gain. It's simpler to directly switch to `C++17` then. I've been using it by default for more than a year now. Sure, there are some things missing from the standard library, but the language part works really well.

gcarleo:
> Concrete types are nicer to work with, I won't argue with that. But you're not just introducing concrete types, you're changing the ownership semantics completely. E.g. graphs, Hilbert spaces etc. are now copied all over the place. Could you, please, list concrete reasons why e.g. "sharing" the graph between a Hilbert object and a Machine is wrong? Why I now need to copy my machine every time I want to sample from it? It least to me it seems obvious that a sampler shouldn't own a machine, but maybe I'm missing something...

@twesterhout indeed you have a very good point here. I agree. 
However, as I was mentioning it should be possible to use `shared_ptr` internally instead of `variant` and safely (hopefully) pass a reference of `Machine` to the `Sampler` etc, in those cases where we actually need to share resources. How the thing is internally handled (i.e. with a variant, with a pointer, with a shared pointer etc) should be absolutely hidden to other C++ classes, and that's the main reason of this PR (the variant is really just a proposal, especially useful for those classes that could be stack allocated, and for those basically immutable types such as `Graph`, `Hilbert` etc). 

The reason why I want to isolate the handler from the interface, is that basically now all the samplers (for example) are constructed from a shared pointer to `AbstractMachine` . So, in this sense, **everybody** needs to know how the Machine is handled... I just want to get rid of that, because it is not very sustainable + it is not how we would do things in C++ anyways (what we really wanted to do on the C++ side was to pass a reference to `Machine` in `Sampler`

gcarleo:
> Well, yes, C++11 version requires a bit more boilerplate. I.e. [...] (auto x) { ... x ...; } becomes

struct ... {
    template <class T>
    auto operator(T x) -> ... { ... x ...; }
    ...
};
There are also things to be gained, though. You can't have a constexpr lambda in C++14.

OK, I am not really strongly in favor of C++14, if there is a solution that works in C++11, we can go for it. The only thing is that the template maybe might be painful to expose to pybind11...

twesterhout:
> However, as I was mentioning it should be possible to use shared_ptr internally instead of variant and safely (hopefully) pass a reference of Machine to the Sampler etc, in those cases where we actually need to share resources. How the thing is internally handled (i.e. with a variant, with a pointer, with a shared pointer etc) should be absolutely hidden to other C++ classes, and that's the main reason of this PR (the variant is really just a proposal, especially useful for those classes that could be stack allocated, and for those basically immutable types such as Graph, Hilbert etc).

So basically you're just against passing around (and storing) `shared_ptr`s. Then what exactly is your suggestion? (sorry, I'm a slow thinker)

* Do you want to get rid of `shared_ptr`s altogether? Then see my previous comment
* Do you want to just wrap a `shared_ptr` into a nicer (in your opinion) interface? Then, please, do explain why, because "everybody needs to know how the Machine is handled" in both cases. `shared_ptr` just makes things more explicit, and it's a matter of opinion whether that's good or bad.
* Do you want to use sum types for some interfaces and keep using `shared_ptr`s for the rest? That's __very__ confusing, because now I have to go look at the code (because there are no docs for the C++ code) every time I use on of these classes just to know whether it's shallow- or deep-copied. I'm very strongly against that!

> The only thing is that the template maybe might be painful to expose to pybind11...

I didn't say anything about exposing templates to pybind11 :) That's totally unnecessary.

gcarleo:
@twesterhout , to be concrete, I would like to have the equivalent of `Eigen::Ref`, IF I want to pass my objects by reference, and just the concrete type if I want to pass them by value. I also want these two things to be python safe (as much as `Eigen::Ref` is, by the way). How those two are internally implemented is not something that the user should know/worry about.
So, can we implement something along the lines of `Eigen::Ref`? You see, `Eigen::Ref` is very nice because the C++ code is completely transparent to the fact that you pass an actual reference, or this custom reference object. The only thing that matters is that `Eigen::Ref` doesn't have memory management problems in python + it has the same syntax of a normal reference

gcarleo:
> I didn't say anything about exposing templates to pybind11 :) That's totally unnecessary.

Oh that would be cool... I thought I needed to write a binding for each type separately, of course if you can avoid that we can then certainly stick to C++11

twesterhout:
@gcarleo, I'm not experienced with Eigen, but it looks like `Eigen::Ref` is what it it -- a reference (or I'm seriously misreading their code). I.e. it doesn't own the data. So how does it help us then?

twesterhout:
@gcarleo to verify my claim, I've created a simple test:
```cpp
struct Foo {
    Eigen::Ref<Eigen::VectorXf> data;
};

...

py::class_<Foo>(m, "Foo")
    .def(py::init([](Eigen::Ref<Eigen::VectorXf> x) { return Foo{x}; }))
    .def("at", [](Foo& x, int const i) { return x.data(i); });
```

```python
>>> import netket
>>> import numpy
>>> import gc
>>> x = numpy.array(list(range(1000)), dtype=numpy.float32)
>>> foo = netket.Foo(x)
>>> x[999]
1000.0
>>> foo.at(999) # Yes, the values are equal
1000.0
>>> x[999] = -1
>>> foo.at(999) # Yes, the data is shared
-1.0
>>> x = None # Destroy the array in the hopes that Eigen::Ref is destroyed as well
>>> gc.collect()
>>> foo.at(999) # NOTE: On my machine(!)
-1.0
>>> # Hm... so the memory hasn't been reused yet. Let's do this then
>>> for i in range(1000):
...     x = numpy.zeros((10000,), dtype=numpy.float32)
...
>>> foo.at(999) # NOTE: On my machine(!)
0.0
>>> # Ha!
```

You might have to play around with sizes a bit, but this proves what I wrote in the previous comment: `Eigen::Ref` has reference semantics and will not keep your data alive. So my question remains: how do you expect a NetKet analogue of `Ref` to help us with object life-time related issues?

PR #128, opened 2018-12-12
Remove dependence from mpi4py directly calling MPI_Init and MPI_Finalize

gcarleo:
A slightly different approach to what proposed by @twesterhout in #109 , with the help of @femtobit


twesterhout:
@gcarleo, so now you're using a global variable in Python to initialize MPI. What's the reason for it? Is `_mpi_environment` guaranteed to be initialised before any other NetKet function uses MPI? Are there guarantees that no other function will use MPI after `_mpi_environment` is destructed?

gcarleo:
@twesterhout I thought that putting it inside a module would guarantee that the member would be destructed by the garbage collector in python. It seems that unfortunately this is not the case, at least this should be the reason why the linux tests are failing here... Your solution to have a static variable however also doesn't have guarantees?

twesterhout:
> I thought that putting it inside a module would guarantee that the member would be destructed by the garbage collector in python.

@gcarleo I my experience it does not. It seems that Python's GC is not run when the interpreter is killed. But I'm not familiar with Python standard, so this might very well be wrong...

> Your solution to have a static variable however also doesn't have guarantees?

He-he, it most certainly does :) Otherwise I wouldn't have suggested it in the first place.

PR #129, opened 2018-12-14
Make samplers deterministically seedable

femtobit:
This PR makes the sampler classes seedable so that simulation results are deterministic when a fixed seed is provided. It also extracts the corresponding code to a reusable class, since it was the same for all samplers.

### DIFF
--- a/NetKet/Sampler/abstract_sampler.hpp
+++ b/NetKet/Sampler/abstract_sampler.hpp
@@ -39,20 +39,17 @@ class AbstractSampler {
   virtual const AbstractHilbert& GetHilbert() const noexcept = 0;
 
   virtual ~AbstractSampler() {}
-};
 
-template <class WfType>
-class SeedableSampler : public AbstractSampler<WfType> {
-  DistributedRandomEngine engine_;
-
- public:
   void Seed(DistributedRandomEngine::ResultType base_seed) {
     engine_.Seed(base_seed);
     this->Reset(true);
   }
 
  protected:
-  default_random_engine &GetRandomEngine() { return engine_.Get(); }
+  default_random_engine& GetRandomEngine() { return engine_.Get(); }
+
+ private:
+  DistributedRandomEngine engine_;
 };
 
 }  // namespace netket
--- a/NetKet/Sampler/abstract_sampler.hpp
+++ b/NetKet/Sampler/abstract_sampler.hpp
@@ -41,5 +41,19 @@ class AbstractSampler {
   virtual ~AbstractSampler() {}
 };
 
+template <class WfType>
+class SeedableSampler : public AbstractSampler<WfType> {
+  DistributedRandomEngine engine_;
+
+ public:
+  void Seed(DistributedRandomEngine::ResultType base_seed) {
+    engine_.Seed(base_seed);
+    this->Reset(true);
+  }
+
+ protected:
+  default_random_engine &GetRandomEngine() { return engine_.Get(); }
+};
+
 }  // namespace netket
 #endif
--- a/NetKet/Sampler/custom_sampler.hpp
+++ b/NetKet/Sampler/custom_sampler.hpp
@@ -30,7 +30,7 @@ namespace netket {
 
 // Metropolis sampling using custom moves provided by user
 template <class WfType>
-class CustomSampler : public AbstractSampler<WfType> {
+class CustomSampler : public SeedableSampler<WfType> {
   WfType &psi_;
   const AbstractHilbert &hilbert_;
   LocalOperator move_operators_;
@@ -38,8 +38,6 @@ class CustomSampler : public AbstractSampler<WfType> {
 
   // number of visible units
   const int nv_;
-
-  DistributedRandomEngine rgen_;
 
   // states of visible units
   Eigen::VectorXd v_;
@@ -67,17 +65,6 @@ class CustomSampler : public AbstractSampler<WfType> {
         hilbert_(psi.GetHilbert()),
         move_operators_(move_operators),
         nv_(hilbert_.Size()) {
-    Init(move_weights);
-  }
-
-  CustomSampler(WfType &psi, const LocalOperator &move_operators,
-                DistributedRandomEngine::ResultType seed,
-                const std::vector<double> &move_weights = {})
-      : psi_(psi),
-        hilbert_(psi.GetHilbert()),
-        move_operators_(move_operators),
-        nv_(hilbert_.Size()),
-        rgen_(seed) {
     Init(move_weights);
   }
 
@@ -125,7 +112,7 @@ class CustomSampler : public AbstractSampler<WfType> {
 
   void Reset(bool initrandom = false) override {
     if (initrandom) {
-      hilbert_.RandomVals(v_, rgen_.Get());
+      hilbert_.RandomVals(v_, this->GetRandomEngine());
     }
 
     psi_.InitLookup(v_, lt_);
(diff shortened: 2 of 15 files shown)


NetKet/Sampler/abstract_sampler.hpp
```diff
@@ -41,5 +41,19 @@ class AbstractSampler {
   virtual ~AbstractSampler() {}
 };
 
+template <class WfType>
```
gcarleo: I think we can reasonably assume all samplers have a Seed method (at least I can't think now of cases where a generic sampler wouldn't have a seed method, but maybe you had something specific in mind?). 
I would suggest to avoid creating a `SeeableSampler` class and just move the `Seed` abstract method in the `AbstractSampler` class

NetKet/Sampler/custom_sampler.hpp
```diff
@@ -108,31 +105,14 @@ class CustomSampler : public AbstractSampler<WfType> {
...
-
-    SendToAll(seeds);
-
-    rgen_.seed(seeds[mynode_]);
-  }
-
   void Reset(bool initrandom = false) override {
     if (initrandom) {
-      hilbert_.RandomVals(v_, rgen_);
+      hilbert_.RandomVals(v_, this->GetRandomEngine());
```
gcarleo: maybe better just to store the random engine as a private variable? the code would look nicer, maybe marginally faster too?

femtobit: The compiler should inline the call to `GetRandomngine()` so the performance should be identical. (Compare https://godbolt.org/z/Ny9zuB for a simplified test, where the same code is generated for `B::f1` and `B::f2`.)

According to the Google Style Guide, [data members should be private](https://google.github.io/styleguide/cppguide.html#Inheritance). Making `engine_` protected would work, but I think it's more confusing than inheriting the protected function.

NetKet/Sampler/abstract_sampler.hpp
femtobit: I wasn't completely sure whether every possible sampler needs random number generation, though I do not have a specific case in mind. So yes, I can move that to the base class.

NetKet/Sampler/custom_sampler.hpp
gcarleo: Alright good points!

NetKet/Utils/random_utils.hpp
```diff
@@ -40,6 +40,82 @@ void RandomGaussian(Eigen::Matrix<std::complex<double>, Eigen::Dynamic, 1> &par,
         std::complex<double>(distribution(generator), distribution(generator));
   }
 }
+
```
gcarleo: Small optional thing: we should maybe include <mpi.h> in this file now and remove it from the samplers, where it is no longer needed (nice!)

twesterhout:
@femtobit from a brief look at the code it's not obvious to me why the seeding is deterministic... Some RNGs (Mersenne Twister among them) have state of size much bigger than one machine word. Correlation issues aside, I guess that these two sequences of operations
```cpp
gen.Seed(67890)
....generate_pars(...)
gen.Seed(12345)
// go on using gen
```
and
```cpp
gen.Seed(67890)
gen.Seed(12345)
// go on using gen
```
will produce different results.

gcarleo:
@twesterhout , the seed bit affects only samplers. 
the `....generate_pars(...)` refers to the random initialization of the machine parameters? You can fix a seed in there too, or just use numpy random generators for those

In our experiment when you fix the seed of the sampler, and the seed of the random parameters, you get deterministic results

twesterhout:
@gcarleo oh no, why did you merge it already? :) I had other comments about the code...

gcarleo:
You can still comment, at your own risk :)

NetKet/Utils/mpi_interface.hpp
```diff
@@ -41,6 +41,14 @@ void SendToAll(std::vector<int> &value, int root = 0,
                const MPI_Comm comm = MPI_COMM_WORLD) {
   MPI_Bcast(&value[0], value.size(), MPI_INT, root, comm);
 }
+void SendToAll(std::vector<unsigned int> &value, int root = 0,
```
twesterhout: These functions are unnecessary, because (read on)

NetKet/Utils/random_utils.hpp
```diff
@@ -40,6 +40,82 @@ void RandomGaussian(Eigen::Matrix<std::complex<double>, Eigen::Dynamic, 1> &par,
...
+    MPI_Comm_rank(MPI_COMM_WORLD, &rank_s);
+    int size_s;
+    MPI_Comm_size(MPI_COMM_WORLD, &size_s);
+    const auto rank = static_cast<size_t>(rank_s);
+    const auto size = static_cast<size_t>(size_s);
+
+    std::vector<ResultType> seeds;
+    seeds.resize(size);
+
+    default_random_engine seed_engine(base_seed);
```
twesterhout: No need to create a new engine. What we really want is `engine_.seed(base_seed)`.

NetKet/Utils/random_utils.hpp
```diff
@@ -40,6 +40,82 @@ void RandomGaussian(Eigen::Matrix<std::complex<double>, Eigen::Dynamic, 1> &par,
...
+   */
+  ResultType GetDerivedSeed(ResultType base_seed) {
+    int rank_s;
+    MPI_Comm_rank(MPI_COMM_WORLD, &rank_s);
+    int size_s;
+    MPI_Comm_size(MPI_COMM_WORLD, &size_s);
+    const auto rank = static_cast<size_t>(rank_s);
+    const auto size = static_cast<size_t>(size_s);
+
+    std::vector<ResultType> seeds;
```
twesterhout: No need to keep a vector of seeds and no need for dynamic memory allocation.

NetKet/Utils/random_utils.hpp
```diff
@@ -40,6 +40,82 @@ void RandomGaussian(Eigen::Matrix<std::complex<double>, Eigen::Dynamic, 1> &par,
...
+    std::uniform_int_distribution<ResultType> dist;
+
+    if (rank == 0) {
+      for (std::size_t i = 0; i < size; ++i) {
+        seeds[i] = dist(seed_engine);
+      }
+    }
+
+    MPI_Barrier(MPI_COMM_WORLD);
+    SendToAll(seeds);
```
twesterhout: There is no need for process 5 to know the seed of process 10. The goal here is for process 0 to create a seed and send it to process 1, then create another and send it to process 2, etc. MPI has point-to-point communication for such tasks, no need for broadcasts.

If you want to save a bit of time, then there is no need for communication at all! Just seed all engines with base_seed and then do
```cpp
auto final_seed = base_seed;
for (auto i = 0; i < rank_s; ++i) {
    final_seed = dist(engine_);
}
engine_.seed(final_seed);
```

twesterhout:
p.s. @gcarleo I had a look at `libc++` implementation of `seed` and indeed it looks like the whole state is initialised. So the only problem with using but one word is correlation.

NetKet/Utils/random_utils.hpp
femtobit: That would work as well, of course. Since this is not performance critical code, I do not think it makes a noticeable difference.

NetKet/Utils/random_utils.hpp
femtobit: It is true that both storing the vector of seeds and the MPI communication is not necessary. I left it mostly because it was done this way in the old code. I might change it later on.

gcarleo: I think it is safe to assume that storing N~100/1000 integers is not of a big deal

twesterhout:
> Further, the seed function is defined such that if we have an engine e of type E, then after e.seed(s) it holds that e == E(s).

Aha, I didn't know the standard required that. Thanks @femtobit!

Issue #107, opened 2018-11-19
Restructure the optimization driver interface for v2.0


gcarleo:
Version 2.0 calls for a re-thinking of the way we handle the output. At the moment, this is done in the (excellent) json output writer developed by @femtobit. 

I think that in 2.0 we should go beyond the json output. 

As a first goal, I propose to go for something similar to what is done in [Keras](https://keras.io/visualization/) , where the training history is saved in python dictionaries. 

```python
import matplotlib.pyplot as plt
... construct various objects here ... 
history = vmc.run(n_iter=300,n_samples=1000,save_every=10)

# Plot energies 
plt.plot(history.history['energy'].mean) #or something similar for plot with error bars 
plt.show()
```

As a second (slightly more ambitious) goal, it would be really cool to have [callbacks](https://keras.io/callbacks/) à la Keras as well. One can use these callback to print each step on a file, or, even better, to visualise the results dynamically. Once the callbacks are in place, live streaming of the results in a browser can be easily achieved using [Bokeh](https://bokeh.pydata.org/en/latest/) (see [here](https://stackoverflow.com/questions/37724660/streaming-two-line-graphs-using-bokeh) for an example). This would be really cool to have. 

@femtobit what do you think about goal 1 for example? Does it look like an easy modification of your output handlers?

femtobit:
I do agree with the goal of providing different forms of output.
When seeing a line like `history = vmc.run(n_iter=300,n_samples=1000,save_every=10)` it makes me wonder whether it makes sense to *return* it from the run function. It would be more useful, I think to be able to monitor the output in real time (like you also suggest with the callbacks).
But that is step 2, I think.

If we just want to load the history after the simulation is finished (i.e., step 1), it is very easy anyway to access the output so I would advise against returning a history object directly from the `driver.run` functions. We could write a convenience function, e.g.,
```python
op = nk.JsonOutputWriter(logfile, statefile)
vmc.run(..., output=op)
history = op.load_all_history()
```
where `load_all_history` would do something like (of course, in reality it'd have to be implemented as part of the pybind11 bindings, but I will write python code here for readability)
```python
import json
class JsonOutputWriter:
    # [...]
    def load_all_history(self):
        return json.load(self.logfilename)
```
[...]

Issue #86, opened 2018-11-13
Add standard typedefs for matrices and vectors


femtobit:
Suggestion for discussion:

There are many places in the codebase where we use typedefs such as `using Vector = Eigen::VectorXcd` or `using Matrix = Eigen::MatrixXcd` which I think is somewhat redundant and can be a bit tedious.

I suggest adding a couple of standard typedefs to `netket.hpp` or a custom header file, such as (current names are just for illustration)
```
using NkVector = Eigen::VectorXcd;
using NkMatrix = Eigen::MatrixXcd;

// if needed:
using NkRealMatrix = ...
using NkRealVector = ...
```
It could also be convenient to have a shorthand for `std::complex<double>`.

Of course, there might be legitimate cases where something else should be used (e.g., fixed-size Eigen vectors for performance reasons), but I think we do not loose much by using the types above by default and declaring them globally and only adding template parameters or different typedefs if they are really needed. (Compare the [YAGNI principle](https://en.wikipedia.org/wiki/You_aren't_gonna_need_it)).

This could also help reduce the amount of template parameters we have which are only ever instantiated with one type.

twesterhout:
@femtobit good point! I'd say something like this
```cpp
namespace netket {
    using Eigen::VectorXcd;
    using Eigen::MatrixXcd;
    ...
} // namespace netket
```
should do the trick, no need to invent new names.
The question where to put them is a more interesting one. Perhaps it's time we created a light-weight `NetKet/config.hpp` and put these typedefs there... I'm somewhat concerned about putting an `#include <Eigen/Core>` into a "config" file, but well, that's life ;)

PR #131, opened 2018-12-14
Basic iterator interface for optimization drivers

femtobit:
This PR contains a first version of the iterator interface for the VMC (and exact imaginary time) driver, addressing #107.
(The implementation is partially modeled after the description [here](https://blog.galowicz.de/2017/07/02/order2_iterator/).)

Basically, it allows to write code such as
```cpp
VariationalMonteCarlo vmc(...);
for (const auto& step : vmc.Iterate()) {
    outfile << 'step=' << step.index << '\n';
    outfile << json(step.observables) << '\n';
}
```
In principle, it is possible to modify the state, Hamiltonian, or other objects in between iteration steps (though this is not done anywhere right now).
The Python bindings expose this function as well, allowing, e.g.,
```python
vmc = nk.gs.vmc.Vmc(...)
for i, step in enumerate(vmc.iter()):
    print(step.observables.Energy)
```

In order to expose `step.observables`, Python bindings for the `ObsManager` class are provided, with an interface similar to a python `dict` (similar enough to make `d = dict(step.observables)` work).

Some open issues (or potentially points to discuss):

1.  The (VMC) iterators dereference to a `struct` containing basic information on the state:
https://github.com/netket/netket/blob/a5c6d62cf46e3cbee63073190e53d23ede0c68bd/NetKet/GroundState/variational_montecarlo.hpp#L98-L103
This contains a copy of the data (as opposed to referencing the corresponding internal state of the `VariationalMonteCarlo` class) in order to make it possible to write code such as
`steps = list(vmc.iter(100))`.
This has the downside of copying this data every step, regardless of whether it is needed. For the machine parameters, an option is provided, so that `step.parameters` can be left empty (`== nonstd::nullopt`) if storage of the machine parameters is not needed.
1. Extracting the data from `ObsManager` still performs an MPI reduction which has to be performed on all MPI processes, leading to code such as
https://github.com/netket/netket/blob/00a7c60382f9fc1005100a9613379d9b5c8456dc/Tutorials/PyNetKet/ground_state_iter.py#L40-L45
The code should probably be changed so that `step.observables` does not perform MPI calls. The question is whether it should contain sensible values only on rank 0 or on all MPI processes (the first option should be sufficient but might also lead to annoying-to-debug errors.)
1. The `VariationalMonteCarlo::Run` function is provided which is now implemented in terms of the iterator interface. This can be used to get essentially the behaviour of NetKet v1:
https://github.com/netket/netket/blob/00a7c60382f9fc1005100a9613379d9b5c8456dc/Tutorials/PyNetKet/ground_state.py#L42-L50 We can discuss whether we want to provide this function or maybe a different kind of simplified interface.
1. This PR also contains a commit introducing a header for common type aliases (related to #86). For now, it contains `Complex = std::complex<double>` (because I think it is cumbersome to always spell-out that type) and `Index`. The `Index` type is added because right now there are several types used as indices (usually `int` or `std::size_t`) in different places in the netket codebase. I suggest standardizing on `Index = std::ptrdiff_t` (which is a 64-bit integer on my system). (Whether signed or unsigned types should be used as array indices is a somewhat controversial question in C++, but note that newer additions to the standard such as `std::span` uses `ptrdiff_t`. This is also in line with what the Google Style Guide [has to say about unsigned integers](https://google.github.io/styleguide/cppguide.html#Integer_Types).)

Let me know what you think.


gcarleo:
Thanks a lot @femtobit, this is excellent work, and the interface is very cool! 

Just let me add a few comments here, before going into the review: 

> The (VMC) iterators dereference to a struct containing basic information on the state:

Just to clarify: if I do something like 
```python
for st in vmc.iter(10, 2):
   obs = st.observables
   ... 
```

are the observables computed at every step or only every 2 steps? (The energy needs anyway to be computed at every step, but maybe we can trigger the calculation of other observables, as added with `add_observable` only every n steps (2 in this case) ? 

On a related note, I think we could just remove `parameters` from the step object entirely, since those are a property of the machine. In practice you can always do `print(ma.parameters)` to retrieve them if you want at any point. The same is true for the acceptance, which is a property of the sampler, I would remove that too. I think we can just keep the observables in the stepper? 

I understand that this would break the possibility of retrieving the parameters/acceptance doing `steps = list(vmc.iter(100))`, but in practice the later usage of `iter` doesn't seem to me very conventional, since in any case one can expect the single iterations to be time consuming. 

> The code should probably be changed so that step.observables does not perform MPI calls. 

I think that, if possible, calling `step.observables` should trigger itself the calculation of the observables from the stored visible units in the `Vmc` (see previous comment). However I am not sure it is possible at all to remove MPI calls from there, since MPI is used to collect the averages from different processors.

NetKet/GroundState/pyground_state.hpp
```diff
@@ -31,33 +31,77 @@ namespace netket {
 
 void AddGroundStateModule(py::module &m) {
   auto subm = m.def_submodule("gs");
```
gcarleo: I think that `netket.gs.vmc.Vmc` is too deep of a module hierarchy, we should try to simplify it. 
`netket.Vmc` doesn't look bad to me, but let's discuss more

NetKet/GroundState/pyground_state.hpp
```diff
@@ -31,33 +31,77 @@ namespace netket {
...
-           py::arg("optimizer"), py::arg("n_samples"), py::arg("niter_opt"),
-           py::arg("output_file"), py::arg("discarded_samples") = -1,
+           py::arg("optimizer"), py::arg("n_samples"),
+           py::arg("discarded_samples") = -1,
            py::arg("discarded_samples_on_init") = 0, py::arg("method") = "Sr",
            py::arg("diag_shift") = 0.01, py::arg("rescale_shift") = false,
-           py::arg("use_iterative") = false, py::arg("use_cholesky") = true,
-           py::arg("save_every") = 50)
+           py::arg("use_iterative") = false, py::arg("use_cholesky") = true)
+      .def_property_readonly("psi", &VariationalMonteCarlo::GetPsi)
```
gcarleo: For consistency we should name this `machine` and not `psi` (my fault, I know)

NetKet/GroundState/pyground_state.hpp
```diff
@@ -31,33 +31,77 @@ namespace netket {
...
+        str << "<VmcStep: step=" << self.index << ">";
+        return str.str();
+      });
+
+  py::class_<VariationalMonteCarlo::Iterator>(vmc, "Iterator")
+      .def("__iter__", [](VariationalMonteCarlo::Iterator &self) {
+        return py::make_iterator(self.begin(), self.end());
+      });
+
+  auto excact = subm.def_submodule("exact");
```
gcarleo: same here, we could do `netket.ed.ImagTimePropagation` and `netket.ed.SparseDiag` etc?

Test/GroundState/exact_diagonalization_input_tests.hpp
```diff
@@ -57,8 +57,8 @@ std::vector<netket::json> GetExactDiagonalizationInputs() {
   std::vector<std::vector<double>> szsz = {
```
gcarleo: I think this file can be just removed

NetKet/GroundState/pyground_state.hpp
```diff
@@ -31,33 +31,77 @@ namespace netket {
...
-      .def("run", &VariationalMonteCarlo::Run);
-
-  py::class_<ImaginaryTimeDriver>(subm, "ImaginaryTimeDriver")
-      .def(py::init<ImaginaryTimeDriver::Matrix &,
-                    ImaginaryTimeDriver::Stepper &, JsonOutputWriter &, double,
-                    double, double>(),
-           py::arg("hamiltonian"), py::arg("stepper"), py::arg("output_writer"),
-           py::arg("tmin"), py::arg("tmax"), py::arg("dt"))
-      .def("add_observable", &ImaginaryTimeDriver::AddObservable,
+      .def("run", &VariationalMonteCarlo::Run, py::arg("filename_prefix"),
```
gcarleo: maybe `filename_prefix` should be `output_prefix` or just `output` as before?

Tutorials/PyNetKet/ground_state_iter.py
```diff
@@ -0,0 +1,47 @@
...
+    optimizer=op,
+    n_samples=1000,
+    diag_shift=0.0,
+    method='Sr')
+
+comm = MPI.COMM_WORLD
+mpi_rank = comm.Get_rank()
+
+for i, st in enumerate(vmc.iter()):
+    obs = dict(st.observables) # TODO: needs to be called on all MPI processes
```
gcarleo: do we need to explicitly convert to `dict` here?

Tutorials/PyNetKet/ground_state_iter.py
```diff
@@ -0,0 +1,47 @@
...
+    if mpi_rank == 0:
+        print("step={}".format(i))
+        print("acceptance={}".format(list(st.acceptance)))
+        print("observables={}\n".format(obs))
```
gcarleo: It seems we cannot do `observables.Energy` is that true that we can use "only" `observables["Energy"]` here? Not a big deal anyways, it's just to understand...

Tutorials/PyNetKet/test_tutorials.py
```diff
@@ -0,0 +1,40 @@
+# This files imports all tutorial examples to make them testable
```
gcarleo: Oh, thanks for this, before indeed it was a mess!

NetKet/GroundState/pyground_state.hpp
femtobit: I think it would make sense to separate the VMC code from the ED code. So `netket.vmc` and `netket.ed` seem fine to me.

femtobit: Although strictly speaking `ed` is a bit of a misnomer for exact time evolution.

Tutorials/PyNetKet/ground_state_iter.py
femtobit: `observables.Energy` works as well.

Tutorials/PyNetKet/test_tutorials.py
femtobit: Though I would not recommend running this as part of the regular tests because the tutorials take very long. However, I found it helpful to discover whether a commit breaks the tutorials.

femtobit:
> On a related note, I think we could just remove parameters from the step object entirely, since those are a property of the machine. In practice you can always do print(ma.parameters) to retrieve them if you want at any point. The same is true for the acceptance, which is a property of the sampler, I would remove that too. I think we can just keep the observables in the stepper?
> 
> I understand that this would break the possibility of retrieving the parameters/acceptance doing steps = list(vmc.iter(100)), but in practice the later usage of iter doesn't seem to me very conventional, since in any case one can expect the single iterations to be time consuming.

Yes, I think this is a valid option.

> I think that what could be useful is to do
> `vmc.add_observable(ma.parameters)`
> `vmc.add_observable(sa.acceptance)`
> so that one can chose what optional things to measure.
> 
> One could even do
> `vmc.add_observable(datetime.datetime.now)`

Looks fun but I'd prefer to keep the observables restricted to actual quantum mechanical observables, i.e., operators and do
```python
for step in vmc.iter():
    date = datetime.datetime.now()
    # ... etc ...
```
What bothers me a bit is that in principle, all the information can be obtained from elsewhere, so there is no real need for the `step` object. (The observables can be currently only retrieved from `step`, but adding something like `sampler.mean(operators)` is possible.) But then again, in a for-loop one should iterate over *something*.

> are the observables computed at every step or only every 2 steps? (The energy needs anyway to be computed at every step, but maybe we can trigger the calculation of other observables, as added with add_observable only every n steps (2 in this case) ?

Good point. I think they are calculated every step right now. I will change that to compute them only when necessary.

gcarleo:
> Looks fun but I'd prefer to keep the observables restricted to actual quantum mechanical observables, i.e., operators and do

Indeed that's true, that would better be some `add_to_log` method or similar rather than `add_observable` that should strictly be for quantum observables 

> What bothers me a bit is that in principle, all the information can be obtained from elsewhere, so there is no real need for the step object. (The observables can be currently only retrieved from step, but adding something like sampler.mean(operators) is possible.) But then again, in a for-loop one should iterate over something.

True, (to retrieve observables one could also just do `vmc.observables` or something, by the way) but using these `add_xx` methods is useful if one wants to just use `vmc.run(..)` and record custom things during the run/dump them somewhere on a file. 

But I agree that the iterator is convenient, yes, but maybe just a `range(n_iter)` would have worked :) (see also my original comment in #107 )

NetKet/GroundState/pyground_state.hpp
femtobit: Yes, `netket.exact` sounds good.

Regarding `netket.vmc`: I see your point, however, we might get `netket.TdVmc` in the future. (In principle, we could also call it `netket.vmc.TimeEvolution` or similar.) Also, there is `netket.vmc.Iterator` (which users would usually not see, but has to be somewhere).

Tutorials/PyNetKet/ground_state_iter.py
femtobit: The opposite, actually: With a dict, `obs.Energy` does not work, with `ObsManager` it does (since the later provides the `__getattr__` method.

femtobit:
> But I agree that the iterator is convenient, yes, but maybe just a range(n_iter) would have worked :) (see also my original comment in #107 )

Of course, it is essentially equivalent to something like
```python
for i in range(1000):
    vmc.advance(step_size)
    print(vmc.observables)
```
Maybe we should have implemented the iterators in pure python and just provide the `Advance` function from C++. Then, implementing the iterator is a matter of writing something like
```python
def iter_obs(vmc):
    while True:
        vmc.advance()
        yield vmc.observables()

for obs in iter_obs(vmc):
    print(obs)
```
But well, in the end the interface is similar and providing control over the simulation to the surrounding code (which was the main goal) can be achieved either way.

For now, I think we keep `add_observables` the way it is and let `step` contain the corresponding expectation values and if more information is needed, users can obtain it from within the loop body.

femtobit:
> Thanks @femtobit , just a reminder to discuss later about adding the observables as a member of vmc, and leave the vmc iterator just as an index counting to advance the iterations

This is done now:
https://github.com/netket/netket/blob/3ddd9a256302f7add3a65f4f77baf7cfd3cced42/Tutorials/PyNetKet/ground_state_iter.py#L40-L42
The observable stats are only computed if the method is called. I chose to use this a bit more verbose function name (instead of, say, a property `vmc.observables`) because this call actually performs computation and the MPI reduction.

femtobit:
Merged into feature branch, so that this code can serve as base for other PRs

PR #132, opened 2018-12-16
Constant shift in LocalOperator and fix QuantumStateReconstruction

gcarleo:
This PR adds the possibility to have a constant shift in `LocalOperator`, such that 
in general one can have LocalOp = \sum_i O_i + constant. 

Algebraic operations on LocalOperator have been updated accordingly. This also removes the ambiguity when constructing a `LocalOperator` without arguments. 

This PR also fixes a few issues in the `QuantumStateReconstruction` application, mostly existing because of the copy constructor in `LocalOperator`  introduced in #117. This has now been removed because intrinsically unsafe (we can't easily copy Hilbert objects) and was most likely giving rise to some issues when creating vectors of `LocalOperator`, as needed by `QuantumStateReconstruction`.


NetKet/Operator/local_operator.hpp
```diff
@@ -284,18 +304,34 @@ class LocalOperator : public AbstractOperator {
...
+  }
+
+  LocalOperator &operator+=(const LocalOperator &rhs) {
+    assert(rhs.hilbert_.LocalStates().size() ==
+           this->hilbert_.LocalStates().size());
+
+    this->sites_.insert(this->sites_.end(), rhs.sites_.begin(),
+                        rhs.sites_.end());
+    this->mat_.insert(this->mat_.end(), rhs.mat_.begin(), rhs.mat_.end());
+    this->constant_ += rhs.constant_;
```
jamesETsmith: Is there a reason to add the level shift here? Shouldn't the level shift be consistent for the entire system or is there a situation where you would be adding together two operators with different level shifts?

gcarleo: here you add two operators: Osum= O1+O2, where O1=constant1 + sum_i o_i and, O2=constant2 +sum_i o'_i , thus here I am adding the two constants such that constant_sum = constant1+constant2

jamesETsmith: Ok, I'm used to level shifts being applied consistently to all operators, but if we want to allow different level shifts, then this works for me.

gcarleo:
Unless there are objections (@GTorlai ?) I will merge this ASAP, because that issue also broke the ability to correctly measure **any** local observable after machine optimization

gcarleo:
OK great !

PR #133, opened 2018-12-17
Adding support for CMake args to setuptools-based build

twesterhout:
#90 version 2.0

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,122 +1,155 @@
-#########################
-# project configuration #
-#########################
-
-# C++ project
-language: cpp
-
-dist: trusty
+language: python
 sudo: required
-
-################
-# build matrix #
-################
 
 matrix:
   include:
+    - os: osx
+      osx_image: xcode9.4
+      language: generic
+      env: PYTHON=2.7 CMAKE_FLAGS="-DCMAKE_BUILD_TYPE=Debug -DCMAKE_CXX_COMPILER=clang++"
 
-  # clang sanitizer
-  - os: osx
-    osx_image: xcode9.4
-    compiler: clang
-    env:
-      - CMAKE_OPTIONS=-DNETKET_Sanitizer=ON
+    - os: osx
+      osx_image: xcode10.1
+      language: generic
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -1,122 +1,159 @@
...
+  # TODO(twesterhout): Remove the following line when #109 is merged.
+  - python -m pip --no-cache-dir install mpi4py
+  - python -m pytest --version
+  - NETKET_CMAKE_FLAGS="$CMAKE_FLAGS" python -m pip install -v .
 
-  # show OS/compiler version
-  - uname -a
-  - $CXX --version
+script:
+  - python -c 'from mpi4py import MPI'
```
gcarleo: Is this a leftover?

twesterhout: No, this line is meant to remain until we remove the dependency on `mpi4py`. It helps with narrowing down MPI-related problems. If you insist, I can remove it of course, but would prefer not to.

femtobit: I agree that it's good to have this. If MPI fails independently of netket it is good to see that early on in the tests.

.travis.yml
```diff
@@ -1,122 +1,159 @@
...
+  - python -c 'from mpi4py import MPI; import netket; g = netket.graph.Hypercube(4, 1); h = netket.hilbert.Spin(g, 0.5); m = netket.machine.RbmSpin(h, 10)'
```
gcarleo: But is also this one needed?

gcarleo:
@twesterhout do I understand correctly that the sanitizer doesn't work yet? (Just to know what's the problem there, we will try to fix that one in another PR, this one is already good for many other reasons!)

.travis.yml
twesterhout: Yes. As you remember `pytest` has uses all kinds of traps, so sometimes this is the easiest way to get some meaningful error messages about low-level stuff such as segfaults and MPI initialisation errors.

twesterhout:
> @twesterhout do I understand correctly that the sanitizer doesn't work yet? (Just to know what's the problem there, we will try to fix that one in another PR, this one is already good for many other reasons!)

@gcarleo yes, this is correct. We discussed sanitizers in #90. [This build](https://travis-ci.org/netket/netket/jobs/469221080) can be used as a visualisation of the problem.

PR #137, opened 2018-12-19
Merge iterator interface into v2.0 branch

femtobit:
Continuing the discussion of #131.


NetKet/Operator/abstract_operator.hpp
```diff
@@ -30,12 +30,12 @@ namespace netket {
 struct ConnectorRef {
   /// The matrix element H(v,v')
   std::complex<double> weight;
```
gcarleo: I think this is called mel (for matrix element) in the other case :)

gcarleo:
Thanks a lot for this @femtobit, this is great work! Thanks also @alexwie for the final touches on the ED, it now looks very nice.

Issue #130, opened 2018-12-14
Add python documentation 


gcarleo:
As we move towards alpha/beta release for version 2.0, we need to have complete documentation for the new python functions. 

The idea is to use the nice parser developed by @ooreilly [here ](https://github.com/ooreilly/mydocstring/blob/pybind/mydocstring/extract.py). This parser + some markdown templates will be used to automatically generate the files for the docs website of the site, [here](https://www.netket.org/docs/home/). 

To do so, the python functions exposed through pybind11 should support a minimally modified version of the [google-style docstrings](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html).

A simple function with signature
```
operation(self: python_example.Operations, i: int, j: int, op_name: str='add') -> int
```
would have an associated docstring: 
```
      A function performing one of the allowed operations between two integers.

      Args:
          i (int): The first parameter.
          j (int): The second parameter.
          op_name (str='add'): The type of operation.

      Returns:
          int: If op_name=='add' it perform the sum param1+param2, otherwise 0.
```

Help is welcome to add documentation following this style, once we finalize the parser bit. 

One can add docstrings to pybind11 functions like this:
```c++
m.def("operation", &operation, R"EOF(
      A function performing one of the allowed operations between two integers.

      Args:
          i (int): The first parameter.
          j (int): The second parameter.
          op_name (str='add'): The type of operation.

      Returns:
          int: If op_name=='add' it perform the sum param1+param2, otherwise 0.
)EOF");
```

PR #141, opened 2018-12-29
Docs

gcarleo:
This PR introduces the tools necessary to extract the documentation from pybind-generated python classes. (see also issue #130) 

These tools are heavily based on [this project](https://github.com/ooreilly/mydocstring), and @ooreilly has played an important role in adapting his library to work with pybind11 generated docs. Thank you!

The tools are in `/Docs` and can be used from command line. For example, to generate markdown docs for a given NetKet class, one can do:  

```
python3 make_class_docs.py netket.graph.Hypercube > test.md 
```

Notice that pytablewriter is needed, and can be easily installed with `pip install pytablewriter`. 

At the moment we only have docs for `netket.graph` classes. Generated markdown documentation can be found in `Docs/Graph/`, for example [here](https://github.com/gcarleo/netket/blob/docs/Docs/Graph/Hypercube.md)

TODO: 
1. The tool `format.py` is not taking into account class functions (it only formats the constructors and class members).  Extending it to class-bound functions should be straightforward. 
2. NetKet website will use these markdown files to build the doc [pages here](https://www.netket.org/docs/home/)

**Action item**
We need to add documentation for all the other classes. Please volounteer to provide documentation for a few NetKet classes, so that we can quickly move to the final release! 
The documentation should follow the Google docstring format, as done for example in `NetKet/Graph/pygraph.hpp`

### DIFF
--- a/Docs/Graph/CustomGraph.md
+++ b/Docs/Graph/CustomGraph.md
@@ -1,0 +1,36 @@
+# CustomGraph
+In addition to built-in graphs, NetKet provides the freedom to define custom graphs, specifying a list of edges.
+## Constructor
+Constructs a new graph given a list of edges.
+
+|    Field    |       Type       |                                                                                                                                                                                                                                     Description                                                                                                                                                                                                                                     |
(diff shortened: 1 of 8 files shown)


jamesETsmith:
@gcarleo, would you like additional documentation in separate pull requests?

gcarleo:
@jamesETsmith happy new year, and thanks for looking into this! Yes please, let's merge this one first and then just open another PR with further docs you want to add

PR #142, opened 2019-01-02
Adding Python Documentation for Hilbert, Operator, and Ground State

jamesETsmith:
Adding documentation to `py*.hpp` files and generating markdown doc files.


NetKet/Hilbert/pyhilbert.hpp
```diff
@@ -46,28 +53,130 @@ void AddHilbertModule(py::module &m) {
...
-           py::arg("graph"), py::arg("s"))
+           py::arg("graph"), py::arg("s"), R"EOF(
+           Constructs a new ``Spin`` given a graph and the value of each spin.
+
+           Args:
+               graph: Graph representation of sites.
+               s: Spin at each site. Must be integer or half-integer.
+
+           Examples:
+               Simple fermionic spin hilbert space.
```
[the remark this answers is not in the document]
jamesETsmith: Ah I see, I'll change this

Docs/Operator/BoseHubbard.md
```diff
@@ -0,0 +1,35 @@
...
+# V= 0
+# mu= 0
+# Nmax= 3
+```
+
+
+## Properties
+|Property|         Type         |          Description          |
+|--------|----------------------|-------------------------------|
+|hilbert |const AbstractHilbert&| ``Hilbert`` space of operator.|
```
jamesETsmith: @gcarleo is there a way to get the type of properties to be deduced automatically when the documentation is created? If there isn't, should I just write in `netket.hilbert.Hilbert` so it's consistent with the constructor?

gcarleo: @jamesETsmith unfortunately not, for properties you need to spell out the type yourself... this is actually a limitation of pybind11. `netket.hilbert.Hilbert` looks good to me!

gcarleo:
@jamesETsmith can you please add this to .travis.yml 

```
python -m pytest --verbose --doctest-glob='*.md' Docs/
```

this will check that the example codes work as expected!

jamesETsmith:
@gcarleo sure thing!

Docs/GroundState/Vmc.md
```diff
@@ -0,0 +1,48 @@
+# Vmc
+Variational Monte Carlo schemes to learn the ground state using stochastic reconfiguration and gradient descent optimizers.
+## Constructor
+Constructs a ``VariationalMonteCarlo`` object given a hamiltonian, 
+sampler, optimizer, and the number of samples.
+
+|          Field          |   Type   |                                                                         Description                                                                          |
+|-------------------------|----------|--------------------------------------------------------------------------------------------------------------------------------------------------------------|
+|hamiltonian              |          |The hamiltonian of the system.                                                                                                                                |
```
gcarleo: @jamesETsmith great work, you wrote a lot of docs already, thanks! 

Do you have a sense why here we don't get the type for hamiltonian etc?

jamesETsmith: @gcarleo I haven't been able to figure it out yet. I think I jumped the gun on the title of my last commit. When I create the Docs for Vmc, I get 

```bash
/projects/jasm3285/apps/pull_requests/netket/Docs/parse.py:225: UserWarning:  Found argument `sampler` in docstring that does not exist in function signature.
  ' not exist in function signature.' % arg, UserWarning)
/projects/jasm3285/apps/pull_requests/netket/Docs/parse.py:225: UserWarning:  Found argument `hamiltonian` in docstring that does not exist in function signature.
  ' not exist in function signature.' % arg, UserWarning)
/projects/jasm3285/apps/pull_requests/netket/Docs/parse.py:225: UserWarning:  Found argument `optimizer` in docstring that does not exist in function signature.
  ' not exist in function signature.' % arg, UserWarning)
```

I'm not sure why it's not finding them in the function signature, the code looks analagous to the documentation in ``pyhilbert.hpp``.

ooreilly: @jamesETsmith  This problem might be due to some corner case that the parser does not handle correctly. I will try to set a side some time later today to investigate. If it is my fault, sorry for the inconvenience.

jamesETsmith: @ooreilly not a problem! Thanks for the help.

jamesETsmith:
@ooreilly I'm running into an issue where the ``extract.py`` script is trying to extract a docstring for an undocumented function and raising an error. This is only occurring for the classes that extend ``AbstractOperator``, not classes that extend ``AbstractHilbert``, which also contains functions without docs. Do you have any suggestions for this?

Here is part of the error message:
```bash
.
.
.
Docs for  <class 'netket.operator.BoseHubbard'>
  File "make_all_docs.py", line 25, in <module>
    markdown = format.format_class(clsm[1])
  File "/projects/jasm3285/apps/pull_requests/netket/Docs/format.py", line 54, in format_class
    match = clex.extract(method[0])
  File "/projects/jasm3285/apps/pull_requests/netket/Docs/extract.py", line 132, in extract
    return types[self.dtype]()
  File "/projects/jasm3285/apps/pull_requests/netket/Docs/extract.py", line 358, in extract_function
    return self.findall(pattern, ids)
  File "/projects/jasm3285/apps/pull_requests/netket/Docs/extract.py", line 201, in findall
    r'Unable to extract docstring for `%s`' % self.query)
NameError: Unable to extract docstring for `get_conn`
```

ooreilly:
@jamesETsmith Interesting. My first thought is that you want to avoid trying to extract docstrings from functions that have none. But it sounds like it works for some other functions that don't have any docstrings. Not sure what to make of that. I will need to investigate.

A simple work-around for now is to put a try-except statement around the call to `extract` in `format.py`. We can then intercept error and maybe issue a warning instead. I can do that, but maybe we should get @gcarleo opinion on that work-around first.

gcarleo:
@ooreilly @jamesETsmith indeed I believe this is due to the fact that some of the functions don't have docs yet. 

Maybe we can have `extract` return a flag set to false if it doesn't find anything, instead of returning an error? Or maybe we can have another function to check if the docs exist... !

jamesETsmith:
@gcarleo it's strange though that it doesn't raise an error about the functions in ``pyhilbert.hpp`` because those also don't have documentation yet. I've also tried adding documentation for ``get_conn`` and the problem persists. I'm going to spend some time investigating this afternoon and see if I can track down where this behavior is coming from.

ooreilly:
@jamesETsmith if you could do that it would be highly appreciated! I am swamped with other things to do right now. But I will look into this if you need me to. Just let me know if you need me.

My recommendation would be to print out what the docstring looks like that you are sending to extract and compare.

gcarleo:
Hi @jamesETsmith, sorry I was not following too closely the issue here, is there something missing to complete the docs or should I merge this?

jamesETsmith:
@gcarleo there is an error when generating the docs for operators. I'm trying to figure it out now. The property and class docs are done for these, but I'd like to add the documentation for the functions as well before we merge.

Docs/operator/LocalOperator.md
```diff
@@ -95,9 +95,9 @@ that the matrix element O(v,v') is different from zero. In general there
 will be several different connected visible units satisfying this 
 condition, and they are denoted here v'(k), for k=0,1...N_connected.
 
-|Argument|Type|                   Description                    |
-|--------|----|--------------------------------------------------|
-|v       |    |A constant reference to the visible configuration.|
+|Argument|         Type          |                   Description                    |
+|--------|-----------------------|--------------------------------------------------|
+|v       |numpy.ndarray[float64[m|A constant reference to the visible configuration.|
```
gcarleo: maybe some residual glitch with types here?

jamesETsmith: I think it's cutting off the type after a certain number of characters. I will look into this

jamesETsmith:
@ooreilly, no worries, I was happy to look into this! The problem I was running into was because regex was having trouble with functions that returned non-void objects. For functions that returned something other than ``void`` or ``None``, the re module would stop parsing after it encountered the first word of the return type. I modified the regex pattern so it would capture the complete return type and move on to parse the rest of the docstring. I will post a detailed example shortly.

jamesETsmith:
@ooreilly here's the example. For both situations, I print the ``pattern``, ``self.txt``, ``matches`` from ``get_matches()`` in ``extract.py``.

Before modifying the regex pattern:

```bash
^\s*(get_conn)(\([\w\W]*?\)\s*(?:->\s*(\w+)))\n+(\s*)([\w\W]*)?((\4.*\n+)+)?
get_conn(self: netket.Operator, v: numpy.ndarray[float64[m, 1]]) -> Tuple[List[complex], List[List[int]], List[List[float]]]


       Member function finding the connected elements of the Operator. Starting
       from a given visible state v, it finds all other visible states v' such 
       that the matrix element O(v,v') is different from zero. In general there
       will be several different connected visible units satisfying this 
       condition, and they are denoted here v'(k), for k=0,1...N_connected.
    
       Args:
           v: A constant reference to the visible configuration.

       

[]
```

After modification of the regex pattern:
```bash
^\s*(get_conn)(\([\w\W]*?\)\s*(?:->\s*([\w\W]*?)))\n+(\s*)([\w\W]*)?((\4.*\n+)+)?
get_conn(self: netket.Operator, v: numpy.ndarray[float64[m, 1]]) -> Tuple[List[complex], List[List[int]], List[List[float]]]


       Member function finding the connected elements of the Operator. Starting
       from a given visible state v, it finds all other visible states v' such 
       that the matrix element O(v,v') is different from zero. In general there
       will be several different connected visible units satisfying this 
       condition, and they are denoted here v'(k), for k=0,1...N_connected.
    
       Args:
           v: A constant reference to the visible configuration.

[('get_conn', '(self: netket.Operator, v: numpy.ndarray[float64[m, 1]]) -> Tuple[List[complex], List[List[int]], List[List[float]]]', 'Tuple[List[complex], List[List[int]], List[List[float]]]', '       ', "Member function finding the connected elements of the Operator. Starting\n       from a given visible state v, it finds all other visible states v' such \n       that the matrix element O(v,v') is different from zero. In general there\n       will be several different connected visible units satisfying this \n       condition, and they are denoted here v'(k), for k=0,1...N_connected.\n    \n       Args:\n           v: A constant reference to the visible configuration.\n\n       \n", '', '')]
```

The difference here is the third capture group in the expression, which now captures all word characters and alphanumeric characters (e.g. Tuple[...] or netket::Something). Does this make sense?

Docs/operator/LocalOperator.md
jamesETsmith: I think you're right, there are other types that are longer.

jamesETsmith: @gcarleo I fixed this in the commit 1f61edf. During the parsing of the function signature is was splitting the string of the entire signature using `,` as a delimiter. Since some of the the objects (e.g. `numpy.ndarray[float64[m, 1]`) have commas in their type, the type names for some arguments were getting garbled.

@ooreilly I changed the `parse_signature()` function in `parse.py`. Let me know what you think. I've tried to make it a bit more general, so it can handle cases with an unknown number of `:`. I think this will be important for a several of our uses (e.g. the type of a sampler `netket::AbstractSampler<netket::AbstractMachine<std::complex<double> > >`)

ooreilly:
@jamesETsmith Thanks for your fixes. Very much appreciated. I updated things on my end by applying your code changes. Some of my tests broke for function signatures that do not contain any `:` . I fixed this. 
I don't know what is easiest, but if you want to get my changes you can find them here: https://github.com/ooreilly/mydocstring/tree/master/mydocstring

jamesETsmith:
@gcarleo if you're ok with this PR, I think it's ready to merge.

gcarleo:
@ooreilly @jamesETsmith thanks, let's merge this. 

No, @ooreilly , I think nobody is working on the doc_diff tool, I would say we actually maybe don't have a case use for it anymore?

PR #143, opened 2019-01-07
Update Docs

ooreilly:
This PR updates the documentation tools to their most recent version (see PR #141).

PR #144, opened 2019-01-08
Add consistency check for documentation tools

ooreilly:
This tool generates html reports whenever there is a mismatch found between any new and old documentation.

Interestingly enough, there is an extra newline character introduced in the of the docs. I will investigate if it is coming from #143.

### DIFF
--- a/Docs/doc_diff.py
+++ b/Docs/doc_diff.py
@@ -1,5 +1,5 @@
 """
-usage: test_docs.py [DocPath1 DocPath2 ...]
+usage: doc_diff.py [DocPath1 DocPath2 ...]
 
 DocPath1, ...   List the directories to perform consistency check of
                 documentation for (optional).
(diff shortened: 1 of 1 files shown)


gcarleo:
@ooreilly great, this seems useful indeed. 

I see that you have some "build" function inside here. I think it would be extremely useful to have a script to re-generate automatically the documentation for the whole netket library, so that we don't have to do this by hand every time. Basically this would create the folders `Graph`, `Machine` etc corresponding to the different submodules of netket and then build the docs for each of the classes in the modules. 

Wanna help out with this? Thanks!

Docs/test_docs.py
```diff
@@ -0,0 +1,145 @@
+"""
```
gcarleo: maybe we can just call this doc_diff or something? otherwise we might think that test_docs performs other checks such as those performed by pytest on the example snippets

ooreilly:
> @ooreilly great, this seems useful indeed.
> 
> I see that you have some "build" function inside here. I think it would be extremely useful to have a script to re-generate automatically the documentation for the whole netket library, so that we don't have to do this by hand every time. Basically this would create the folders `Graph`, `Machine` etc corresponding to the different submodules of netket and then build the docs for each of the classes in the modules.
> 
> Wanna help out with this? Thanks!

I agree it would be good to automate the build process for the documentation. I can help out with that.

Docs/test_docs.py
ooreilly: That is good suggestion. I renamed it.

PR #146, opened 2019-01-09
Add automatic generation of markdown docs + extract all class methods

gcarleo:
To generate documentation now just go to '/Docs' and do

```
python make_all_docs.py
```

if a new submodule is added to main netket module, it must be manually added in `make_all_docs.py`,
appending its name to the list `submodules=[graph,...]`


gcarleo:
@ooreilly @jamesETsmith I just added a tool to generate all docs for the clsses in the submodules automatically + improved the parser to support generic class methods. @ooreilly , I didn't do the submodule yet

jamesETsmith:
@gcarleo this looks good to me! It's much cleaner than the bash script I was using.

gcarleo:
@ooreilly yes switching to non capitalized directories is intentional, since modules are lowercase, and it makes more sense to keep their non-capitalized names for folders

And true, the empty documentation is because we don't have documentation yet for those classes, we are in the process of adding them. I guess for the moment it's good to keep empty files, also as a placeholder for the work to be done. Maybe one could just generate non empty files with only the class name in the header... I will try to do this change and also moving things in separate functions later

PR #149, opened 2019-01-17
Fix doc-diff

ooreilly:
As explained in #142, doc-diff crashed when trying to run it on certain modules. This PR fixes the problem.
doc diff now uses the same function as `make_all_docs` for temporarily building the documentation.

As stated in #144, this tool generates html reports whenever there is a mismatch found between any new and old documentation. To see it in action, make a temporary change to any of your already generated docs, e.g., `graph/Hypercube.md`.


Docs/make_all_docs.py
```diff
@@ -1,29 +1,3 @@
-import netket
-import format
-import inspect
-import sys
-import os
-import pkgutil
+import build
```
jamesETsmith: If we're going to get rid of nearly everything here, I think we should just call `build_docs()` from the `build.py` and get rid of `make_all_docs.py`.

ooreilly: Thanks for your feedback @jamesETsmith! I agree with everything you said. Latest commit addresses your remarks and here are my responses to them:
 1. I went through the explanations and cleaned them up a bit. I also added some explanations on how to use `build.py`.
2. I removed `make_all_docs.py`. You now use `python3 build.py` to build the docs. If you don't want to build everything simply use `python3 build.py graph`, to only build the `graph` submodule.

gcarleo: I would just suggest to rename build into build_docs, to be a bit more specific

ooreilly: @gcarleo done!

PR #150, opened 2019-01-19
WIP: Extract stochastic reconfiguration code from VMC class

femtobit:
The main commit in this PR extracts the SR code from `VariationalMonteCarlo` to a separate class. 
So far, the interface of the VMC class is left unchanged. However, this could be a good point to discuss whether it should stay this way for v2.0.

Currently, the method to update the variational parameters is configurable in two ways:
1. First, the method to compute the parameter update `dp`. Can be either `Gd` or `Sr` and is set via several VMC constructor parameters. The actual computation steps happen in the `Gradient` and `UpdateParameters` methods of `VariationalMonteCarlo`.
1. Second, the method to apply `dp`, which is determined by the `AbstractOptimizer` passed to VMC. In the simplest case (the `Sgd` class), this is of the form `p(t+1) = p(t) + a(t) * dp` where `a(t)` is the current learning rate.

Part 2 is under control of the users since they can pass in a custom optimizer. Part 1 is currently less configurable. We could discuss whether we should make this extendable as well by providing a suitable base class with subclasses for the current SR and GD methods. What do you think?

(In any case, I think extracting the SR part from the VMC code makes it easier to understand.)

### DIFF
--- a/NetKet/GroundState/variational_montecarlo.hpp
+++ b/NetKet/GroundState/variational_montecarlo.hpp
@@ -151,7 +151,9 @@ class VariationalMonteCarlo {
         sampler_(sampler),
         psi_(sampler.GetMachine()),
         opt_(optimizer),
-        elocvar_(0.) {
+        elocvar_(0.),
+        totalnodes_(MpiSize()),
+        mynode_(MpiRank()) {
     Init(nsamples, discarded_samples, discarded_samples_on_init, method,
          diag_shift, rescale_shift, use_iterative, use_cholesky);
   }
@@ -167,9 +169,6 @@ class VariationalMonteCarlo {
     Okmean_.resize(npar_);
 
     setSrParameters();
-
-    MPI_Comm_size(MPI_COMM_WORLD, &totalnodes_);
-    MPI_Comm_rank(MPI_COMM_WORLD, &mynode_);
 
     nsamples_ = nsamples;
 
--- a/NetKet/GroundState/variational_montecarlo.hpp
+++ b/NetKet/GroundState/variational_montecarlo.hpp
@@ -39,6 +39,137 @@
 
 namespace netket {
 
+namespace detail {
+/**
+ * Computes the average of the matrix or vector `t` of samples over all MPI
+ * nodes.
+ * @tparam T Eigen vector or matrix type
+ * @param t Vector or matrix of samples (should have size n_samples or n_samples
+ * x n_samples.
+ * @param n_samples Number of samples per node.
+ * @param n_nodes Number of MPI nodes.
+ */
+template <class T>
+void AverageSamplesOverNodes(T &t, Index n_samples, Index n_nodes) {
+  SumOnNodes(t);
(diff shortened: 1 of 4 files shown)


NetKet/GroundState/variational_montecarlo.hpp
```diff
@@ -452,13 +512,17 @@ class VariationalMonteCarlo {
     MPI_Barrier(MPI_COMM_WORLD);
   }
 
-  void setSrParameters(double diagshift = 0.01, bool rescale_shift = false,
+  void SetSrParameters(double diag_shift = 0.01, bool rescale_shift = false,
                        bool use_iterative = false, bool use_cholesky = true) {
```
femtobit: Maybe we can change the interface here (and in the VMC constructor) to take a `StochasticReconfiguration::Solver` value directly instead of having the two flags `use_iterative` and `use_cholesky`?

NetKet/GroundState/variational_montecarlo.hpp
```diff
@@ -377,71 +501,7 @@ class VariationalMonteCarlo {
...
-          grad_ /= std::sqrt(nor.real());
-        }
-
-        // if(mynode_==0){
-        //   std::cerr<<it_solver.iterations()<<"
-        //   "<<it_solver.error()<<std::endl;
-        // }
-        MPI_Barrier(MPI_COMM_WORLD);
-      }
+      grad_ = sr_.ComputeUpdate(Ok_, 0.5 * grad_);
```
femtobit: Is it correct that the right-hand side passed to SR here is _half_ the gradient? This is what was used previously, because it was computed as
```cpp
Eigen::VectorXcd b = Ok_.adjoint() * elocs_;
SumOnNodes(b);
b /= double(nsamp * totalnodes_);
```
while the gradient is 
```cpp
 grad_ = 2. * (Ok_.adjoint() * elocs_);
// ...
```

gcarleo: Yes, there is a factor of two, but it's not important since it provides a global rescaling to the learning rate if you want...

femtobit:
@gcarleo Indeed, merging with the optimizer interface is another option that makes sense to discuss.
As you say, this requires a more general interface. Passing the log-derivatives and E_loc should be fine for most of the current methods, though I am not sure whether this might be too specific to the current VMC scheme.

Maybe we should have a way to add additional terms to the loss function, not just expectation values of quantum operators. For example, it might be useful at some point to put a constraint on some norm of the variational parameters to prevent them from growing very large or to favor sparseness. At least for benchmarking in very small systems where the states can still be reasonably stored, it could be useful to minimize the (Fubini-Study) distance from some target state for state preparation.

I am not sure where to best incorporate these kinds of things in the current scheme, but maybe we should think of an architecture that allows this.

gcarleo:
@femtobit can we close this for now?

femtobit:
> @femtobit can we close this for now?

Yes. This discussion should be superseded by the one in #180.

PR #157, opened 2019-01-30
ipython notebook tutorials

kchoo1118:


Tutorials/Notebooks/netket-VMC.ipynb
```diff
@@ -0,0 +1,426 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 28,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Define custom graph\n",
+    "G = nx.Graph()\n",
```
gcarleo: Can we do this without using iGraph?

Tutorials/Notebooks/netket-VMC.ipynb
```diff
@@ -0,0 +1,426 @@
...
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Exact ground state energy = -26.308\n"
+     ]
+    }
+   ],
+   "source": [
+    "res = nk.exact.lanczos_ed(op, first_n=1, compute_eigenvectors=True)\n",
```
gcarleo: It is possible to add observables to ED results basically doing 
```python
# Compute exact neel order
print("g.s. neel order:", res.mean(neel_order, 0))
```

Tutorials/Notebooks/netket-VMC.ipynb
```diff
@@ -0,0 +1,426 @@
...
+    "print(\"Exact ground state energy = {0:.3f}\".format(res.eigenvalues[0]))"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "So we see that the energy we obtained is somewhat in agreement with the value obtained via exact diagonalisation."
```
gcarleo: somewhat in agreement, LOL

Tutorials/Notebooks/netket-VMC.ipynb
```diff
@@ -0,0 +1,426 @@
+{
```
gcarleo: I would maybe move all existing "tutorials" in a folder Examples/, and keep only notebooks with actual tutorials in Tutotorials/ (without the extra Notebooks/ folder) and with a consistent naming (for example this file should maybe called J1J2 or something like that?)

Tutorials/Notebooks/netket-VMC.ipynb
kchoo1118: But this is networkx

Tutorials/Notebooks/netket-VMC.ipynb
kchoo1118: ok i will add that

Tutorials/Notebooks/netket-VMC.ipynb
gcarleo: I meant, just using a NetKet graph and not an external library to do this... I don't know if that's ugly or it can be done nicely with a CustomGraph (I guess it's fine, since this is just a 1D lattice?)

kchoo1118: ok its removed

Tutorials/Notebooks/netket-VMC.ipynb
kchoo1118: Done

kchoo1118:
I think the tutorials should be abit more generic. Like to cover a full workflow rather than specific things like ED or ovservables.

gcarleo:
> I think the tutorials should be abit more generic. Like to cover a full workflow rather than specific things like ED or ovservables.
@kchoo1118 I agree, what should we do then? change name from j1j2 to custom hamiltonian or something? I think it's a bit steep to do the j1j2 as a first tutorial on vmc... :)

kchoo1118:
Ok I agree. Yea lets just keep it as J1J2 then.

PR #158, opened 2019-02-01
Docs for samplers and single python headers 

gcarleo:
Add docs for all samplers and add single headers for python bindings of Sampler classes.


gcarleo:
This is still work in progress, I still need to split other python bindings headers into separate files

jamesETsmith:
@gcarleo, this looks great so far. I think splitting up the documentation is a great idea and will keep the individual files to a much more manageable size.

PR #161, opened 2019-02-09
Add Lattice graph

martamau:
This pull request adds the Lattice graph.
A generic lattice built translating a unit cell. The unit cell can contain  an arbitrary number of atoms, located at arbitrary positions. Periodic boundary conditions can also be imposed along the desired directions.

Here is an example usage to construct a rectangular 3X4 lattice with periodic boundaries

```python
>>> from netket.graph import Lattice
>>> g=Lattice(basis_vectors=[[1,0],[0,1]],extent=[3,4])
>>> print(g.n_sites)
12
```


kchoo1118:
I think is helpful to expose the basisvectors as well.

NetKet/Graph/lattice_impl.hpp
```diff
@@ -0,0 +1,430 @@
...
+  }
+
+  nlatticesites_ = 1;
+  for (int k = 0; k < ndim_; k++) nlatticesites_ *= extent_[k];
+
+  // PATHOLOGIC 2: nlatticesites_=0,1
+  if (nlatticesites_ <= 1) {
+    ErrorMessage()
+        << "A well-defined lattice must have at least 2 sites. Exit.\n";
+    exit(EXIT_FAILURE);
```
gcarleo: The way we are handling errors is through exceptions, please do something like 
```c++
throw InvalidInputError{
          "A well-defined lattice must have at least 2 sites."};
```
then you can remove all the exit statements without problems

NetKet/Graph/lattice_impl.hpp
```diff
@@ -0,0 +1,430 @@
...
+
+};  // namespace netket
+Lattice::~Lattice(){};
+
+// Get private member
+int Lattice::Ndim() const noexcept { return ndim_; };
+int Lattice::Nsites() const noexcept { return nsites_; };
+int Lattice::Size() const noexcept { return Nsites(); };
+
+std::vector<std::vector<double>> Lattice::Coordinates() const noexcept {
```
twesterhout: You allocate memory using `std::allocator<...>`, which is a potentially throwing operation. Marking such functions `noexcept` may lead to crashes later on.

NetKet/Graph/lattice_impl.hpp
```diff
@@ -0,0 +1,430 @@
...
+  std::vector<std::vector<double>> result;
+  for (int k = 0; k < nsites_; k++) result.push_back(R_[k]);
+  return result;
+}
+std::vector<Lattice::Edge> const &Lattice::Edges() const noexcept {
+  return edges_;
+};
+std::vector<std::vector<int>> Lattice::SymmetryTable() const noexcept {
```
twesterhout: Again, you really shouldn't use `noexcept` here.

NetKet/Graph/lattice_impl.hpp
```diff
@@ -0,0 +1,430 @@
...
+// Lattice representations (site = k, vector = n_i, coord = cartesian
+// coordinates)
+std::vector<int> Lattice::Site2Vector(int i) const {
+  int ndim = extent_.size();
+  std::vector<int> result(ndim, 0);
+  int ip;
+  if (i < nlatticesites_)
+    ip = i;
+  else
+    ip = i % nlatticesites_;
```
twesterhout: You don't need an `if` here. `i % nlatticesites_ == i` when `0 <= i < nlatticesites_`.

NetKet/Graph/lattice_impl.hpp
```diff
@@ -0,0 +1,430 @@
...
+            }
+          }
+        });
+  }
+  return is_bipartite;
+}
+
+// Lattice representations (site = k, vector = n_i, coord = cartesian
+// coordinates)
+std::vector<int> Lattice::Site2Vector(int i) const {
```
twesterhout: Could you please add an assert that `i >= 0`?

NetKet/Graph/lattice_impl.hpp
```diff
@@ -0,0 +1,430 @@
...
+};
+std::vector<double> Lattice::Site2Coord(int k) const { return R_[k]; }
+int Lattice::Vector2Site(const std::vector<int> &n) const {
+  int k = 0;
+
+  for (int i = 0; i < Ndim(); i++) {
+    int base = 1;
+    for (int j = i + 1; j < Ndim(); j++) base *= extent_[j];
+    k += n[i] * base;
+  }
```
twesterhout: `base` and `k` can overflow

NetKet/Graph/lattice_impl.hpp
```diff
@@ -0,0 +1,430 @@
...
+
+  return k;
+}
+
+// Neighbours
+std::vector<std::vector<int>> Lattice::PossibleLatticeNeighbours() const {
+  const int nsymbols = 3;
```
twesterhout: You can use `constexpr` for compile-time constants.

NetKet/Graph/lattice_impl.hpp
```diff
@@ -0,0 +1,430 @@
...
+  const int max = nsymbols - 1;
+  std::vector<int> b(ndim_, 0);
+  std::vector<int> result_vector;
+  std::vector<std::vector<int>> result_matrix;
+  std::vector<int> row(ndim_, 0);
+  do {
+    for (std::vector<int>::const_iterator ib = b.begin(); ib != b.end(); ++ib)
```
twesterhout: We're using C++11, so you can do
```cpp
for (auto const& x : b) { result_vector.push_back(x); }
```
or
```cpp
std::copy(std::begin(b), std::end(b), std::back_inserter(result_vector));
```

NetKet/Graph/lattice_impl.hpp
```diff
@@ -0,0 +1,430 @@
...
+      result_vector.push_back(*ib);
+  } while (next_variation(b.begin(), b.end(), max));
+
+  for (std::size_t i = 0; i < result_vector.size() / ndim_; i++) {
```
twesterhout: Any chance you could use Eigen for matrix-related stuff? Your re-implementing slicing and copying makes the code a bit difficult to check (at least for me).

NetKet/Graph/lattice_impl.hpp
```diff
@@ -0,0 +1,430 @@
...
+
+  return result;
+};
+
+bool Lattice::IsBipartite() const noexcept { return is_bipartite_; }
+
+bool Lattice::IsConnected() const noexcept { return is_connected_; }
+
+// Utilities
+bool Lattice::RelativelyEqual(double a, double b,
```
twesterhout: This function should probably be `static`.

NetKet/Graph/lattice_impl.hpp
```diff
@@ -0,0 +1,430 @@
...
+                              double maxRelativeDiff) const {
+  const double difference = std::abs(a - b);
+
+  // Scale to the largest value.
+  a = std::abs(a);
+  b = std::abs(b);
+  const double scaledEpsilon = maxRelativeDiff * std::max(a, b);
+  return difference <= scaledEpsilon;
+}
+double Lattice::GetNorm(const std::vector<double> &coord) const {
```
twesterhout: This functionality is already implemented in Eigen. Why reinvent the wheel?

NetKet/Graph/lattice_impl.hpp
```diff
@@ -0,0 +1,430 @@
...
+  double distance = 0;
+  for (std::size_t i = 0; i < coord.size(); i++)
+    distance += coord[i] * coord[i];
+  return distance;
+};
+double Lattice::GetSquaredDistance(const std::vector<double> &v1,
```
twesterhout: Again, this is a one-liner using Eigen.

NetKet/Graph/lattice_impl.hpp
```diff
@@ -0,0 +1,430 @@
...
+  std::vector<std::vector<int>> adlist(nsites_);
+  for (int k = 0; k < nsites_; k++) {
+    for (std::size_t i = 0; i < edges_.size(); i++) {
+      if (edges_[i][0] == k)
+        adlist[k].push_back(edges_[i][1]);
+      else if (edges_[i][1] == k)
+        adlist[k].push_back(edges_[i][0]);
+    }
+  }
+  return adlist;
```
twesterhout: There is a function for that already in `netket::detail`: `AdjacencyListFromEdges`

NetKet/Graph/lattice_impl.hpp
```diff
@@ -0,0 +1,430 @@
...
+      for (std::size_t i = 0; i < neighbours.size(); i++) {
+        if (k + iatom * nlatticesites_ < neighbours[i])
+          edge_vector.push_back({{k + iatom * nlatticesites_, neighbours[i]}});
+      }
+    }
+  }
+  return edge_vector;
+};
+
+std::vector<std::vector<int>> Lattice::AdjacencyList() const noexcept {
```
twesterhout: You really shouldn't use noexcept here.

NetKet/Utils/array_search.hpp
```diff
@@ -0,0 +1,31 @@
...
+
+namespace netket {
+// Finds the smallest non zero element of an array
+template <class ForwardIterator>
+ForwardIterator min_nonzero_elem(ForwardIterator first, ForwardIterator last) {
+  if (first == last) return last;
+  ForwardIterator smallest = first;
+
+  while (++first != last)
+    if (*first < *smallest and *first != 0) smallest = first;
```
twesterhout: Using `and` and `or` in place of `&&` and `||` is considered bad practice, doesn't work on MSVC, and was removed in `C++17`. Could you, please, replace `and` with `&&`?

NetKet/Graph/lattice_impl.hpp
```diff
@@ -0,0 +1,430 @@
...
+  std::vector<std::vector<int>> neighbours_matrix_in;
+  std::vector<std::vector<int>> result;
+  std::vector<double> distance;
+  double min_distance;
+
+  neighbours_matrix_in = PossibleLatticeNeighbours();
+
+  distance = NeighboursSquaredDistance(neighbours_matrix_in, iatom);
+
+  min_distance = *min_nonzero_elem(distance.begin(), distance.end());
```
twesterhout: This looks like undefined behaviour. Your implementation of `min_nonzero_elem` can return `distance.end()`. If you're absolutely sure it can't happen, consider adding an assert either here or in `min_nonzero_elem`.

NetKet/Graph/lattice_impl.hpp
martamau: Done!

NetKet/Graph/lattice_impl.hpp
```diff
@@ -0,0 +1,430 @@
...
+      basis_vectors_(std::move(basis_vector)),
+      atoms_coord_(std::move(V_atoms)) {
+  ndim_ = basis_vectors_.size();
+
+  // check if dimensions are consistents
+  for (int i = 0; i < ndim_; i++) {
+    if (basis_vectors_[i].size() != ndim_) {
+      ErrorMessage() << "Each element of basis_vectors must have ndim"
+                        "components. Dimensions are inconsistent. Exit.\n";
+      exit(EXIT_FAILURE);
```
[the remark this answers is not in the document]
martamau: Done!

NetKet/Graph/lattice_impl.hpp
```diff
@@ -0,0 +1,430 @@
...
+    exit(EXIT_FAILURE);
+  }
+
+  // PATHOLOGIC 1: pbc(i) for extent(i)<=2 + extent(i)=0)
+  for (int i = 0; i < ndim_; i++) {
+    if (extent_[i] <= 2 and pbc_[i] == true) {
+      ErrorMessage() << "PBC are not allowed when extent<2. Exit.\n";
+      exit(EXIT_FAILURE);
+    }
+    if (extent_[i] == 0) {
```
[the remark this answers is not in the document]
martamau: Thanks, I've added this check too

NetKet/Graph/lattice_impl.hpp
martamau: fixed

NetKet/Graph/lattice_impl.hpp
martamau: Thanks :)

NetKet/Graph/lattice_impl.hpp
martamau: Done

NetKet/Graph/lattice_impl.hpp
martamau: Changed!

NetKet/Graph/lattice_impl.hpp
martamau: I put the function in `Utils/math_helpers.hpp` as @femtobit suggested so it should be ok without the static keyword

NetKet/Graph/lattice.hpp
```diff
@@ -0,0 +1,87 @@
...
+  int Vector2Site(const std::vector<int> &n) const;
+  // Neighbours
+  std::vector<std::vector<int>> PossibleLatticeNeighbours() const;
+  std::vector<double> NeighboursSquaredDistance(
+      const std::vector<std::vector<int>> &neighbours_matrix, int iatom) const;
+  std::vector<std::vector<int>> LatticeNeighbours(int iatom) const;
+
+  std::vector<int> FindNeighbours(int k, int iatom) const;
+  // FindNeighbours returns the index of the neighbours of site k (supposing
+  // that the site 0 is at the origin of the coordinates)
```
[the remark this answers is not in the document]
martamau: Ok, I've changed it. Thanks to your suggestion I also added other comments, hopefully useful!

NetKet/Graph/lattice_impl.hpp
```diff
@@ -0,0 +1,430 @@
...
+      int new_index = n[j] + lattice_neighbours_vector[i][j];
+      if (pbc_[j]) {
+        if (new_index < 0)
+          new_index = extent_[j] - 1;
+        else if (new_index >= extent_[j])
+          new_index = 0;
+        single_neighbour_vector.push_back(new_index);
+
+      } else if (new_index >= 0 and new_index < extent_[j])
+        single_neighbour_vector.push_back(new_index);
```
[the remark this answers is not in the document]
martamau: done!

NetKet/Graph/lattice.hpp
```diff
@@ -0,0 +1,87 @@
...
+  std::vector<std::vector<int>> BuildSymmetryTable() const;
+  // BuildSymmetryTable build the symmetrytable = st[i][j] contiene la
+  // i-esima permutazione equivalente dei siti
+
+  double GetNorm(const std::vector<double> &coord) const;
+  // GetNorm returns the norm (squared distance from 0) of its argument
+  double GetSquaredDistance(const std::vector<double> &v1,
+                            const std::vector<double> &v2) const;
+  // GetSquaredDistance returns the distance between its argument
+  bool RelativelyEqual(double a, double b,
```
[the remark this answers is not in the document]
martamau: Done!

NetKet/Graph/lattice_impl.hpp
```diff
@@ -0,0 +1,430 @@
...
+      exit(EXIT_FAILURE);
+    }
+    if (extent_[i] == 0) {
+      ErrorMessage() << "Extent components must be >0. Exit.\n";
+      exit(EXIT_FAILURE);
+    }
+  }
+
+  nlatticesites_ = 1;
+  for (int k = 0; k < ndim_; k++) nlatticesites_ *= extent_[k];
```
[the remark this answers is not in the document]
gcarleo: This would overflow if you have 2^32 or 2^64 (more optimistic) lattice sites... not very realistic anyways

NetKet/Graph/lattice_impl.hpp
```diff
@@ -0,0 +1,430 @@
...
+#include <iostream>
+#include <vector>
+#include "Utils/array_search.hpp"
+#include "Utils/messages.hpp"
+#include "Utils/next_variation.hpp"
+
+namespace netket {
+// Constructor
+Lattice::Lattice(std::vector<std::vector<double>> basis_vector,
+                 std::vector<int> extent, std::vector<bool> pbc,
```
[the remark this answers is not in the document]
gcarleo: what would you use instead?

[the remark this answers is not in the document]
twesterhout: Well, I'd probably use something like `std::vector<uint8_t>`. `uint8_t` is implicitly convertible to `bool` so it's not like the implementation has to change. `std::vector<bool>` is weird if you know what I mean :)

NetKet/Graph/lattice_impl.hpp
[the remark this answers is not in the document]
twesterhout: @gcarleo That's not the issue. Of course, we should try to support graphs with more than 2^32 nodes. The point here is the safety of the code. This function is part of the public interface (i.e. it is exposed to Python) so we better make sure *all* kinds of crazy inputs are handled properly. For example, if I try to create a lattice with `extent = vector{2^31-432, 102, 475, 205}`, current version of the code will say that `nlatticesites_ = 4235296` (invoking undefined behaviour along the way) and proceed with construction. That's obviously not the intended behaviour.
This constructor won't be on the hot path anyway, why can't we spare a few nanoseconds to ensure overflows are handled properly?

NetKet/Graph/abstract_graph.hpp
```diff
@@ -52,9 +52,10 @@ class AbstractGraph {
   virtual int Size() const noexcept = 0;
 
   /**
-  Returns the graph edges.
+  Member function returning the graph edges.
+  @return Graph edges.
   */
-  virtual std::vector<Edge> const &Edges() const noexcept = 0;
+  virtual std::vector<Edge> const &Edges() const = 0;
```
twesterhout: Why is that? Could you revert it, please?

twesterhout:
Thanks for the update @martamau! Do you plan to replace hand-written loops with functions from Eigen?

NetKet/Graph/abstract_graph.hpp
martamau: I've cut too many noexcept...thanks

martamau:
Hi everyone! 
@femtobi Right, the current code deals only with nearest neighbor edges (I'll add this specification to the documentation). There's indeed the idea to extend the class to neighbors of higher order, I'll probably need it in the next months!

@twesterhout sure it would be great to use Eigen instead of rewriting already implemented code, but right now there's not enough time to replace the functions I've written, since this operation should be handled with care...anyway I'll try to see if I can do it quickly in the next days.

gcarleo:
Thanks all for the comments and suggestions for @martamau ! 
I guess we are already in a good shape to merge this, given the time constraints and other practical considerations (I would like to release a 2.0b1 no later than Friday). 
I agree that parts of the code should be rewritten using Eigen, but this does not affect the interface and functionality, so we can postpone this to a future PR. 

@martamau there is just the issue that checks online fail because igraph is not installed on Travis. You'd need either switch to tests that rely on networkx or add igraph to the Travis configuration script.

@twesterhout are there other urgent matters or can we unlock this?

twesterhout:
> @twesterhout are there other urgent matters or can we unlock this?

Hoping for an implementation in terms of Eigen, I haven't checked most of the "working with a bunch of indices"-type of loops. So _assuming you did_ ;) I'm approving this PR.

NetKet/Graph/lattice_impl.hpp
martamau: Solved

Issue #156, opened 2019-01-30
Replace old tutorials with Jupyter notebooks


gcarleo:
The tutorials [on the website ](https://www.netket.org/tutorials/home/) should be replace by Jupyter notebooks, and should be part of the main NetKet repository. 
This has several advantages: 
1. ipython notebooks are easily converted to markdown, thus can be immediately put on the website 
2. both the documentation and the tutorials will leave on the main repo, thus being updated more regularly with PRs and also being tested more efficiently 

@kchoo1118 has a very nice example of how the tutorials should be, please do a PR with it and let's help him finish the other tutorials!

jamesETsmith:
The build of NetKet might be a bit complicated for use with [Binder](https://mybinder.org/), but I think it's worth mentioneing if we're going to have tutorials as Jupyter notebooks. It lets you run jupyter notebooks on virtual machines so users could run a few simple tutorials using only a browser. For more information about Binder you can check out their docs [here](https://mybinder.readthedocs.io/en/latest/). Do you think this would be worthwhile?

fabienalet:
Hi 
sorry I'm asking a very very basic question. I'd like perhaps to contribute to tutorials while learning Netket v2.0 on the way. 
But I'm currently unsure on how should I install / compile / use the python bindings proposed in the v2.0. I tried to do cmake, make and make install, but that didn't seem to install the python library netket, only a shared object (thus any python code gets stucks when importing netket).
Probably I'm missing something very basic (but I guess the basic user of netket could as well).

gcarleo:
Hi @fabienalet !

There is some installation instructions here https://github.com/netket/netket/pull/82 , basically you don't need to run make yourself, but do 
```
pip install . 
```
in the main repo. This will automatically register the library in python, and you can start using it as shown in the folder Examples/. Only requirement here is the c++11 mpi compiler. 

For the stable version, one will not need to download the repository, but just go on the command line and do 
```
pip install netket
```
this uses che centralized python distribution systems and automatically downloads/compiles everything. If you do that now you will get some earlier version that I registered on the server maybe two weeks ago. I would advise at this stage to always follow the first strategy, to use the latest version. 

It would be great if you could help us write tutorials! We are actually very close to releasing a beta for 2.0 and update the website, but we are still missing several tutorials...

gcarleo:
@everthemore @ShHsLin , do you like python notebooks? 
It would be great to have a short tutorial on supervised learning, similar to what you find in `Tutorials/` !

[...]

PR #166, opened 2019-02-18
Setting up Binder to run Tutorials

jamesETsmith:
This PR sets up [binder](https://mybinder.readthedocs.io/en/latest/index.html), which would allow users to run small tutorials on virtual machines hosted by binder. I mentioned it in the issue #156.

Binder uses the `environment.yml` file to set up a `Conda` environment with all of the dependencies (mpi4py, an mpicompiler, numpy, etc). I've added a badge in the `README.md` that links to the binder build of NetKet. I've tested `CustomHamiltonian.ipynb`, which runs without issue, and 'j1j2.ipynb`, which crashes when the user creates the FFNN. I'm not totally sure why this is yet, but I'll continue to look into it. 

If you want to check out the binder you can use the badge in the `README.md` or this link: https://mybinder.org/v2/gh/jamesETsmith/netket/mybinder


gcarleo:
@jamesETsmith thanks! This indeed is a good idea, especially to facilitate the online workflow. I was just experiencing some issue running the second Tutorial, somehow the code was freezing when defining the `Machine`, I am not sure this is related to me or if it's an issue you see too?

jamesETsmith:
@gcarleo that sounds like a similar problem to the one I was experiencing. You can also run terminals on binder, which actually show error messages (unlike the notebooks), and from a few quick tests on these, I think this an mpi4py/MPI problem. I will have a look at this more over the next few days!

gcarleo:
@jamesETsmith can you try to remove mpi4py at once ? Basically we don't need it anymore, and maybe it is causing some issue... anyways if you remember, we were having issues with conda installs on Mac. We should try again now that we don't use mpi4py

jamesETsmith:
Apologies for the delay on this, I will try to get this up and running this week so we can link to a binder notebook in the paper.

jamesETsmith:
@gcarleo, it looks like things were already working after my last commit and I remembered incorrectly. Do you want me to close this PR and open a new one on the main branch?

gcarleo:
Yes please!

PR #175, opened 2019-03-14
Adding Files for Binder

jamesETsmith:
This a new PR, which adds the files necessary to build a Docker image with NetKet on the Binder servers. This allows users to run jupyter notebooks __and__ terminals on these docker images so they tinker with the tutorials and other small jobs without having to install NetKet locally.


gcarleo:
Thanks James, looks like there is still some glitch with cmake?

jamesETsmith:
I just noticed something wasn't quite right, I'm not sure why it's showing up in this PR and not the other. I'll try to get this figured out today.

jamesETsmith:
@gcarleo the tutorials all run on binder. The J1J2 model tutorial is a bit big for binder so it's pretty slow, but users can decrease the size if they don't want to wait for it to finish.

gcarleo:
Great James, this works nicely now! Do you think it would be possible to move this from your repository to the main one, or is that going to be automatic once we merge? 
I mean, I can see now something like " loading jamesETsmith/netket/mybinder", but maybe that gets replaced when on the main repo

jamesETsmith:
Good catch, part of this will be done automatically as long as we point people to the correct url. To do this, I'll update the badge in `README.md` to link to the `netket/netket/master` branch of NetKet instead of my forked branch.

PR #179, opened 2019-04-09
Remove some templates

gcarleo:
This PR removes some templates (in `Machine`) that are not used anymore (they were used in an archaical version of NetKet), and also are not necessary to handle machine with real parameters only (see issue #178 ) 

A future PR will handle issue #178 separately, also connecting to #150 .

### DIFF
--- a/NetKet/GroundState/matrix_replacement.hpp
+++ b/NetKet/GroundState/matrix_replacement.hpp
@@ -36,8 +36,7 @@ namespace internal {
 // MatrixReplacement looks-like a SparseMatrix, so let's inherits its traits:
 template <>
 struct traits<netket::MatrixReplacement>
-    : public Eigen::internal::traits<
-          Eigen::SparseMatrix<std::complex<double>>> {};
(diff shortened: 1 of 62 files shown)


femtobit:
I am fine with removing the template parameters as well, we have only ever used the same instances anyway.

> A future PR will handle issue #178 separately, also connecting to #150 .

Are you working on this at the moment? I was also thinking of improving on #150 but did not get to working on it yet. If you have a specific design in mind, I'd be interested in learning more about that.

gcarleo:
> Are you working on this at the moment? I was also thinking of improving on #150 but did not get to working on it yet. If you have a specific design in mind, I'd be interested in learning more about that.

Yes, I hope to find some time to do that... The idea is to define 1) a general class that does expectation minimization, to be used in all other applications as special cases (variational, supervised, unsupervised etc) and more. 2) separate the SR interface as you were already doing, more or less 3) allow for non-analytic machines, thus generalizing the SR updates to the case of purely real parameters (one for the real and one for the imaginary part, as treated independently)

Issue #176, opened 2019-04-02
Machines with real parameters


gcarleo:
We need to remove some unused templates in `AbstractMachine` and allow to have purely real-valued parameters as well as non analytic wave-functions

PR #180, opened 2019-04-15
Real parameters

gcarleo:
This PR addresses #176 , providing support for non-holomorphic wave-functions (and, as a special case, wave-functions with only real-valued parameters). 

I also separated the SR module, taking inspiration from #150  and fixed an issue with Unsupervised example. 

These changes will ease the developments associated with interfacing to pytorch in the near future. 

The way I currently handle the holo/non-holo cases is fine and relatively compact, but maybe not too elegant, please suggest alternative solutions if you want!


twesterhout:
@gcarleo I like the initiative, thanks! We've been using a very similar approach since last September, because we had everything implemented using PyTorch which didn't play well with complex numbers.

About the realisation... Can we, please, split this into (at least) two PRs: one dealing with SR and the second dealing with RBMs?

gcarleo:
Thanks @twesterhout! 
The thing is that RBM changes are really about introducing two new wave-functions, they don't touch the existing states. It's somehow necessary to have at least one example of machine with purely real variational parameters...

twesterhout:
> The thing is that RBM changes are really about introducing two new wave-functions, they don't touch the existing states. It's somehow necessary to have at least one example of machine with purely real variational parameters...

@gcarleo Ah, you mean for testing. I see. That makes sense.

Okay, but I'll still split the comments into two parts. Let's start with the machines since they are simpler. Any chance we can get rid of all the copy-pasting? I mean the algorithms are really the same, and they're now implemented three (or more?) times in different places.

twesterhout:
P.S. `constexpr` is not the reason for compile failures. It was just used incorrectly -- the "definition" was missing.

gcarleo:
> Okay, but I'll still split the comments into two parts. Let's start with the machines since they are simpler. Any chance we can get rid of all the copy-pasting? I mean the algorithms are really the same, and they're now implemented three (or more?) times in different places.

I know, this is ugly... Ideally, the RbmSpinReal should be just a templated version of RbmSpin, and RbmSpinPhase should be just the sum of two RbmSpinReal (in log space...). This should be done, I agree. I hear that in summer there will be many volunteers to do that :)

twesterhout:
> I know, this is ugly... Ideally, the RbmSpinReal should be just a templated version of RbmSpin, and RbmSpinPhase should be just the sum of two RbmSpinReal (in log space...). This should be done, I agree. I hear that in summer there will be many volunteers to do that :)

Ah, I didn't you you were planning to wait till summer to merge this :)

`RbmSpinPhase` -- why is it called this way? What's so special about the phase here?

gcarleo:
> RbmSpinPhase -- why is it called this way? What's so special about the phase here?

It's the wave-function we used in [here](https://www.nature.com/articles/s41567-018-0048-5) . Basically you parameterize logpsi(X) = log(RBM_1(x)) + i*log(RBM_2(x)) where the two RBMs have real parameters, as opposed to the "standard" psi(X)=RBM(X) where RBM here has complex parameters

Actually all these cases are just special cases of feed-forward networks with specific activation functions. In the grand scheme of things, we really only need one machine, which is a feed-forward network, and leave the others as architecture implementations (that can be done at the python level with a collection/library of architectures, i.e. just specifying how many layers and what activations are used).

gcarleo:
That's also why it's very important to have an efficient and general implementation of `FFNN` (say, based on the Pytorch backend, if available), so that we can safely remove all `RBM` and alikes

twesterhout:
> It's the wave-function we used in here . Basically you parameterize logpsi(X) = log(RBM_1(x)) + i*log(RBM_2(x)) where the two RBMs have real parameters, as opposed to the "standard" psi(X)=RBM(X) where RBM here has complex parameters
>
> Actually all these cases are just special cases of feed-forward networks with specific activation functions. In the grand scheme of things, we really only need one machine, which is a feed-forward network, and leave the others as architecture implementations (that can be done at the python level with a collection/library of architectures, i.e. just specifying how many layers and what activations are used).

Sorry, I think you might've misunderstood me. I know all that and like I said we've been using this approach for a long time now. The only thing bothering me here is the class name.

gcarleo:
> Sorry, I think you might've misunderstood me. I know all that and like I said we've been using this approach for a long time now. The only thing bothering me here is the class name.

It's because you model the **phase** of the wave-function separately from the amplitude. It's just a name...

twesterhout:
> It's because you model the phase of the wave-function separately from the amplitude. It's just a name...

So not **amplitude** separately from phase? Jokes aside, is this PR ready for review? In that case give me a day or two to look through the code a little bit more carefully.

gcarleo:
Sure, go ahead...

NetKet/GroundState/variational_montecarlo.hpp
```diff
@@ -64,25 +63,18 @@ class VariationalMonteCarlo {
...
-  double sr_diag_shift_;
-  bool sr_rescale_shift_;
-  bool use_iterative_;
 
   int totalnodes_;
   int mynode_;
 
   AbstractOptimizer &opt_;
+  SR sr_;
+  bool dosr_;
```
femtobit: I have been thinking about whether we can get rid of the `dosr_` flag (which at the same time would make the code more modular).

One option would be to make a small class hierarchy like this (just pseudo-code):
```
class Updater {
public:
  void ComputeUpdate(...) = 0;
};

class GradientBasedUpdater : public Updater {
  Optimizer opt_;
public:
  override void ComputeUpdate(...) {
    // Compute deltap using opt.Update
  }
};

 // The current SR class with only the code for the holomorphic update
class SRHolomorphic  : public Updater {
public:
  override void ComputeUpdate(...) {
    // ...
  }
};

 // The current SR class with only the code for the non-holomorphic update
class SRReal  : public Updater {
public:
  override void ComputeUpdate(...) {
    // ...
  }
};
```

This has the additional benefit that a) we could more easily incorporate strategies that are neither SR nor `Optimizer`-like and b) it more cleanly separates the holomorphic and non-holomorphic SR code, which should make the code easier to understand.

gcarleo: Good idea @femtobit , the remaining issue is really defining an interface that makes sense/is general enough for `ComputeUpdate` in `Updater` then. I agree that this structure would eliminate a lot of ugly and redundant code though....

gcarleo:
Just an update: since this PR is rather "urgent", I will leave the refinements about how to integrate SR and general updates possibly for future work (also because I believe we need more thoughts on the design of the "super updater" class). At the moment I will try though to eliminate some of the ugly things, most notably the fact that for analytic machines the parameters updates need to be done explicitely considering twice as large real-valued updates

gcarleo:
The solution [here](https://github.com/netket/netket/pull/180/commits/e25927a3d15093de387a7a5c15ead04d0f225512) removes the ugly updates in VMC etc. It also greatly simplifies the optimizer interface, removing a lot of duplicate code. I think this solution is good enough for the moment...

NetKet/Optimizer/stochastic_reconfiguration.hpp
```diff
@@ -0,0 +1,162 @@
...
+        S.setShift(sr_diag_shift_);
+        S.setScale(1. / nsamp);
+
+        it_solver.compute(S);
+        deltaP = it_solver.solve(grad);
+        MPI_Barrier(MPI_COMM_WORLD);
+      }
+    } else {
+      if (!use_iterative_) {
+        // Explicit construction of the S matrix
```
twesterhout: Not sure if it's actually wrong, but it does look weird to me... So if I'm reading this correctly, what you have here is `Sδw = F` where `S` and `F` are complex and `δw` should be real. Then, you decide to solve `Re[S]δw = Re[F]` instead of the initial system. I'm wondering why? I mean, obviously, it's very unlikely that `Sδw = F` will have a purely real solution. But then I'd expect one to resort to solving a LLS (Linear Least Square) problem: `min‖Sδw - F‖`.

gcarleo: This is a bit subtle indeed. The system Re[S]dw=Re[F] is the correct way of doing the SR update when the function is not holomorphic. When the function is holomorphic instead you solve S*dw=F , and consider both real and imaginary part of dw... I have some notes on that, I will maybe attach them to the repository as technical notes

gcarleo: To some extent this is explained also in this [book](https://www.amazon.com/Quantum-Monte-Approaches-Correlated-Systems/dp/1107129931), around page 139. But it's a bit obscure from that derivation why those formulas hold

twesterhout: Please, do! Some time ago my numerical experiments showed the opposite so that's why I was asking...

PR #184, opened 2019-05-08
Improve consistency of the matrix wrapper interface in Python

femtobit:
Continuing the discussion in the comments here:
https://github.com/netket/netket/commit/32af4c6db6648383790703b76d251a7416e549bc#r33439404

This PR is my suggestion for exposing the matrix wrapper interface to Python (without any of the more radical ideas regarding the `AbstractOperator` class hierarchy).

Now, in order to get a SciPy sparse matrix from the operator, one needs to write:
https://github.com/netket/netket/blob/bc8e0f3bcb8c456b78cea49260e3fffe08cf520e/Examples/ExactDiag/ising1d.py#L28
(The `type='sparse'` is optional, since `sparse` is the default in my suggestion.)


gcarleo:
OK thanks @femtobit, this indeed is a good strategy. 
I'd just suggest to add a syntactic sugar
such that ```as_matrix('matrix_type')``` is equivalent to ```to_matrix('matrix_type').data``` 

This also means that ```as_matrix``` will work only if ```matrix_type``` is sparse or full (as it should) and that data can stay private (as you did before). Does it make sense?

femtobit:
> This also means that as_matrix will work only if matrix_type is sparse or full (as it should) and that data can stay private (as you did before). Does it make sense?

I would find it a bit confusing to have both `as_matrix` and `to_matrix` (that's extremely easy to mix up).

How about this:
```python
op.to_matrix() # NetKet matrix wrapper
op.to_bare_matrix() # NumPy/SciPy array
# or, alternatively:
op.to_raw_matrix() # NumPy/SciPy array
```

jamesETsmith:
Pytorch tensors have a method called `numpy()` that converts from the Pytorch tensor class to a numpy array (see [here](https://pytorch.org/tutorials/beginner/former_torchies/tensor_tutorial.html#numpy-bridge)). Using `op.numpy()` would keep the function calls unambiguous and provide some continuity for uses of both packages. Just an idea!

gcarleo:
Thanks @jamesETsmith it's good to look at what Pytorch does too... !

Let me try to make my point of view more explicit: to me, we have three logically distinct operations, and as such we should have as many different methods: 

```python
op.as_linear_operator() 
op.as_full_matrix()
op.as_sparse_matrix()
```
and the more I think about it, the more I am convincing myself that the first operation should just return a `scipy.sparse.linalg.LinearOperator` , since this will allow to automatically access many other algorithms (like matrix exponentiation, full Arpack support etc) that we don't want to reimplement from scratch. So in this sense, the internal netket matrix wrapper is never exposed in python, but only scipy/numpy converters are. Then we can further argue if we can combine `as_full_matrix()` and `as_sparse_matrix` into a single method with options (I think that `numpy()` is not too appropriate though, because the sparse matrix is not a numpy object but a scipy one...!), but that's largely a matter of taste?

femtobit:
@gcarleo One problem I see with `to_sparse_matrix` and `to_dense_matrix` is that if we then want to provide both the matrix wrappers and NumPy/SciPy classes we already need four methods (five, if we add another one for the direct wrapper).

I'd still go with `as_matrix()` and `as_bare_matrix()` with an argument for the type. But as you say, it is mostly a matter of taste.

twesterhout:
Can we not employ a little bit of Python magic and make all operators be `LinearOperator`s by default (using inheritance) when the corresponding Hilbert spaces are "indexable"? Then we're only left with two methods. I would advise against merging them into one, because that makes passing arguments specific to dense/sparse matrices tricky (think column vs row compression). A suggestion for names: `todense()` (`scipy` uses it I believe) and `tosparse()`.

gcarleo:
@femtobit indeed, but I am not sure we need to expose the internal wrapper... we only need to expose converters to numpy/scipy types, thus at most three methods in total. 

`to_matrix()` and `to_bare_matrix()` are not too informative though, at least it's not clear to me that bare is a synonym of full matrix and and matrix alone should be a sparse one! `to_dense` and `to_sparse` are also good, I agree @twesterhout 

@twesterhout the problem is that converting a `LinearOperator` to a `SparseMatrix` or `DenseMatrix` cannot be done automatically, or at least I believe that scipy does not implement that. The conversion from matrices to linear operators however is automatic, yes

femtobit:
> to_matrix() and to_bare_matrix() are not too informative though, at least it's not clear to me that bare is a synonym of full matrix and and matrix alone should be a sparse one! 

The "bare" matrix would be the NumPy/SciPy one, the "to_matrix" would be the wrapper.

> indeed, but I am not sure we need to expose the internal wrapper...

Okay, yes, we could indeed decide to not expose the wrappers in the Python interface. This comes at the cost of having all methods that need to actually use a wrapper take an optional argument for the type (like `lanczos_ed` right now).

In that case, `to_dense` and `to_sparse` sounds good to me.

gcarleo:
> The "bare" matrix would be the NumPy/SciPy one, the "to_matrix" would be the wrapper.

Sorry I find it confusing...  :) 

> Okay, yes, we could indeed decide to not expose the wrappers in the Python interface. This comes at the cost of having all methods that need to actually use a wrapper take an optional argument for the type (like lanczos_ed right now).

But do we need to pass the wrapper to Lanczos? For example, the constructor takes only an Operator, not a wrapper 
```python
res = nk.exact.lanczos_ed(ha, first_n=3, compute_eigenvectors=True)
```
I have to double check, but I think we already specify in a flag wether it should be treated as sparse etc? Also, the lanczos_ed is nice to have as an internal, but it is not needed anymore, since we can just use scipy with a similar computational cost

```python
vals, vecs = eigs(ha.to_sparse(), k=3, which='SR')
```

> In that case, to_dense and to_sparse sound good to me.

Fantastic! I just don't know the detail of how to make our method return something that is a `scipy.sparse.linalg.LinearOperator`... this is a bit tricky?

femtobit:
> Sorry I find it confusing... :)

Okay, I understand. The fact that each of us finds different conventions confusing is probably one of the best arguments for removing the wrappers from the interface, so that users do not have to think about the distinction at all.

> But do we need to pass the wrapper to Lanczos? For example, the constructor takes only an Operator, not a wrapper

Indeed, that was my point. We can pass the abstract operator and a flag for the wrapper type if needed, just like `lanczos_ed`. The only part of the code that takes a wrapper directly is the imaginary time propagation driver. I can change that soon.

femtobit:
> Fantastic! I just don't know the detail of how to make our method return something that is a scipy.sparse.linalg.LinearOperator... this is a bit tricky?

According to the SciPy documentation, one can subclass `LinearOperator` and implement `_matvec` (which does essentially the same thing as `Apply` of the matrix wrappers) and a few other simple methods.

Maybe something like this Python class would already (I did not get to try it yet):
```python
class NetketLinearOperator:
    def __init__(self, operator):
        self.op = operator
        self.wrapper = nk.operator.DirectMatrixWrapper(operator)

    def _matvec(vec):
        return self.wrapper.apply(vec)

    # some other required methods...
```

gcarleo:
> We could then add a method as_linear_op() and get it to return this Python class. Since pybind11 also allows calling Python function from C++, this should be doable.

OK great, this would be nice. Otherwise we need to start thinking about adding functions to the netket module on the python side.. 

> [Side note: I would like to call the methods to_sparse/to_dense but as_linear_op. This is because they behave differently: The first two really copy and convert something, hence the to_, while the as_ method just gives you a different view on an existing object, without copying it (just retaining a reference).]

OK agreed

twesterhout:
> We could then add a method as_linear_op() and get it to return this Python class. Since pybind11 also allows calling Python function from C++, this should be doable.

Perhaps, I wasn't clear. What I meant is that `as_linear_op()` should be a no-op, i.e. it should just return `self`. We just make `AbstractOperator` derive from `scipy.sparse.LinearOperator`. And after you have that it becomes obvious that you don't even need `as_linear_op` method, since your operator is already a `LinearOperator`.

> the problem is that converting a LinearOperator to a SparseMatrix or DenseMatrix cannot be done automatically, or at least I believe that scipy does not implement that. The conversion from matrices to linear operators however is automatic, yes

Of course, converting `LinearOperator` to a matrix shouldn't be done automatically. That would defeat the purpose of having a `LinearOperator` concept. I fail to see how this is related to the solution which I'm suggesting. @gcarleo could you elaborate?

gcarleo:
> Perhaps, I wasn't clear. What I meant is that as_linear_op() should be a no-op, i.e. it should just return self. We just make AbstractOperator derive from scipy.sparse.LinearOperator. And after you have that it becomes obvious that you don't even need as_linear_op method, since your operator is already a LinearOperator.

OK, that's possibly the best solution, but I don't know how to do within pybind. I am sure it's possible though 

> Of course, converting LinearOperator to a matrix shouldn't be done automatically. That would defeat the purpose of having a LinearOperator concept. I fail to see how this is related to the solution which I'm suggesting. @gcarleo could you elaborate?

I meant that the conversion from LinearOperator to numpy matrix or scipy sparse matrix should be done via two other methods, but it's actually what you mean too, we agree.

twesterhout:
> OK, that's possibly the best solution, but I don't know how to do it within pybind. I am sure it's possible though

I don't know many examples of this, but here's one from PyTorch: https://github.com/pytorch/pytorch/blob/v1.1.0/torch/csrc/api/include/torch/python.h#L47.

For people, who are not too familiar with PyTorch, here's a bit of background. PyTorch's Python frontend uses `torch.nn.Module` as a base class for all the neural networks. C++'s frontend has `torch::nn::Module` which is similar in functionality, but bears no relation with Python. Sometimes you create a model in C++, but would like to use it from Python. So you'd like your model to subclass `torch.nn.Module`. The code snippet I linked achieves that (it's actually a little bit more complicated in their use case, but never mind :)).

femtobit:
With the new commits, this PR implements the `to_dense`/`to_sparse` methods as discussed. Also, no matrix wrappers need to be created from user code any more (`ImaginaryTimePropagation` now also takes the Hamiltonian plus an optional parameter for the matrix type).

Regarding `as_linear_operator`: The difference between `AbstractOperator` and `DirectMatrixWrapper` is that the wrapper creates and stores a `HilbertIndex` object, which is currently required for `Apply` etc.
We could think about moving the functionality from the `HilbertIndex` directly to `AbstractHilbert` (with the corresponding methods having to return some error or throw an exception, if the Hilbert space is too large).
Then, the data contained in `HilbertIndex` only has to be created once and can be used to implement `Apply` etc. for `AbstractOperator`.

Otherwise, we would still need the `DirectMatrixWrapper` to represent the operator plus `HilbertIndex`.

In any case, we could do this in a separate PR. In my opinion, this one can be merged, so that we have the `to_dense`/`to_sparse` interface ready.

femtobit:
By the way: While the following code is not perfect with regard to the interface and involves ore than one layer of wrapper classes, it does make it possible to use NetKet operators as SciPy linear operators (the same idea as the one in https://github.com/netket/netket/pull/184#issuecomment-490957704, but this is not just pseudo-code but actually working):

As mentioned above, having the `AbstractOperator` work as a linear operator directly would be better in the long term.

```python
#!/usr/bin/env python3
import scipy
from scipy.sparse.linalg import LinearOperator as ScipyLinearOperator

class LinearOperator(ScipyLinearOperator):
    def __init__(self, wrapper):
        self.wrapper = wrapper
        dim = self.wrapper.dimension
        super().__init__(shape=(dim, dim), dtype=scipy.complex128)

    def _matvec(self, vec):
        return self.wrapper.apply(vec)

import netket as nk
from netket.operator import DirectMatrixWrapper

def as_linear_operator(op):
    return LinearOperator(DirectMatrixWrapper(op))

def test_as_linop():
    cube = nk.graph.Hypercube(3, 2)
    hilbert = nk.hilbert.Spin(cube, s=1/2)
    index = nk.hilbert.HilbertIndex(hilbert)
    hamiltonian = nk.operator.Heisenberg(hilbert)

    linop = as_linear_operator(hamiltonian)

    vec = scipy.rand(index.n_states)
    
    # Use as LinearOperator
    v1 = linop.dot(vec)

    # Compare to dense and sparse matrix-vector products
    v2 = hamiltonian.to_sparse().dot(vec)
    v3 = scipy.dot(hamiltonian.to_dense(), vec)

    assert scipy.allclose(v1, v2)
    assert scipy.allclose(v1, v3)
```

gcarleo:
Thanks Damian! I think this is really good for now. We will work in a separate PR the automatic conversion to `LinearOperator`, it's not superurgent, I agree

PR #185, opened 2019-05-15
Generalize samplers to arbitrary function of machine

gcarleo:
This PR allows to sample from other functions of the machine. In general it is now possible to sample from F(machine(x)), for arbitrary real-valued F. By default, F(X)=|X|^2. e.g. 
Sampling from |machine(x)| is useful when dealing with density matrices for example.


twesterhout:
How about generalising the approach a bit further? Any function `F` for which `std::is_invocable_r<double, F, std::complex<double>>::value == true` can be used as `MachineNorm`. So why not just make `std::function<double(std::complex<double>)>` a parameter rather than an `int` denoting order?

gcarleo:
> How about generalising the approach a bit further? Any function `F` for which `std::is_invocable_r<double, F, std::complex<double>>::value == true` can be used as `MachineNorm`. So why not just make `std::function<double(std::complex<double>)>` a parameter rather than an `int` denoting order?

I originally wanted to pass just a `std::function` actually, but the problem is that at some point I needed to compare two functions, to make sure that the initialization of the sampler was correctly done. (see what happens in `Sampler/ExactSampler.hpp`) and this is not too safe, as far as I understand.... In practical cases however one never needs to sample from crazy functions of the machine, so I settled to just the norm int.

twesterhout:
Well, if a sampler already has `order` as a member variable, why pass it to `Sweep` as function argument as well?

As for weird functions... you can never predict user needs in advance. Nice thing about `std::function` is that pybind11 can wrap Python callables in `std::function`s which means you can experiment with different norms without ever touching C++!

gcarleo:
> Well, if a sampler already has order as a member variable, why pass it to Sweep as function argument as well?

That's the thing, I don't want to pass order in the constructor, but only in `Sweep`, because that's more flexible. However there might be some samplers, like `ExactSampler`, that can speed up the sampling if they know in advance `order`, so that's why it is stored as a private variable in there. Notice that this variable is just used to check if the current order requested by `Sweep` is equal to previous one, if it is not, it recomputes some quantities from scratch, otherwise it can do it faster. 

If I do this with a generic function, I'd need to check if the currently stored function is the same I am asking for when I call `Sweep`...

twesterhout:
> That's the thing, I don't want to pass order in the constructor, but only in Sweep, because that's more flexible.

Could you, please, give an example when using a member variable is not flexible enough?

The design I have in mind is having a function as a member variable + a member function allowing you to change it. This functionality can even be implemented in some base class, no need to redo it in every sampler. Then, 1) if you want to change the norm in the middle of a Monte Carlo simulation, just call the setter; and 2) `ExactSampler` can cache values since it knows the norm beforehand. It also knows when to clear the cache -- whenever someone calls the setter.

gcarleo:
You are right, that's a better solution. I am doing that

gcarleo:
@twesterhout this is closer to a final PR, however I have some weird run-time error when exposing the `std::function` to pybind. Maybe for the moment it's easier not to bind that functionality, and keep it only for C++...

gcarleo:
@twesterhout , unless there are further objections I am going to merge this tomorrow

NetKet/Sampler/abstract_sampler.hpp
```diff
@@ -50,8 +51,35 @@ class AbstractSampler {
...
+  virtual void SetMachineFunc(
+      std::function<double(const Complex&)> machine_func) {
+    machine_func_.Set(machine_func);
+  }
+
+  virtual std::function<double(const Complex&)> GetMachineFunc() {
+    return machine_func_.Get();
+  }
+
+  class MachineFunc {
```
twesterhout: What is the purpose of this wrapper class? `using MachineFunc = std::function<double(Complex)>;` should do the trick.

NetKet/Sampler/abstract_sampler.hpp
```diff
@@ -50,8 +51,35 @@ class AbstractSampler {
...
+
+    void Set(std::function<double(const Complex&)> machine_func) {
+      mf_ = machine_func;
+    }
+    std::function<double(const Complex&)> Get() { return mf_; }
+  };
+
  protected:
   default_random_engine& GetRandomEngine() { return engine_.Get(); }
+  MachineFunc machine_func_;
```
twesterhout: Random number generator `engine_` is private, but `machine_func_` is exposed to the children, why is that? I think `GetMachineFunction` should be used to obtain a reference to `machine_func_` similar to how `GetRandomEngine` works.

One of the lesser known facts about `std::function` is that it can have a empty state. This is undesirable in our case. By initialising `machine_func_` with `std::norm` and checking for the empty state in `SetMachineFunc` we can prevent ever having an empty state.

NetKet/Sampler/abstract_sampler.hpp
```diff
@@ -50,8 +51,35 @@ class AbstractSampler {
     this->Reset(true);
   }
 
+  virtual void SetMachineFunc(
+      std::function<double(const Complex&)> machine_func) {
+    machine_func_.Set(machine_func);
+  }
+
+  virtual std::function<double(const Complex&)> GetMachineFunc() {
```
twesterhout: This should return a const reference, be marked `const noexcept`, and shouldn't be virtual.

NetKet/Sampler/abstract_sampler.hpp
```diff
@@ -50,8 +51,35 @@ class AbstractSampler {
     this->Reset(true);
   }
 
+  virtual void SetMachineFunc(
```
twesterhout: Don't forget to check for empty state here! Also, you should probably use move constructor instead of copying. I.e.
```cpp
if (!machine_func) {
    throw SomeError{};
}
machine_func_ = std::move(machine_func);
```

P.S. Maybe even `machine_func_.swap(machine_func)` instead of using the assignment operator, because `operator=` is not `noexcept`, but `swap` is. And `machine_func` is destroyed anyway so it doesn't matter. Here, I'm assuming you got rid of the wrapper class as suggested in a comment below.

twesterhout: ~And `SetMachineFunc` shouldn't be virtual.~

Ah, `ExactSampler` overrides `SetMachineFunc`... For now, let's leave it `virtual` then, but `ExactSampler` should then also check for empty state. In the future, code duplication can be avoided by using an extra level of indirection, but it can be discussed in another PR since we're not claiming ABI stability.

NetKet/Sampler/exact_sampler.hpp
```diff
@@ -133,6 +133,11 @@ class ExactSampler : public AbstractSampler {
     }
     return acc;
   }
+
+  void SetMachineFunc(std::function<double(const Complex&)> machine_func) {
```
twesterhout: Should be marked `override`. Also, see comments above regarding the empty state of `std::function`.

Test/Machine/test_machine.py
```diff
@@ -82,11 +82,35 @@
 np.random.seed(12346)
 
 
+def same_derivatives(der_log, num_der_log):
+    assert(np.max(np.real(der_log - num_der_log))
+           == approx(0., rel=1e-4, abs=1e-4))
```
twesterhout: Since we're using `double`s everywhere, is `1e-4` low enough?

NetKet/Sampler/py_sampler.hpp
```diff
@@ -86,14 +88,20 @@ void AddSamplerModule(py::module &m) {
...
+                       and distributed according to $$F(\Psi(v))$$ )EOF")
       .def_property_readonly("acceptance", &AbstractSampler::Acceptance, R"EOF(
         numpy.array: The measured acceptance rate for the sampling.
         In the case of rejection-free sampling this is always equal to 1.  )EOF")
       .def_property_readonly("hilbert", &AbstractSampler::GetHilbert, R"EOF(
         netket.hilbert: The Hilbert space used for the sampling.  )EOF")
       .def_property_readonly("machine", &AbstractSampler::GetMachine, R"EOF(
-        netket.machine: The machine used for the sampling.  )EOF");
+        netket.machine: The machine used for the sampling.  )EOF")
+      .def_property("machine_func", &AbstractSampler::GetMachineFunc,
```
twesterhout: Is this property still causing run-time errors? (I can't check it myself right now, unfortunately). If so, probably better to remove it for now.

NetKet/Sampler/abstract_sampler.hpp
gcarleo: It's mostly because I wanted machine_func_ to be initialized by default with std::norm, so I created a wrapper class with such a default initialization (instead of a null function default initialization)

NetKet/Sampler/abstract_sampler.hpp
gcarleo: Sure but I don't want that all the derived samplers need to initialize it on its default value, I would like that to be done in the abstract class somehow

NetKet/Sampler/py_sampler.hpp
gcarleo: no, it works... there is also a unit test that checks that, we can leave it

NetKet/Sampler/abstract_sampler.hpp
twesterhout: Default constructor of `AbstractSampler` can do that.

gcarleo: good point

NetKet/Sampler/abstract_sampler.hpp
gcarleo: actually I would make `engine_` protected too, I don't see a valid reason for not just using it without using setters and getters, maybe I am missing something?

Test/Machine/test_machine.py
gcarleo: yeah it seems that we can't go much below 1.0e-6 though, I changed that. The thing is that these functions are typically highly non-linear and higher order terms can be very large in finite differences approximations

NetKet/Sampler/abstract_sampler.hpp
gcarleo: I think @femtobit designed `engine_` to be private and not protected, can we change it to protected ?

twesterhout: @gcarleo typically, the reason to use getters and setters rather than exposing a member variable is to impose pre- and post-conditions. So the question is whether we trust the implementation of derived classes. If we ever wish to allow extending NetKet through C++, we should probably go the safer (getter + setter) route. Otherwise, NetKet maintainers are the only people who'll ever touch `machine_func_` and we can make it protected.

gcarleo: fine, this is really too much precaution, but anyway... it's done!

fabienalet:
Hi @gcarleo 
Sorry I was just testing this today... I think a working example with a machine function in the python help would be nice (I couldn't find one, but maybe I missed it).
Something like
```
sa=nk.sampler. ...
# if one wants to change the sampling function of the machine
myfunc = lambda x: np.absolute(x) 
sa.machine_func=myfunc
```
Also, when running a ground-state variational calculations, all expectations values of operators will not be correct  if the machine function is not |\Psi|^2, except those which commute with the Hamiltonian, is this correct ? If yes, perhaps it would be worth mentioning it somewhere.

In the long term, it could be nice to have an explicit example where a non-|\Psi|^2 sampling is useful.

PR #186, opened 2019-05-16
Improve Quantum State Reconstruction and add convenient numpy wrappers

gcarleo:
This PR improves the quantum state reconstruction interface, examples, and scripts. 
It also adds a few useful wrappers/converts to numpy/scipy types for both `AbstractOperator` (that can be converted to sparse, dense matrix and linear operators) and `AbstractMachine` that can be converted to a numpy array.


gcarleo:
This is finally ready to be merged @femtobit @GTorlai , let me know if there is something you'd like to change, otherwise I'll merge this tomorrow or later this week

PR #187, opened 2019-05-17
Move HilbertIndex to AbstractHilbert

femtobit:
This PR contains the first step towards getting rid of the `DirectMatrixWrapper` by removing the need of creating `HilbertIndex` objects outside `AbstractHilbert`.

With the changes in this PR, only one `HilbertIndex` per Hilbert space is created. This happens when `AbstractHilbert::GetIndex()` is first called. For non-indexable Hilbert spaces (users can check this by calling `AbstractHilbert::IsIndexable()`) this will throw an exeption, which I think is appropriate.

In Python, the Hilbert index can be accessed via the `index` property (e.g. `dim = hi.index.n_states`).

Note that, due to the way `GetIndex()` is implemented, calling `to_dense`/`to_sparse` on an operator of a non-indexable Hilbert space will also throw an exeption.

The next step is to implement the linear operator functions for `AbstractOperator` directly. Let me know what you think of the PR so far, then I'll continue with that part.


Examples/Supervised/Ising/ed.py
```diff
@@ -33,9 +33,8 @@ def load_ed_data(L):
     ttargets = []
 
     tsamples = []
-    hind = nk.hilbert.HilbertIndex(hi)
-    for i in range(hind.n_states):
-        tsamples.append(hind.number_to_state(i).tolist())
+    for i in range(hi.index.n_states):
```
gcarleo: This is nice! but what would be **really cool** is an iterator

```python
for state in hi.states():
   tsample.append(state)
```

NetKet/Hilbert/abstract_hilbert.hpp
```diff
@@ -80,7 +105,10 @@ class AbstractHilbert {
 
   virtual const AbstractGraph &GetGraph() const noexcept = 0;
 
-  virtual ~AbstractHilbert() {}
+  virtual ~AbstractHilbert() = default;
+
+ private:
+  mutable nonstd::optional<HilbertIndex> index_;
```
gcarleo: mutable? why?

NetKet/Hilbert/hilbert_index.hpp
```diff
@@ -24,37 +24,20 @@
 #include <nonstd/span.hpp>
 #include <vector>
 
-#include "Hilbert/abstract_hilbert.hpp"
 #include "Utils/next_variation.hpp"
 
 namespace netket {
 
 class HilbertIndex {
-  const std::vector<double> localstates_;
```
gcarleo: Do we have code style guidelines for where to declare private variables in classes?

NetKet/Machine/py_machine.hpp
```diff
@@ -147,7 +147,7 @@ void AddMachineModule(py::module &m) {
            )EOF")
       .def("to_array",
            [](AbstractMachine &self) -> AbstractMachine::VectorType {
-             HilbertIndex hind(self.GetHilbert());
+             const auto &hind = self.GetHilbert().GetIndex();
              AbstractMachine::VectorType vals(hind.NStates());
 
              double maxlog = std::numeric_limits<double>::lowest();
```
gcarleo: also these operations here would be vastly more elegant with c++ range iterators...

NetKet/Hilbert/abstract_hilbert.hpp
femtobit: Calling `GetIndex()` should also work on a const reference (creating the HilbertIndex does not change any observable property of the Hilbert object). However, since `index_` needs to be modified the first time `GetIndex` is called, this would not work on a const object, unless it is declared as mutable.

The other option would be to create the HilbertIndex in the constructor, but then it will also be created even when it is never used (which could be ok, creating it on demand like here is just an optimization).

NetKet/Hilbert/hilbert_index.hpp
femtobit: According to https://google.github.io/styleguide/cppguide.html#Declaration_Order:
> A class definition should usually start with a public: section, followed by protected:, then private:. Omit sections that would be empty.

Examples/Supervised/Ising/ed.py
femtobit: Agreed. I can give it a try.

NetKet/Hilbert/abstract_hilbert.hpp
femtobit: See also here: https://isocpp.org/wiki/faq/const-correctness#mutable-data-members

NetKet/Machine/py_machine.hpp
femtobit: The code in `to_array` also uses the index `i`, so this would not work well as a range-based for loop.
Still, with the commits I just pushed, one can write
```cpp
for (const auto state : hilbert) {
    // ...
}
```
to iterate over all configurations.

NetKet/Hilbert/abstract_hilbert.hpp
gcarleo: I see cool, then there is many other places we should use that too :)

NetKet/GroundState/py_exact.hpp
```diff
@@ -64,11 +64,11 @@ void AddExactModule(py::module &m) {
                >>> L = 20
                >>> graph = nk.graph.Hypercube(L, n_dim=1, pbc=True)
                >>> hilbert = nk.hilbert.Spin(graph, 0.5)
-               >>> idx = nk.hilbert.HilbertIndex(hilbert)
+               >>> n_states = hilbert.index.n_states
```
gcarleo: honestly, at this point I would go for removing completely the index object and make those just members of `hilbert`. Would there be an issue to do that? Especially if the space cannot be indexed?

femtobit: Sounds good. I did this for the Python interface. The `HilbertIndex` object still exists in C++, because I think it is nice to have all the methods that only work for indexable spaces in one place.

gcarleo:
This is failing because the numdifftools hasn't released a new version in a while, and they still use a function in scipy they are not supposed to use. I guess we will need to write our own function to compute numerical gradients in tests... I'll do that later

femtobit:
> It looks good now (tests are failing because of a problem in numdifftools I fixed already in the other PR, not a big deal, I can fix this when merging in master). Can I merge this into improve_qst for the moment or do you need to do further work?

Thanks for looking into the issue with `numdifftools` and feel free to merge this PR into `improve_qst`.

PR #190, opened 2019-05-22
Automatic formatting of Python source code with Black

femtobit:
I have recently encountered the [Black](https://black.readthedocs.io/en/stable/) code formatter for Python.

Since we already use clang-format for the C++ part of the code, which seems to work very well, I would like to suggest we discuss whether we want to introduce automatic formatting for our Python code too.
Using Black makes it a lot easier to have a consistent code style throughout our Python code, regardless of the author (and consistent with the official [PEP 8](https://www.python.org/dev/peps/pep-0008/) style guidelines).
Also, Black is written to format code in a way that diffs are small, which is useful for code review.

An [Atom package](https://atom.io/packages/python-black) is available and Visual Studio Code seems to support Black as part of the [official Python extension](https://marketplace.visualstudio.com/items?itemName=ms-python.python). It can be installed via `pip3 install black`. (Black itself requires Python 3 but it can also format Python 2 code.)

This PR shows what happens when letting Black reformat our current codebase. I have also included a proposed section on Black for `CONTRIBUTING.md`.

Let me know what you think about this proposal.


jamesETsmith:
@femtobit this is a really good idea! I hadn't heard of `black` before and I usually work with `yapf`, but `black` looks great.

gcarleo:
thanks @femtobit, I agree that this is something we should look into. @jamesETsmith , indeed `yapf` looks like a much more mature project than `black` and maybe a better option? what are the main differences?

CONTRIBUTING.md
```diff
@@ -103,6 +103,18 @@ which can be configured with Google style.
 To do so, after having installed the plugin, go to
 `Atom -> Preferences... -> Packages -> clang-format -> Settings` and type `google` in the `Style` field.
 
+#### Python coding style
+
+Python code should follow [PEP 8](https://TODO).
+For consistency, we use the [black](https://TODO) code formatter, which you can install in you Python environment using `pip install black`.
+If you add or change Python code, you should run black via
```
twesterhout: I think it's "you can" rather "you should", because for example I've never used `bash` to run `black` in my entire life and I've been using `black` extensively for the last ~9 months.

femtobit:
So far, I have only used Black a couple of times and YAPF not at all, so I cannot compare them from experience.

However, what I like about Black is that it is opinionated (it has a clear style and not many configuration options), which leads to a uniform look of Python code even across projects (and prevents discussion about which settings to use).

While Black is still in beta, [according to the documentation](https://black.readthedocs.io/en/stable/installation_and_usage.html#note-this-is-a-beta-product) this mostly means that the formatting may still change a bit between releases (but beside bug fixes, the developers said this is not likely). I think we can live with that (in fact, we have already lived with this in the case of `clang-format`). Besides, it is already used by some major Python projects, so it probably is mature enough (and I like that it checks whether the reformatted code is still equivalent to the original one by checking the AST).

So, while I don't have a strong opinion in this matter, I would at least slightly prefer trying out Black over YAPF.

PR #193, opened 2019-06-05
Split netket into Python and C++ parts

twesterhout:
At some point (actually, pretty soon) we'll probably have some high-lever functionality which is simpler to implement in Python than in C++. Not simple enough though to require every user to re-implement it themselves.

It is then nice to convert NetKet into a "proper" Python package (i.e. with `__init__.py` files) and use the C++ extension just as an implementation detail. This PR achieves that. C++ code is compiled into `_C_netket` module which is loaded by `__init__.py`.

There's a bunch of one-line-long files in `netket` directory. The mirror the module structure of our C++ code in Python. This means that stuff like `from netket.layer import DenseLayer, LogCosh` still works. If someone knows a cooler way to achieve, I'm open to suggestions. But for now, that's the best I could come up with.

P.S.: Formatting changes in `setup.py` are made by Black. It looks like #190 didn't reformat `setup.py` properly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -149,8 +149,9 @@ install:
   - python -m pip install numpy networkx pytest python-igraph
   - python -m pytest --version
   - NETKET_CMAKE_FLAGS="$CMAKE_FLAGS" python -m pip install -v .
+  - mkdir workdir && cd workdir
 
 script:
   - python -c 'import netket; g = netket.graph.Hypercube(4, 1); h = netket.hilbert.Spin(g, 0.5); m = netket.machine.RbmSpin(h, 10)'
-  - python -m pytest --verbose Test/
-  - python -m pytest --verbose --doctest-glob='*.md' Docs/
+  - python -m pytest --verbose ../Test/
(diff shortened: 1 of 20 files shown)


gcarleo:
@twesterhout this looks good, just a question: can we maybe move the one-line files in the corresponding folders? Say hilbert.py -> Hilbert/hilbert.py etc?

twesterhout:
> @twesterhout this looks good, just a question: can we maybe move the one-line files in the corresponding folders? Say hilbert.py -> Hilbert/hilbert.py etc?

@gcarleo we can, but then we'd have to write `from netket.Hilbert.hilbert import Spin` in place of `from netket.hilbert import Spin`. So I'd rather not do that.

gcarleo:
> but then we'd have to write from netket.Hilbert.hilbert import Spin in place of from netket.hilbert import Spin. So I'd rather not do that.

Alright, that would be ugly, I agree. I was just hoping that we could move the files without going through this extra indirection when importing. But it's fine... it's not a big deal, if we find a better solution in the future we will remove these one-line files

gcarleo:
merge it when you are ready

PR #196, opened 2019-06-05
Fix some compilation warnings

femtobit:
This commit fixes warnings related to
 * unnecessary semicolons,
 * a signed/unsigned integer comparison, and
 * default parameters in a lambda function (only supported since C++14).


femtobit:
I don't know why GItHub shows the Travis build as pending. Following the Details link shows that is has completed successfully.

PR #197, opened 2019-06-05
Rename C++ sources directory from NetKet to Sources

femtobit:
This prevents a conflict with the Python sources directory which is named `netket` (which, aside from being confusing, creates conflicts on case-insensitive file systems).


gcarleo:
@PhilipVinc don't hate us

PhilipVinc:
...
Il 5 giu 2019, 22:12 +0200, Giuseppe Carleo <notifications@github.com>, ha scritto:

Issue #195, opened 2019-06-05
Make netket::StateIterator an iterator


twesterhout:
Just noticed that `netket::StateIterator` doesn't satisfy the InputIterator concept, but is used as one.

PR #200, opened 2019-06-06
Move implementation of .iter methods on driver classes to Python

femtobit:
Implementing the iterator interface for the simulation drivers is quite a bit more complicated in C++ than to do it in Python (also compare #195, #198).

Now that we also have a straightforward way to add functionality in Python (#193), we can replace this with, essentially, something like
```python
for i in range(n_iter):
    driver.advance()
    yield i
```
which is done in this PR.

### DIFF
--- a/Sources/GroundState/imaginary_time.hpp
+++ b/Sources/GroundState/imaginary_time.hpp
@@ -13,8 +13,6 @@ namespace netket {
 
 class ImagTimePropagation {
  public:
-  class Iterator;
-
   using StateVector = Eigen::VectorXcd;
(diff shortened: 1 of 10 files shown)


gcarleo:
There is an iterator also in `Unsupervised/quantum_state_reconstruction.hpp`

gcarleo:
Sorry can you also update the Examples for both vmc and unsupervised? (do they need changes actually?)

femtobit:
> Sorry can you also update the Examples for both vmc and unsupervised? (do they need changes actually?)

They shoudn't need updates. This PR should not change the behavior of the simulation methods. (With the exception of fixing an off-by-one error.)

PR #218, opened 2019-06-13
Extract VMC functionality from driver class

femtobit:
This PR extracts the VMC functionality of NetKet from the driver class. Instead, separate functions implementing the VMC sampling and estimation of expectation values, variances, and gradients are now provided in `vmc_sampling.{hpp,cc}` as part of the `netket::vmc` namespace.

The Python functions still need documentation, which I will add soon. In the meantime, let me know what you think of the rest of the PR.


gcarleo:
Thanks @femtobit this is a necessary step indeed! We need however to be careful about the general reusability of these intermediate functions you have introduced, otherwise multiplying the number of functions wouldn't pay off, eventually. For example, can we reuse these for the unsupervised learning or for the Supervised one? Ideally, it should be so, especially when it comes to measuring <O> or related

femtobit:
Indeed, reusability of the functions is highly desirable here. Since they are relatively general (computing expectation, variance, and gradient given samples from the wavefunction), they should be useful in other pplaces.

Though I would not say this PR increases the complexity of the code. It actually decreases it. `vmc::ComputeSamples` is mostly equivalent to `VariationalMonteCarlo::Sample` and essentially there are now the functions `Expectation`, `Variance`, and `Gradient` (having two overloads, but that's just for optionally reusing precomputed local values), which contain (parts of) the code that was in the driver class. These functions are nice to have available directly in Python anyway, which they now are.

It is probably not too efficient, but we can for example write a quick VMC loop in Python now:
```python
from netket.hilbert import Spin
from netket.operator import Ising
from netket.machine import RbmSpin
from netket.sampler import MetropolisLocal
from netket.variational import compute_vmc_samples, expectation, gradient

hi = Spin(s=0.5, graph=Hypercube(8, 1))
ham = Ising(hi, h=1.0)
psi = RbmSpin(hi, alpha=2)
psi.init_random_parameters(sigma=0.1)
sampler = MetropolisLocal(psi)

for step in range(100):
    data = compute_vmc_samples(sampler, 10000, 1000)

    ex = expectation(data, psi, ham)
    print("E={Mean:.4f} ± {Sigma:.4f}".format(**ex))

    grad = gradient(data, psi, ham)
    psi.parameters -= 0.1 * grad
```

Sources/GroundState/py_variational_montecarlo.hpp
```diff
@@ -160,15 +160,186 @@ void AddVariationalMonteCarloModule(py::module &m) {
...
+            py::arg("nsamples"), py::arg("ndiscard") = 0);
+
+  m_vmc.def(
+      "expectation",
+      [](const vmc::Result &result, AbstractMachine &psi,
+         const AbstractOperator &op, bool return_locvals) {
+        if (return_locvals) {
+          VectorXcd locvals;
+          auto ex = vmc::Expectation(result, psi, op, locvals);
+          return py::object(py::make_tuple(ex, locvals));
```
twesterhout: Could you do `static_cast<py::object>(...)` instead of the C-style cast?

Sources/GroundState/py_variational_montecarlo.hpp
```diff
@@ -160,15 +160,186 @@ void AddVariationalMonteCarloModule(py::module &m) {
...
+                 ex, lv = expectation(data, psi, ham)
+                 var = variance(data, psi, ham, ex["Mean], lv)
+               ```
+            )EOF");
+
+  using GradType1 = VectorXcd (*)(const vmc::Result &, AbstractMachine &,
+                                  const AbstractOperator &);
+  m_vmc.def("gradient", (GradType1)&vmc::Gradient, py::arg("vmc_data"),
+            py::arg("op"), py::arg("psi"), R"EOF(
+           Computes the expectation gradient a Hermitian operator ∇⟨O⟩
```
twesterhout: Could you rephrase, please?

Sources/GroundState/py_variational_montecarlo.hpp
```diff
@@ -160,15 +160,186 @@ void AddVariationalMonteCarloModule(py::module &m) {
...
+
+  m_vmc.def(
+      "local_values",
+      [](const vmc::Result &result, AbstractMachine &psi,
+         const AbstractOperator &op) {
+        return vmc::LocalValues(op, psi, result.SampleMatrix());
+      },
+      py::arg("vmc_data"), py::arg("psi"), py::arg("op"), R"EOF(
+           Computes the local values of the operator `op` for all visible
+           configurations sotred in `vmc_data`.
```
twesterhout: stored*

Sources/GroundState/vmc_sampling.cc
```diff
@@ -0,0 +1,165 @@
+#include "vmc_sampling.hpp"
+
+namespace netket {
+namespace vmc {
+
+Result ComputeSamples(AbstractSampler &sampler, Index nsamples,
+                      Index ndiscard) {
```
gcarleo: I think we should an option to compute or not compute the log derivatives in this loop (by default I would say we shouldn't, and VmcResult should make sure they are computed if one asks for them later on?)

Sources/GroundState/vmc_sampling.cc
```diff
@@ -0,0 +1,165 @@
+#include "vmc_sampling.hpp"
```
gcarleo: I guess that vmc_sampling and the more general routines should sit at different level than `GroundState`. Maybe `Sources/MonteCarlo/machine_sampling.cc` or something similar? 
In this way it is also more obvious that these routines can be reused in other type of learning algorithms, beyond ground state

gcarleo: Even `Sources/Sampler/machine_sampling.cc` would be a good option, maybe even better than creating a new folder

femtobit: Moved the code to `Sampler/`.

Sources/GroundState/vmc_sampling.cc
femtobit: There is now an option to disable computation of the log-derivatives.

Sources/GroundState/py_variational_montecarlo.hpp
```diff
@@ -160,15 +160,198 @@ void AddVariationalMonteCarloModule(py::module &m) {
...
           R"EOF(
         Calculate and return the value of the operators stored as observables.
 
-        )EOF");
+        )EOF")
+      .def_property_readonly("vmc_data", &VariationalMonteCarlo::GetVmcData);
+
+  py::class_<vmc::Result>(m_vmc, "_VmcResult");
+
+  m_vmc.def("compute_samples", vmc::ComputeSamples, py::arg("sampler"),
```
gcarleo: is this `compute_samples` or `compute_vmc_samples` as in the docs?

Sources/GroundState/py_variational_montecarlo.hpp
```diff
@@ -160,15 +160,198 @@ void AddVariationalMonteCarloModule(py::module &m) {
...
+           Computes a sequence of visible configurations based on Monte Carlo sampling using `sampler`.
+
+           Args:
+               sampler: The sampler used to perform the MC sweeps.
+               nsamples: The number of MC samples that are stored (one MC sweep
+                    is performed between each sample)
+               ndiscard: The number of sweeps to be discarded before starting
+                    to store samples.
+               compute_logderivs: Whether to store the logarithmic derivatives
+                    of the wavefunction as part of the returned VMC result.
```
gcarleo: what does it happen if `compute_logderivs=False` and one tries to compute the gradient using this `vmc_data`?

Sources/GroundState/py_variational_montecarlo.hpp
femtobit: I renamed it, but forgot to change it in the docs.

gcarleo:
How come the Docs tests pass then??

Il giorno ven 14 giu 2019 alle ore 13:49 Damian Hofmann <
notifications@github.com> ha scritto:

Sources/GroundState/py_variational_montecarlo.hpp
femtobit: It will throw an exception, like in this test:
https://github.com/netket/netket/blob/12b8e68abf931c9ef44d078ac9bbd924b769c293/Test/GroundState/test_vmc.py#L60-L68

femtobit:
> How come the Docs tests pass then??

Ah, I did not update the Docs yet, good point.

femtobit:
> Ah, I did not update the Docs yet, good point.

Updating the docs currently fails, due to #222.

gcarleo:
Can you run build
docs only on the specific module then?

femtobit:
> Can you run build docs only on the specific module then?

Unfortunately, the module here is `netket.variational`, which also contains `Vmc.iter` which causes the issue.

femtobit:
Using @ooreilly's suggestion from #222, I git the docs to build here now. There was a failure caused by the docs of `NdmSpinSymm`, so unrelated to this PR, but I added a quick fix to get the tests to pass in 
e7e13cd .

femtobit:
@gcarleo The tests (and doctests) pass. Can we merge this PR now?

Sources/GroundState/py_variational_montecarlo.hpp
gcarleo: OK very good. I just have the feeling that by default this should be False and not True...

femtobit: I think the fact that most of our current code requires the log-derivatives suggests that computing them by default makes sense for us. Omitting them appears more like an optimization to me (which might be useful if your system is large and you really just need energy and variance).

PR #227, opened 2019-06-17
A fix for OpenMP problems on OS X

twesterhout:
For now, we can borrow `FindOpenMP.cmake` from PyTorch's repo. They maintain a patched version which seems to work on Travis CI with OS X. If I can lay my hands on a Mac, I'll check whether there's a simpler workaround we can use (see points about AppleClang here https://github.com/pytorch/pytorch/tree/master/cmake/Modules).

@gcarleo @PhilipVinc and other OS X users, please let me know whether this works for you.


gcarleo:
This works if one install openmp with `brew install libomp`, but otherwise it fails. 
I would strongly suggest to set up openmp as an optional library. If it finds it, we use it. Otherwise nope.

twesterhout:
> This works if one install openmp with brew install libomp, but otherwise it fails.
I would strongly suggest to set up openmp as an optional library. If it finds it, we use it. Otherwise nope.

@gcarleo we have MPI in our dependencies, so one has to do `brew install openmpi` anyway (or whichever implementation one uses). So what's the problem with doing `brew install openmpi libomp` instead? OpenMP is way more portable that MPI :)

twesterhout:
Also, PyTorch doesn't work without OpenMP, so if we're going to rely on it for NNs, we need OpenMP (see https://github.com/pytorch/pytorch/issues/20030).

gcarleo:
Alright, it's fine... we need to add a note in the web page though, I will do that with the next pip release... In any case, since we are here, would you mind adding numpy to the dependencies to be installed automatically? Thanks!

gcarleo:
@twesterhout ? Do we add the numpy dependency in setup.py here or in another PR?

twesterhout:
> Do we add the numpy dependency in setup.py here or in another PR?

@gcarleo in another PR I think.

PR #229, opened 2019-06-19
Getting rid of various matrix wrappers

twesterhout:


Sources/GroundState/imaginary_time.hpp
```diff
@@ -15,26 +14,57 @@ class ImagTimePropagation {
...
-  using ObsEntry = std::pair<std::string, std::unique_ptr<Matrix>>;
+  using Operator = std::function<StateVector(StateVector const&)>;
+  using ObsEntry = std::pair<std::string, Operator>;
+
+  static Operator MakeOperator(const AbstractOperator& op,
+                               const std::string& type) {
+    if (type == "dense") {
+      struct Function {
+        Eigen::MatrixXcd matrix;
+        StateVector operator()(StateVector const& x) const {
```
femtobit: `const StateVector&` :wink:

Sources/GroundState/imaginary_time.hpp
```diff
@@ -69,7 +99,7 @@ class ImagTimePropagation {
   void SetTime(double t) { t_ = t; }
 
  private:
-  std::unique_ptr<Matrix> matrix_;
+  std::function<StateVector(StateVector const&)> matrix_;
```
femtobit: `const StateVector&`

External/ietl/ietl/eigeninterface.h
```diff
@@ -44,7 +44,7 @@ void mult(const Eigen::SparseMatrix<TCoeffs>& a,
 template <class matrix_t, class TCoeffs>
 void mult(const matrix_t& a, const Eigen::Matrix<TCoeffs, Eigen::Dynamic, 1>& x,
           Eigen::Matrix<TCoeffs, Eigen::Dynamic, 1>& y) {
-  y = a.Apply(x);
+  y = a(x);
```
gcarleo: What's wrong with the method Apply? The operator () is really not suitable to indicate the intent of a multiplication. If you want to move from a method to an operator, I think this should be the * operator !

twesterhout: Changes to this particular file are temporary (I am working on removing dependencies on IETL altogether). `mult` is a template which we instantiate with `Eigen::MatrixXcd`, `Eigen::SparseMatrix<Complex>`, and `netket::AbstractOperator`. In such cases it's pretty standard to pass function objects around and they implement `operator()` rather than `Apply`.

gcarleo: I do not agree that this is a good naming, but it seems that you have already merged this

femtobit: @gcarleo Sorry. I merged the PR despite this point being still open since I was expecting this code to be removed anyway in #232.

PR #232, opened 2019-06-20
Replacing IETL with ARPACK

twesterhout:
This PR removes the dependency on IETL remaining fully API backward-compatible. All existing code should continue to work except that now SciPy's ARPACK wrappers are used under the hood. Also, as long as one doesn't call ED routines, having SciPy installed on the system is optional.


gcarleo:
I was wrong, it actually works, but it is substantially slower than the old version

gcarleo:
OK, the **actual**  problem is that arpack does not have a way to compute eigenvalues only, if both are asked to compute eigenvalues and eigenvectors in the matrix_free mode, then performances are equivalent

netket/exact.py
```diff
@@ -25,3 +65,132 @@ def _ImagTimePropagation_iter(self, dt, n_iter=None):
...
+    Args:
+        operator: The operator to diagnolize.
+        matrix_free: If true, matrix elements are computed on the fly.
+            Otherwise, the operator is first converted to a sparse matrix.
+        first_n: The number of eigenvalues to compute.
+        max_iter: The maximum number of iterations.
+        seed: **Ignored**. Accepted for backward compatibility only.
+        precision: The precision to which the eigenvalues will be
+            computed.
+        compute_eigenvectors: Whether or not to compute the
```
gcarleo: we should comment on the fact that this option does not have any influence on the actual computation of the eigenvectors, when using the scipy backend

gcarleo:
@twesterhout OK, looks great actually, we can maybe just add that comment about computing eigenvalues only, otherwise it's good to go!

@alexwie what do you think? This is mostly in the spirit of simplifying our reliance on external (and possibly unmaintained) libraries...

netket/exact.py
```diff
@@ -1,6 +1,46 @@
...
 
+import warnings
+import functools
 import itertools
+import inspect
+
+from ._C_netket.exact import *
+
+# NOTE: If more modules end up requiring this functionality, we can create a
+# `netket._core` module and move `deprecated` there.
```
femtobit: Why not just do that right now?

netket/exact.py
```diff
@@ -25,3 +65,134 @@ def _ImagTimePropagation_iter(self, dt, n_iter=None):
...
+def lanczos_ed(
+    operator,
+    matrix_free=False,
+    first_n=1,
+    max_iter=1000,
+    seed=None,
+    precision=1e-14,
+    compute_eigenvectors=False,
+):
+    r"""Uses Lanczos algorithm to diagonalize the operator.
```
femtobit: Strictly speaking, computing the first `n` eigenvalues (or even eigenvectors) does not diagonalize the operator. Only computing the full basis would. So maybe this comment should be more specific (e.g., `Computes the first n eigenvalues and, optionally, eigenvectors of a Hermitian operator using the Lanczos method.`).

Issue #235, opened 2019-06-21
Switch from internal ODE solver to scipy


gcarleo:
At this point we should move from our home-made implementation of Runge Kutta, and just rely on scipy.integrate.ode. Using `to_sparse`, `to_linear_operator` etc, it should be really straightforward to rewrite the exact real/imaginary time evolution in a pure python function!

gcarleo:
The only thing I find quite surprising is that the solvers from scipy (say [the Runge-Kutta one](https://docs.scipy.org/doc/scipy/reference/generated/scipy.integrate.RK45.html#scipy.integrate.RK45)) **don't have the possibility to do fixed-step updates**... (at least it seems to me, I hope I am wrong!). This is a bit of a problem, especially for stochastic applications where adaptive stepping can be completely unstable ! 

If this is true, we can either try to convince scipy to introduce fixed-step updates, (might take a long time) or have some other internal fixed-step solver... @femtobit what do you think?

gcarleo:
Done in v3.0

Issue #234, opened 2019-06-21
ImagTimePropagation segfaults


twesterhout:
The following code (on master branch) segfaults:

```python
import netket
import numpy

hilbert = netket.hilbert.Spin(netket.graph.Hypercube(8, n_dim=1, pbc=True), s=0.5)
hamiltonian = netket.operator.Ising(h=1.0, hilbert=hilbert)
d = netket.exact.ImagTimePropagation(
    hamiltonian,
    netket.dynamics.create_timestepper(hilbert.n_states),
    0.0,
    numpy.random.rand(hilbert.n_states),
)
d.advance(200)
```

Even if some of the arguments are invalid, the code shouldn't segfault.

gcarleo:
Can I also ask why `create_timestepper` is not just `timestepper`?

PR #236, opened 2019-06-21
Changes to exact time propagation (and some minor improvements)

femtobit:
This PR contains a bunch of mostly minor changes, mostly related to time evolution:

* The most noticeable one: Rename `ImagTimePropagation` to `ExactTimePropagation` and add option for solving the real-time Schrödinger equation. (In light of #235 this change may not be relevant for much longer, but since #235 is marked as version 2.1 and the work is already done, I think it still makes sense to merge it for now.)
* A fix for #234.
* Removing some typos from the docs.

* Making `UpdateConf` take `span`s, which removes the need for additional copies of data in a few cases.

### DIFF
--- a/Sources/GroundState/py_exact.hpp
+++ b/Sources/GroundState/py_exact.hpp
@@ -24,7 +24,7 @@
 #include <pybind11/stl.h>
 #include <pybind11/stl_bind.h>
 
-#include "imaginary_time.hpp"
+#include "Dynamics/exact_time_propagation.hpp"
 
 namespace py = pybind11;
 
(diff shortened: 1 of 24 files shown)


Sources/Hilbert/py_hilbert.cc
```diff
@@ -215,16 +218,31 @@ void AddHilbertModule(py::module m) {
 
            ```
        )EOF")
-          .def("update_conf", &AbstractHilbert::UpdateConf, py::arg("v"),
-               py::arg("to_change"), py::arg("new_conf"), R"EOF(
+          .def(
+              "update_conf",
+              [](const AbstractHilbert &self, Eigen::Ref<Eigen::VectorXd> v,
+                 PyCArray<int> tochange, PyCArray<double> newconf) {
+                if (tochange.ndim() > 1 || newconf.ndim() > 1) {
```
gcarleo: making these checks at every function call can be rather expensive, are we sure we don't want to put those only in debug mode?

gcarleo: Since we know these are 1d arrays, we should maybe use this https://pybind11.readthedocs.io/en/stable/advanced/pycpp/numpy.html#direct-access for performance reasons

femtobit: I would expect the overhead of these checks to be insignificant compared to the function call overhead when are calling C++ from Python, which we have here anyway.

The access to the data within `AbstractHilbert::UpdateConf` is done via the `span`s constructed below, which omit bounds checking (just as the direct access functions you mentioned would).

Sources/GroundState/py_exact.hpp
```diff
@@ -55,29 +56,34 @@ void AddExactModule(py::module &m) {
...
                ```python
                >>> import netket as nk
                >>> import numpy as np
-               >>> L = 20
+               >>> L = 8
                >>> graph = nk.graph.Hypercube(L, n_dim=1, pbc=True)
                >>> hilbert = nk.hilbert.Spin(graph, 0.5)
                >>> n_states = hilbert.n_states
                >>> hamiltonian = nk.operator.Ising(hilbert, h=1.0)
                >>> stepper = nk.dynamics.create_timestepper(n_states, rel_tol=1e-10, abs_tol=1e-10)
```
gcarleo: Alright, since we will address the scipy ode integrator in a later (but just a few months away!) release, we should maybe fix the APIs at this stage. Can we rename `create_timestepper` -> `timestepper` ? It seems to me that the action of creating is implicit in the constructor of the object (we don't have `create_rbm` etc)

femtobit: Ok, I have renamed it.

gcarleo: Thanks, great!

Issue #231, opened 2019-06-19
Add dependencies to setup.py


twesterhout:
Our setup.py files doesn't list the dependencies required for compiling NetKet (see https://github.com/netket/netket/pull/227#issuecomment-503617668).

PR #238, opened 2019-06-25
Add requirements to setup.py and change ED example to use scipy

gcarleo:


gcarleo:
close #231

gcarleo:
@twesterhout the quick fix for your PyRbm didn't work, apparently it breaks the derivatives

PR #240, opened 2019-07-02
Change sweep size to be odd by default

gcarleo:
Correcting a potential ergodicity issue in samplers arising when the size of the sweep is even

### DIFF
--- a/Sources/Sampler/custom_sampler.hpp
+++ b/Sources/Sampler/custom_sampler.hpp
@@ -179,7 +179,7 @@ class CustomSampler : public AbstractSampler {
   }
 
   void SetSweepSize(int sweep_size) {
-    if (sweep_size_ <= 0) {
+    if (sweep_size <= 0) {
       throw InvalidInputError{"Sweep size should be a positive integer."};
     }
     sweep_size_ = sweep_size;
--- a/Sources/Sampler/custom_sampler.hpp
+++ b/Sources/Sampler/custom_sampler.hpp
@@ -55,6 +55,8 @@ class CustomSampler : public AbstractSampler {
   int nstates_;
   std::vector<double> localstates_;
 
+  int sweep_size_;
+
  public:
   CustomSampler(AbstractMachine &psi, const LocalOperator &move_operators,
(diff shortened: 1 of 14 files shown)


Sources/Sampler/custom_sampler.hpp
```diff
@@ -101,6 +103,13 @@ class CustomSampler : public AbstractSampler {
     nstates_ = GetHilbert().LocalSize();
     localstates_ = GetHilbert().LocalStates();
 
+    // Always use odd sweep size to avoid possible ergodicity problems
+    if (nv_ % 2 == 0) {
```
twesterhout: This can be written somewhat shorter as `SetSweepSize(nv_ + (nv_ % 2 == 0))`, but if you find the current version easier to read, let's stick to it.

Sources/Sampler/custom_sampler.hpp
```diff
@@ -168,6 +178,15 @@ class CustomSampler : public AbstractSampler {
     return acc;
   }
 
+  void SetSweepSize(int sweep_size) {
+    if (sweep_size <= 0) {
+      throw InvalidInputError{"Sweep size should be a positive integer."};
```
twesterhout: It's helpful to actually print `sweep_size`. Something like this:
```cpp
std::ostringstream msg;
msg << "invalid sweep size: " << sweep_size << "; expected a positive integer";
throw InvalidInputError{msg.str()};
```

Sources/Sampler/custom_sampler.hpp
```diff
@@ -168,6 +178,15 @@ class CustomSampler : public AbstractSampler {
...
+    }
+    sweep_size_ = sweep_size;
+  }
+
+  int GetSweepSize() const { return sweep_size_; }
```
twesterhout: This function should be marked `noexcept`.

Sources/Sampler/custom_sampler.hpp
gcarleo: Yes, let's stick to it

PR #246, opened 2019-07-09
Batched versions of RbmSpin and MetropolisLocal

twesterhout:
### Breaking changes

* `variational`
    * `compute_samples` returns a `MCResult` type instead of `_VmcResult` (`netket::vmc::Result` in C++). `netket::vmc::Result` didn't expose any methods or data members to Python and was just completely useless for end users. `MCResult` exposes all the data obtained during sampling.
    * `compute_samples` argument names changed: `nsamples` is now `n_samples` and `ndiscard` is replaced by `n_discard` (as requested [here](https://github.com/netket/netket/pull/246/files#r307822759)).
    * Number of samples returned from `compute_samples` may differ from the argument value for samplers with batch sizes greater than 1.
    * `expectation` and `variance` functions are removed. A combination of calls to `local_values` and `statistics` can be used instead.
    * Interface of `gradient` function has changed.
    * Interface of `local_values` function has changed.
    * `local_value` function is removed. `local_values` can be used instead.
    * Added `statistics` function for calculating means, variances, errors, correlation times, etc. of Monte Carlo data.
    
* `machine`
    * order of arguments in `init_random_parameters` has changed. Moreover, it is no longer necessary to provide a `seed`. Random number generator will be initialised from `/dev/random` (or `/dev/urandom`).
    * Machines now have a `state_dict` function which returns internal state as a dictionary (similar to `torch.nn.Module.state_dict` in PyTorch). Currently, only `RbmSpinV2` supports it. All other machines throw errors saying that the requested functionality is not implemented.
    * `log_val` and `der_log` can now receive either 1D or 2D tensors corresponding to a single visible configuration or a batch of visible configurations.

* `sampler`
    * `machine_func` now works with arrays instead of scalars.
    * `acceptance` is no longer part of `Sampler`, but most of the samplers still implement it.
    * `visible` property is now a 2D tensor and behaves slightly differently.
    * Removed `hilbert` property. Use `.machine.hilbert` instead.

* probably more

### Code changes

* Machines are free to use their own custom type for lookup tables. It is passed around using `netket::any` (a C++11 port of `std::any`).
* `AbstractMachine::LogVal` is renamed to `LogValSingle`. `LogVal` now works with batches of visible configurations. `AbstractMachine` contains default implementation for `LogVal` in terms of `LogValSingle`.
* `AbstractMachine::DerLog` is renamed to `DerLogSingle`. `DerLog` now works with batches of visible configurations. `AbstractMachine` contains default implementation for `DerLog` in terms of `DerLogSingle`.
* `RbmSpinV2` -- a safer and more efficient version of `RbmSpin`.
* Stochastic reconfiguration is now exposed to Python as `SR`.
* Samplers now have a `BatchSize` method.
* Samplers now have `CurrentState` function which returns both the current visible configuration and the corresponding logarithm of the wavefunction.
* etc.

### New dependencies
* [Sleef](https://sleef.org/) vectorized maths library. Currently used only for implementing faster activation function for RBMs. It also has a DFT component which might be useful if at some point we decide to estimate correlation time using lag-t correlations.
* [linb-any](https://github.com/thelink2012/any) implementation of C++17 `std::any` for C++11-compatible compilers.


gcarleo:
Thanks @twesterhout ! Can we already use the machine and the sampler for full variational calculations?Do you have an example file for testing?

Sources/Machine/rbm_spin_v2.cc
```diff
@@ -0,0 +1,325 @@
...
+    *b_ = parameters.segment(i, b_->size());
+    i += b_->size();
+  }
+  Eigen::Map<Eigen::VectorXcd>(W_.data(), W_.size()) =
+      parameters.segment(i, W_.size());
+}
+
+void RbmSpinV2::LogVal(Eigen::Ref<const RowMatrix<double>> x,
+                       Eigen::Ref<Eigen::VectorXcd> out, const any &) {
+  if (x.cols() != Nvisible()) {
```
gcarleo: Worth moving these checks in a recyclable function to be called from derivatives too 
[I see you have done this for other checks, I guess you will do it here too]

Sources/Machine/rbm_spin_v2.cc
```diff
@@ -0,0 +1,325 @@
...
+                         .array()
+                         .tanh();
+    out.block(0, i, BatchSize(), b_->size()) = theta_;
+    i += b_->size();
+  } else {
+    theta_.array() = theta_.array().tanh();
+  }
+
+  // TODO: Rewrite this using tensors
+#pragma omp parallel for schedule(static)
```
gcarleo: maybe I'd remove explicit calls to omp, this might conflict with MPI / negatively affect performance on multiple cores?

Sources/Sampler/exact_sampler.hpp
```diff
@@ -89,7 +89,7 @@ class ExactSampler : public AbstractSampler {
 
     for (int i = 0; i < dim_; ++i) {
       auto v = hilbert_index_.NumberToState(i);
-      logpsivals_[i] = GetMachine().LogVal(v);
+      logpsivals_[i] = GetMachine().LogValSingle(v);
```
gcarleo: Reminder here to discuss wether it makes sense to have a single `LogVal` or add a `LogValSingle` version instead of an if statement in `LogVal`

gcarleo:
OK, I guess I can just use the `v2` in python and see. I would maybe suggest to create a new branch (v2.1) instead of merging to master

gcarleo:
` ``
../../Sources/Utils/log_cosh.cc:30:15: error: use of undeclared identifier 'Sleef_logd4_u35avx2'
      auto real = Sleef_logd4_u35avx2(
```
is there a workaround?

**Comment by twesterhout** [comment]
> ```
> ../../Sources/Utils/log_cosh.cc:30:15: error: use of undeclared identifier 'Sleef_logd4_u35avx2'
>       auto real = Sleef_logd4_u35avx2(
> ```
> 
> is there a workaround?

@gcarleo yes, this is the AVX2 issue I mentioned in the description. Could you try the following command, please?
```bash
lscpu | grep avx2
```
If avx2 is found, then you can simply add `-mavx2` or `-march=native` to C++ flags and the code should compile.

gcarleo:
OK, `NETKET_CMAKE_FLAGS="-DCMAKE_CXX_FLAGS='-march=native'" pip3 install .` for future reference

gcarleo:
OK, now it compiles but can you please provide a benchmark python script/example usage to get a handle on the new sampler?

twesterhout:
Here's the file I'm using locally (sorry for a long copy-paste, I'll split it into a test and example for the PR):
```python
import netket
import numpy
import timeit

import re
import mpmath
import math


def test_metropolis_local_v2():
    n = 20
    batch_size = 64
    n_samples = 200000
    n_discard = 1000
    test_values = True
    test_derivatives = True

    h = netket.hilbert.Spin(netket.graph.Hypercube(n, n_dim=1, pbc=True), s=0.5)
    m = netket.machine.RbmSpinV2(h, alpha=3)
    m.init_random_parameters()

    # Try passing invalid batch size
    try:
        sampler = netket.sampler.MetropolisLocalV2(m, batch_size=-12)
    except ValueError as e:
        assert re.match(r".*invalid batch size.*", str(e)) is not None
    else:
        assert False

    sampler = netket.sampler.MetropolisLocalV2(m, batch_size=batch_size)

    # Try passing invalid steps tuple
    try:
        result = netket.sampler.compute_samples_v2(
            sampler, (100, -4, 10), test_derivatives
        )
    except ValueError as e:
        assert re.match(r".*invalid steps range.*", str(e)) is not None
    else:
        assert False

    def mk_steps(nsamples, ndiscard):
        return (ndiscard, ndiscard + n * nsamples // batch_size, n)

    result = netket.sampler.compute_samples_v2(
        sampler, mk_steps(n_samples, n_discard), test_derivatives
    )
    if test_derivatives:
        samples, values, gradients = result
    else:
        samples, values = result

    expected_number_samples = batch_size * ((n_samples + batch_size - 1) // batch_size)
    print(samples.shape, expected_number_samples)
    # assert samples.shape == (expected_number_samples, n)
    # assert values.shape == (expected_number_samples,)
    # if test_derivatives:
    #     assert gradients.shape == (expected_number_samples, m.n_par)

    if test_values:
        for (x, y) in zip(samples, values):
            assert numpy.isclose(m.log_val(x), y)

    if test_derivatives:
        for (x, y) in zip(samples, gradients):
            assert numpy.all(numpy.isclose(m.der_log(x), y))

    def compute_counts(xs):
        counts = {}
        for x in xs:
            i = h.state_to_number(x)
            counts[i] = counts.get(i, 0) + 1
        return counts

    def normalisation():
        total = mpmath.mpf(0.0)
        for i in range(h.local_size ** h.size):
            total += mpmath.fabs(mpmath.exp(m.log_val(h.number_to_state(i)))) ** 2
        return total

    norm = normalisation()
    counts = compute_counts(samples)
    metric = 0.0
    for i in range(h.local_size ** h.size):
        X = counts.get(i, 0) / values.size
        Y = float(mpmath.fabs(mpmath.exp(m.log_val(h.number_to_state(i)))) ** 2 / norm)
        metric += (X - Y) ** 2

    metric = metric ** 0.5 / h.local_size ** h.size
    print(metric)
    assert metric <= 1e-5


def benchmark():
    batch_size = 16
    h = netket.hilbert.Spin(netket.graph.Hypercube(100, n_dim=1, pbc=True), s=0.5)

    m1 = netket.machine.RbmSpin(h, alpha=2)
    m1.init_random_parameters(seed=4321)
    m2 = netket.machine.RbmSpinV2(h, alpha=2, batch_size=batch_size)
    m2.parameters = m1.parameters

    op = netket.operator.Heisenberg(h)

    sampler1 = netket.sampler.MetropolisLocal(m1)
    sampler2 = netket.sampler.MetropolisLocalV2(m2, batch_size=batch_size)

    n_samples, n_discard = 2000, 100
    steps = (
        h.size * n_discard,
        h.size * n_discard + (h.size * n_samples + batch_size - 1) // batch_size,
        h.size,
    )

    def go1():
        result = netket.variational.compute_samples(
            sampler1, n_samples, n_discard, True
        )
        result = netket.variational.local_values(xs.T, m1, op)
        return result.mean()

    def go2():
        (xs, ys, grad) = netket.sampler.compute_samples_v2(sampler2, steps, True)
        result = netket.sampler.local_values_v2(xs, ys, m2, op, batch_size)
        return result.mean()

    # print(go1())
    # print(go2())
    print(numpy.array(timeit.repeat(lambda: go1(), number=3, repeat=1)))
    print(numpy.array(timeit.repeat(lambda: go2(), number=3, repeat=1)))


benchmark()
# test_metropolis_local_v2()
```

gcarleo:
Alight, then I'll wait for these to be added to the PR I guess
```
  File "test.py", line 124, in go1
    result = netket.variational.local_values(xs.T, m1, op)
NameError: name 'xs' is not defined
```

twesterhout:
> Alight, then I'll wait for these to be added to the PR I guess
> 
> ```
>   File "test.py", line 124, in go1
>     result = netket.variational.local_values(xs.T, m1, op)
> NameError: name 'xs' is not defined
> ```

@gcarleo you were too fast with copying. I just tried copy-pasting my message into a new file, and the code runs.

By the way, for optimal performance, you may have to play around with batch sizes and number of threads.

gcarleo:
Alright, no beer yet

twesterhout:
> Alright, no beer yet

Why not? For
```python
t1 = numpy.array(timeit.repeat(lambda: go1(), number=2, repeat=2))
t2 = numpy.array(timeit.repeat(lambda: go2(), number=2, repeat=2))
print(t1, t1.mean(), t1.std())
print(t2, t2.mean(), t2.std())
```
I get
```sh
✔␣twesterhout␣at␣sccloan005␣in␣~/src/netket␣on␣master␣$ OMP_NUM_THREADS=2 python3 foo.py
# RBM Initizialized with nvisible = 100 and nhidden = 200
# Using visible bias = 1
# Using hidden bias  = 1
# Local Metropolis sampler is ready 
[53.02048492 53.11186881] 53.06617686455138 0.04569194547366351
[4.89975831 4.85624072] 4.877999514574185 0.02175879350397736
```
I.e. more than 10x speed-up.

gcarleo:
```
(env) MBP-di-Giuseppe:Ising1d giuscarl$ OMP_NUM_THREADS=2 python3 test.py 
# RBM Initizialized with nvisible = 100 and nhidden = 200
# Using visible bias = 1
# Using hidden bias  = 1
# Local Metropolis sampler is ready 
[22.95861549 23.18452988]
[6.27776219 6.45110383]
```
and 
```
(env) MBP-di-Giuseppe:Ising1d giuscarl$ OMP_NUM_THREADS=1 python3 test.py 
# RBM Initizialized with nvisible = 100 and nhidden = 200
# Using visible bias = 1
# Using hidden bias  = 1
# Local Metropolis sampler is ready 
[22.79966077 22.81597879]
[7.83558088 7.83599558]
```

no beer, just kombucha

gcarleo:
```
 ../../Sources/Utils/log_cosh.hpp:24:44: error: function multiversioning is not supported on the current target
    __attribute__((target("default"))) Complex SumLogCosh(
```

twesterhout:
Which target is that? Could you give a bit more info about your setup, please?

Sources/Sampler/metropolis_local_v2.hpp
```diff
@@ -0,0 +1,154 @@
...
+  Index end_;
+  Index step_;
+};  // namespace netket
+
+std::tuple<RowMatrix<double>, Eigen::VectorXcd,
+           nonstd::optional<RowMatrix<Complex>>>
+ComputeSamples(MetropolisLocalV2& sampler, StepsRange const& steps,
+               bool compute_gradients);
+
+Eigen::VectorXcd LocalValuesV2(Eigen::Ref<const RowMatrix<double>> samples,
```
gcarleo: Samplers should not hold localvalues/gradients, since the form of the gradient is dependent on the specific application (say variational versus supervised, unsupervised etc)

gcarleo:
> Which target is that? Could you give a bit more info about your setup, please?

It's my macbook pro, I think the setup is very similar to the OsX setup in Travis

Sources/Machine/abstract_machine.cc
```diff
@@ -12,20 +12,106 @@
...
+  if (rank == root) {
+    default_random_engine generator{seed.has_value() ? *seed
+                                                     : ReallyRandomSeed()};
+    std::generate(parameters.data(), parameters.data() + parameters.size(),
+                  [&generator, sigma]() {
+                    std::normal_distribution<double> dist(0.0, sigma);
+                    return Complex{dist(generator), dist(generator)};
+                  });
+  }
+  auto status = MPI_Bcast(parameters.data(), parameters.size(),
```
gcarleo: Use MPI helpers here

Sources/Machine/ffnn.hpp
```diff
@@ -135,13 +135,18 @@ class FFNN : public AbstractMachine {
     }
   }
 
-  void InitRandomPars(int seed, double sigma) override {
+  void InitRandomPars(double sigma, nonstd::optional<unsigned> seed) override {
+    if (!seed.has_value()) {
+      std::random_device rd;
```
gcarleo: very dangerous, different processors might end up with different initial parameters for the wave function, to be removed/improved...

gcarleo:
```
seed=1234
```
is a better option actually

Sources/Sampler/metropolis_local_v2.cc
```diff
@@ -0,0 +1,450 @@
...
+
+ private:
+  RbmSpinV2& machine_;
+  Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor> X_;
+  Eigen::VectorXcd Y_;
+  Eigen::VectorXcd coeff_;
+  Index i_;
+};
+
+struct Accumulator {
```
gcarleo: Let's please try a way to significantly simplify this part of the code

Sources/Sampler/metropolis_local_v2.hpp
```diff
@@ -0,0 +1,154 @@
...
+#include <nonstd/span.hpp>
+
+#include "Machine/rbm_spin_v2.hpp"
+#include "Operator/abstract_operator.hpp"
+#include "Utils/exceptions.hpp"
+#include "Utils/random_utils.hpp"
+
+namespace netket {
+
+struct Suggestion {
```
femtobit: I think this is more commonly called `Proposal` (or proposed update) in MC literature.

gcarleo:
OK it compiles now, but it's somehow slower than the previous version... strange, now I get only a factor 2 speedup

Sources/Sampler/metropolis_local_v2.hpp
gcarleo: I think that if we go through this route, we need to write a general markov chain monte carlo abstract class, and implement the other cases as derived ones. Doing this change only here has the effect of making the code too complex without achieving generality

gcarleo:
We can drop support for clang 4 if necessary, the oldest version of Xcode we are supporting is aligned to clang 5 I think anyways

Sources/Machine/abstract_machine.cc
```diff
@@ -12,20 +12,106 @@
...
+AbstractMachine::VectorType AbstractMachine::LogVal(
+    Eigen::Ref<const RealRowMatrixType> v, const any &cache) {
+  VectorType out(v.rows());
+  LogVal(v, out, cache);
+  return out;
+}
+
+void AbstractMachine::DerLog(Eigen::Ref<const RealRowMatrixType> v,
+                             Eigen::Ref<RowMatrixType> out, const any &cache) {
+  if (v.cols() != Npar()) {
```
femtobit: The error checks here are not correct. For example, `v.cols()` should probably be `== Nvisible()`, not `Npar()`.

Sources/Sampler/metropolis_local_v2.cc
```diff
@@ -0,0 +1,450 @@
...
+  if (locals.size() != gradients.rows()) {
+    std::ostringstream msg;
+    msg << "incompatible dimensions: [" << locals.size() << "] and ["
+        << gradients.rows() << ", " << gradients.cols()
+        << "]; expected [N] and [N, ?]";
+    throw InvalidInputError{msg.str()};
+  }
+  Eigen::VectorXcd force(gradients.cols());
+  Eigen::Map<VectorXcd>{force.data(), force.size()}.noalias() =
+      gradients.conjugate() * locals;
```
femtobit: This should be `gradients.adjoint() * locals`.

femtobit:
I used @twesterhout's new sampling method and RBM in the VMC driver and tested it for 20 SR iterations with the `Examples/Ising1d/ising1d.py` code (I have used 8000 samples instead of 1000).

The speed-up observed in the benchmark (at least on non-Mac systems) is still noticeable to some degree in this full SR simulation:
```
Benchmark #1: python3 Examples/Ising1d/ising1d.py (Using netket/master @ e5c9f5a79697d06)
  Time (mean ± σ):     62.951 s ±  2.938 s    [User: 66.434 s, System: 2.238 s]
  Range (min … max):   61.268 s … 69.848 s    10 runs

Benchmark #2: python3 Examples/Ising1d/ising1d.py (Using twesterhout/master @ 391fbae3d1fb40a)
  Time (mean ± σ):      8.202 s ±  0.424 s    [User: 14.986 s, System: 2.386 s]
  Range (min … max):    7.571 s …  8.584 s    10 runs
```
So in this case, the SR run appears to be about `7.7 ± 0.5` times faster.

gcarleo:
OK, it seems to me that supporting vectorized `logcosh` with AVX is just too complex for what we get out of it practice... (if we stick to pytorch, hopefully they will have highly efficient implementations of most activation functions?). Can we just drop default support for it now so that we can simplify things and move on? I think it's highly prioritary to finish this PR by this week, so that we can finally concentrate on the actual goal, i.e. pytorch-ize things!!

twesterhout:
@gcarleo this week as in today or tomorrow? I probably won't get to it until Monday.

gcarleo:
I meant next week aha

Sources/Sampler/abstract_sampler.hpp
```diff
@@ -39,7 +39,7 @@ class AbstractSampler {
   // Computes the derivative of the machine on the current visible
   // Using the lookUp tables if possible
   virtual AbstractMachine::VectorType DerLogVisible() {
```
gcarleo: Maybe this function should be just removed

Sources/Optimizer/py_stochastic_reconfiguration.cc
```diff
@@ -0,0 +1,37 @@
...
+             Eigen::Ref<const VectorXcd> grad,
+             Eigen::Ref<VectorXcd> out) { self.ComputeUpdate(Oks, grad, out); },
+          py::arg{"Oks"}.noconvert(), py::arg{"grad"}.noconvert(),
+          py::arg{"out"}.noconvert());
+}
+}  // namespace
+}  // namespace detail
+
+void AddSR(PyObject* m) {
+  detail::AddSR(py::module{py::reinterpret_borrow<py::object>(m)});
```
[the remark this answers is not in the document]
twesterhout: That's pybind11's suggested way of constructing `pybind11::object` instances from `PyObject*` when you're borrowing a reference (instead of stealing it).

Sources/Machine/rbm_spin.cc
```diff
@@ -144,18 +140,15 @@ void RbmSpin::SetParameters(VectorConstRefType pars) {
...
-
-  return (v.dot(a_) + lnthetas_.sum());
-}
-
 // Value of the logarithm of the wave-function
 // using pre-computed look-up tables for efficiency
-Complex RbmSpin::LogVal(VisibleConstType v, const LookupType &lt) {
+Complex RbmSpin::LogValSingle(VisibleConstType v, const any &lookup) {
+  if (lookup.empty()) {
+    RbmSpin::lncosh(W_.transpose() * v + b_, lnthetas_);
```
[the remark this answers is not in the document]
twesterhout: Yes, we can. Would it be okay to do it in another PR though? This one already has a lot...

Sources/Machine/rbm_spin_v2.cc
```diff
@@ -0,0 +1,321 @@
...
+}
+
+namespace detail {
+namespace {
+/// Saves Python object \par raw to a file \par filename using pickle.
+///
+/// \param raw reference to a Python object. It is borrowed, not stolen.
+///
+/// \note This is a hacky solution which should be implemented in pure Python.
+void WriteToFile(const std::string &filename, PyObject *raw) {
```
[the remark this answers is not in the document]
twesterhout: Why not? `RbmSpinV2` is currently the only class which uses this function. When we see that multiple classes need this function, we can move it to Utils.

Sources/Optimizer/py_optimizer.hpp
```diff
@@ -49,6 +50,7 @@ void AddOptimizerModule(py::module &m) {
   AddAdaMax(subm);
   AddAdaGrad(subm);
   AddAdaDelta(subm);
+  AddSR(subm.ptr());
```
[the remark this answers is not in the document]
twesterhout: Because `AddSR` receives a `PyObject*`.

Sources/Sampler/abstract_sampler.hpp
twesterhout: Does this have to happen in this PR though?

Sources/Machine/rbm_spin.cc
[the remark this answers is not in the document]
gcarleo: fine

gcarleo:
To recap, I think it is vital (so to fix a common interface on which others can work in future PR) to clean up the samplers and make them compatible/ possibly extending the existing samplers. Specifically, the same work done to integrate batched machines with the existing `AbstractMachine` infrastructure should be done also for the batched samplers. 

It seems to me that in practice this just amounts to adding a few methods to `AbstractSampler`, such as `SetBatchSize` `GetBatchSize`, and doing minor changes to the interface, such as making `Visible` return a batch of visibile instead of a single visible. Then all other methods are already in place, but with different names. Specifically, `Sweep` already exists and it is what is called `Next` in the proposal, `Reset` already exists etc. 
Notice that it is straightforward to make the other samplers compatible, it just suffices to provide a default implementation of `SetBatchSize` that sets it by default to 1.

Sources/Sampler/vmc_sampling.cc
```diff
@@ -179,6 +180,307 @@ VectorXcd GradientOfVariance(const Result &result, AbstractMachine &psi,
...
+
+  RowMatrix<double> samples(num_samples, sampler.SystemSize());
+  Eigen::VectorXcd values(num_samples);
+  auto gradients =
+      compute_gradients
+          ? nonstd::optional<RowMatrix<Complex>>{nonstd::in_place, num_samples,
+                                                 sampler.Machine().Npar()}
+          : nonstd::nullopt;
+
+  struct Skip {
```
gcarleo: Is it really necessary to define specialized structs for very simple operations?

Sources/Sampler/vmc_sampling.cc
```diff
@@ -179,6 +180,307 @@ VectorXcd GradientOfVariance(const Result &result, AbstractMachine &psi,
...
+  VectorXcd mean = gradients.colwise().mean();
+  assert(mean.size() == gradients.cols());
+  MeanOnNodes<>(mean);
+  gradients.rowwise() -= mean.transpose();
+}
+}  // namespace detail
+
+std::tuple<RowMatrix<double>, Eigen::VectorXcd,
+           nonstd::optional<RowMatrix<Complex>>>
+ComputeSamples(MetropolisLocalV2& sampler, StepsRange const& steps,
```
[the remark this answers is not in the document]
gcarleo: Also, the way `ComputeSamples` is currently written seems to me much more transparent and easy to maintain and understand for newcomers (disclaimer, I didn't write it!). 
In the new version there are many `struct` introduced for very simple operations that effectively happen in a 4-lines loop...  this doesn't seem necessary/beneficial and introduces effectively a very different coding style that is adopted nowhere else in the code!

Sources/Sampler/vmc_sampling.cc
```diff
@@ -179,6 +180,307 @@ VectorXcd GradientOfVariance(const Result &result, AbstractMachine &psi,
...
+  }
+
+  void operator()(Complex log_val) {
+    assert(!forward_.Full());
+    states_.emplace_back(0, log_val);
+  }
+
+  /// Number of visible configurations we've processed is not necessarily a
+  /// multiple of batch size.
+  void Finalize(Eigen::Ref<const Eigen::VectorXd> v) {
```
gcarleo: The logic of this can be greatly simplified just imposing that the number of samples must be a multiple of the batch size. Worst-case-scenario, we print a warning saying that the number of samples has been rounded to the nearest multiple of batch size

Sources/GroundState/py_variational_montecarlo.hpp
```diff
@@ -185,8 +200,36 @@ void AddVariationalMonteCarloModule(py::module &m) {
...
+           Returns:
+               A tuple `(samples, values, gradients)` if `compute_logderivs == True`
+               and a pair `(samples, values)` otherwise. `samples` are visible
+               configurations visited during sampling and `values` are the
+               corresponding values of the logarithm of the wavefunction.
             )EOF");
+#endif
+
+  m_vmc.def("compute_samples", &ComputeSamples, py::arg{"sampler"},
+            py::arg{"nsamples"}, py::arg{"ndiscard"},
```
gcarleo: n_samples, n_discard ,etc

Sources/GroundState/py_variational_montecarlo.hpp
```diff
@@ -185,8 +200,36 @@ void AddVariationalMonteCarloModule(py::module &m) {
...
+            py::arg{"compute_logderivs"} = true,
+            R"EOF(
+      Computes a sequence of visible configurations based on Monte Carlo sampling using `sampler`.
+
+      Args:
+          sampler: sampler to use for Monte Carlo sweeps.
+          nsamples: number of samples to record (returned number of samples
```
gcarleo: may differ how?

Sources/GroundState/variational_montecarlo.hpp
```diff
@@ -154,8 +158,10 @@ class VariationalMonteCarlo {
    */
   void ComputeObservables() {
     for (std::size_t i = 0; i < obs_.size(); ++i) {
-      auto ex = vmc::Expectation(vmc_data_, psi_, *obs_[i]);
-      observable_stats_[obsnames_[i]] = ex;
+      auto local_values = LocalValues(mc_data_.samples, mc_data_.log_values,
+                                      psi_, *obs_[i], sampler_.BatchSize());
+      auto stats = Statistics(local_values, sampler_.BatchSize());
```
gcarleo: why do we need to pass explicitly the batchsize to Statistics?

Sources/Sampler/vmc_sampling.cc
```diff
@@ -15,15 +16,19 @@ Result ComputeSamples(AbstractSampler& sampler, Index nsamples, Index ndiscard,
...
     log_derivs.emplace(npar, nsamples);
     for (Index i = 0; i < nsamples; i++) {
       sampler.Sweep();
-      samples.col(i) = sampler.Visible();
-      log_derivs->col(i) = sampler.DerLogVisible();
+      samples.col(i) = VisibleLegacy(sampler);
+      log_derivs->col(i) = sampler.GetMachine().DerLogSingle(samples.col(i));
+      log_vals(i) = sampler.GetMachine().LogValSingle(samples.col(i));
     }
+    // std::ofstream out_file_3{"gradients_2.txt"};
```
gcarleo: clean up

Sources/Sampler/vmc_sampling.cc
```diff
@@ -87,6 +99,142 @@ VectorXcd LocalValueDeriv(const AbstractOperator& op, AbstractMachine& psi,
   return grad;
 }
 
+namespace detail {
```
gcarleo: should be moved in Stats

Sources/GroundState/py_variational_montecarlo.hpp
twesterhout: Yes. Returned number of samples will be `nsamples` rounded up to a multiple of `BatchSize()`.

Sources/GroundState/variational_montecarlo.hpp
twesterhout: Yes, because `Statistics` need to know how many Markov Chains there are.

gcarleo: but mc_data_.samples and mc_data.log_values shouldn't already contain this information?

Sources/GroundState/py_variational_montecarlo.hpp
```diff
@@ -34,7 +34,7 @@ void AddVariationalMonteCarloModule(py::module &m) {
 
   py::class_<Stats>(m_vmc, "Stats")
       .def_readonly("mean", &Stats::mean)
-      .def_readonly("error_on_mean", &Stats::error)
+      .def_readonly("error_of_mean", &Stats::error)
```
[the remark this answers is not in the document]
twesterhout: It will not be backward compatible even if I add `sigma` property.

Sources/GroundState/variational_montecarlo.hpp
twesterhout: No, they don't, but I can add `num_chains` member to `MCResult`.

Sources/GroundState/py_variational_montecarlo.hpp
gcarleo: Can this be a bit dangerous when storing the samples in a pre-allocated vector?

Sources/Sampler/abstract_sampler.hpp
```diff
@@ -17,30 +17,25 @@
...
-  virtual void SetVisible(const Eigen::VectorXd& v) = 0;
-
-  virtual Eigen::VectorXd Acceptance() const = 0;
-
-  // Computes the derivative of the machine on the current visible
-  // Using the lookUp tables if possible
-  virtual AbstractMachine::VectorType DerLogVisible() {
-    return GetMachine().DerLog(Visible());
-  }
+  virtual void SetVisible(Eigen::Ref<const RowMatrix<double>> v) = 0;
```
gcarleo: Why remove `Visible`? we can keep it I guess...

Sources/GroundState/py_variational_montecarlo.hpp
```diff
@@ -29,9 +29,84 @@ namespace py = pybind11;
...
+}  // namespace detail
+
 void AddVariationalMonteCarloModule(py::module &m) {
   auto m_vmc = m.def_submodule("variational");
 
+  py::class_<Stats>(m_vmc, "Stats")
+      .def_readonly("mean", &Stats::mean)
+      .def_readonly("error_of_mean", &Stats::error_of_mean)
+      .def_readonly("variance", &Stats::variance)
+      .def_readonly("autocorrelation", &Stats::correlation)
```
femtobit: Calling this `autocorrelation` (or `correlation` in C++) is somewhat misleading, because this quantity is actually (an estimate of) the _autocorrelation time_ of the Markov chains. Even though this is a bit long, I'd recommend using this name, otherwise it is just confusing (since the real autocorrelation is a different quantity of different dimensions).

If we can write `error_of_mean`, then we can also write `autocorrelation_time`.

Sources/GroundState/py_variational_montecarlo.hpp
```diff
@@ -29,9 +29,84 @@ namespace py = pybind11;
...
+          py::keep_alive<1, 0>{},
+          R"EOF(A matrix of logarithmic derivatives.
+
+                Each row in the matrix corresponds to the gradient of
+                `Ψ(vᵢ)` with respect to variational parameters.)EOF")
+      .def_readonly("n_chains", &MCResult::n_chains,
+                    R"EOF(Number of Markov Chains which this object represents.
+
+                    If `n_chains > 1`, then the first visible configuration
+                    comes from the first Markov Chain, the second -- from the
```
femtobit: Why the `--`?

Sources/GroundState/py_variational_montecarlo.hpp
```diff
@@ -29,9 +29,84 @@ namespace py = pybind11;
...
+                    second Markov chain, etc. The `n_chains + 1`st visible
+                    configuration is again from the first Markov Chains.
```
femtobit: `Chain` (singular) not `Chains`, right?

Sources/GroundState/py_variational_montecarlo.hpp
```diff
@@ -29,9 +29,84 @@ namespace py = pybind11;
...
+                    Wavefunction logarithms and derivatives are interleaved in a
+                    similar fashion.
```
femtobit: `similar` or `the same`? Writing _similar_ is potentially confusing as it does not make clear, what is the same and what is different.

Sources/GroundState/py_variational_montecarlo.hpp
```diff
@@ -256,189 +331,84 @@ void AddVariationalMonteCarloModule(py::module &m) {
...
+            py::arg{"n_samples"}, py::arg{"n_discard"},
+            py::arg{"der_logs"} = py::none(),
+            R"EOF(Runs Monte Carlo sampling using `sampler`.
 
-  m_vmc.def("compute_samples", vmc::ComputeSamples, py::arg("sampler"),
-            py::arg("nsamples"), py::arg("ndiscard") = 0,
-            py::arg("compute_logderivs") = true, R"EOF(
-           Computes a sequence of visible configurations based on Monte Carlo sampling using `sampler`.
+                  First `n_discard` sweeps are discarded. Results of the next
+                  `≈n_samples` sweeps are saved. Since samplers work with
```
femtobit: Maybe write `>=` or `⪆` or similar. `≈` looks like the number of samples could be smaller as well.

Sources/GroundState/py_variational_montecarlo.hpp
gcarleo: this was called `tau_corr` before... just saying

Sources/GroundState/py_variational_montecarlo.hpp
gcarleo: good point

Sources/GroundState/py_variational_montecarlo.hpp
```diff
@@ -29,9 +29,84 @@ namespace py = pybind11;
...
+                      py::none()});
+            }
+            return py::none();
+          },
+          py::keep_alive<1, 0>{},
+          R"EOF(A matrix of logarithmic derivatives.
+
+                Each row in the matrix corresponds to the gradient of
+                `Ψ(vᵢ)` with respect to variational parameters.)EOF")
+      .def_readonly("n_chains", &MCResult::n_chains,
```
femtobit: Why is there `MCResult.n_chains` but not `n_samples`? Since as far as I understand `result.samples.shape == n_samples, n_chains, _`, we should probably expose both or none of them.

Sources/GroundState/py_variational_montecarlo.hpp
```diff
@@ -41,19 +50,52 @@ void AddVariationalMonteCarloModule(py::module &m) {
...
+          "samples",
+          [](const MCResult &self) {
+            assert(self.samples.rows() % self.n_chains == 0);
+            return detail::as_readonly(py::array_t<double, py::array::c_style>{
+                {self.samples.rows() / self.n_chains, self.n_chains,
+                 self.samples.cols()},
+                self.samples.data(),
+                py::none()});
+          },
+          py::keep_alive<1, 0>{},
```
gcarleo: py::keep_alive<0,1>

Sources/GroundState/py_variational_montecarlo.hpp
```diff
@@ -29,12 +29,75 @@ namespace py = pybind11;
...
+  py::detail::array_proxy(array.ptr())->flags &=
+      ~py::detail::npy_api::NPY_ARRAY_WRITEABLE_;
+  return array;
+}
+}  // namespace detail
+
 void AddVariationalMonteCarloModule(py::module &m) {
   auto m_vmc = m.def_submodule("variational");
 
+  py::class_<Stats>(m_vmc, "Stats")
```
gcarleo: but now I am confused, why are these in variational.Stats ?

Sources/GroundState/py_variational_montecarlo.hpp
```diff
@@ -256,189 +319,91 @@ void AddVariationalMonteCarloModule(py::module &m) {
...
-            )EOF");
+                  Returns:
+                      A `MCResult` object with all the data obtained during sampling.)EOF");
 
   m_vmc.def(
-      "local_values",
-      [](const vmc::Result &result, AbstractMachine &psi,
-         const AbstractOperator &op) {
-        return vmc::LocalValues(op, psi, result.SampleMatrix());
+      "gradient_of_expectation",
```
gcarleo: I am still thinking how to structure these helper functions better... I mean, in python I'd expect just two functions:
```python 
s_results=compute_samples(...)
grad=gradient_of_expectation(my_hamiltonian, s_results...) 
```

I mean, intuitively I don't like the idea to specifically call `local_values` first, and basically enter into (unecessary) details about how the gradient of expectation is internally computed (i.e. first forming the local values)

PR #256, opened 2019-07-19
Refactor SR and add more options for least-squares solvers

femtobit:
This PR refactors the code in the `SR` class by introducing generic methods for constructing the S matrix and solving the least-squares update equation which work for both the real and complex case.
This makes the code easier to read and allows changes to regularization and solvers in one place in the code.

Further, this PR makes available more of the [solvers available in Eigen](https://eigen.tuxfamily.org/dox/group__LeastSquares.html) and adds an option to store the rank of the S matrix and even the full S matrix (which might be useful for debugging purposes).


Sources/Optimizer/stochastic_reconfiguration.cc
```diff
@@ -0,0 +1,96 @@
...
+  if (use_iterative_ && enabled) {
+    throw std::logic_error{
+        "Cannot store full S matrix with `use_iterative = true`."};
+  }
+  store_full_S_matrix_ = enabled;
+  if (!enabled) {
+    last_S_ = nonstd::nullopt;
+  }
+}
+}  // namespace netket
```
gcarleo: no new line at end of file?

Sources/Supervised/supervised.hpp
```diff
@@ -461,9 +460,12 @@ class Supervised {
 
   void setSrParameters(double diag_shift = 0.01, bool use_iterative = false,
```
gcarleo: can we make this and unsupervised also accept more general solvers for Sr?

Docs/variational/Vmc.md
```diff
@@ -5,19 +5,19 @@ Variational Monte Carlo schemes to learn the ground state using stochastic recon
...
+|use_iterative            |bool = False             |Whether to use the iterative solver in the Sr method (this is extremely useful when the number of parameters to optimize is very large). The default is false.                                                                                                                                                                                                                                                                                                |
+|sr_lsq_solver            |str = 'BDCSVD'           |The solver used to solve the least-squares equation in the SR update. Only used if `method == "SR" and not use_iterative`. Available options are "BDCSVD", "ColPivHouseholder", "LDLT", and "LLT". See the [Eigen documentation](https://eigen.tuxfamily.org/dox/group__TutorialLinearAlgebra.html) for a description of the available solvers. Also note that LLT and LDLT only work for positive definite or positive semidefinite S matrices, respectively.|
```
gcarleo: maybe worth setting the default to cholesky?

gcarleo: Yes, I'd set `LLT` the default value, it is much faster and it should also be relatively safe if there is a finite `diag_shift`...

PR #257, opened 2019-07-22
Refactor SR and add more options for least-squares solvers

femtobit:
This PR refactors the code in the `SR` class by introducing generic methods for constructing the S matrix and solving the least-squares update equation which work for both the real and complex case.
This makes the code easier to read and allows changes to regularization and solvers in one place in the code.

Further, this PR makes available more of the [solvers available in Eigen](https://eigen.tuxfamily.org/dox/group__LeastSquares.html) and adds an option to store the rank of the S matrix and even the full S matrix (which might be useful for debugging purposes).


femtobit:
(Created this new PR because GitHub for some reason could not complete the check for conflicts with the base branch, preventing a merge.)

Sources/Optimizer/stochastic_reconfiguration.hpp
```diff
@@ -19,143 +19,255 @@
...
+  void SetStoreFullSMatrix(bool enabled);
+
+ private:
+  LSQSolver solver_;
   double sr_diag_shift_;
   bool use_iterative_;
-  bool use_cholesky_;
   bool is_holomorphic_;
 
+  bool scale_invariant_pc_ = false;
```
gcarleo: Can we make this active by default and deactivate it automatically (for the moment) if the sparse solver is used?

femtobit: Sure.

Issue #261, opened 2019-07-27
Backport 2.1 API changes to stable 2.X series 


gcarleo:
Version 2.1 is introducing batching (among other things), and this results in a (small) but necessary change in the APIs of machines and samplers. Since we are still in beta for version 2.0, we should backport these API changes to 2.0 and do a release candidate for 2.0. 

In practice, API changes in 2.0 should happen only at the level of the python bindings, and hopefully never touch the C++ code. I can try to do that soon.

PR #262, opened 2019-07-27
Backport 2.1 API changes in AbstractMachine

gcarleo:
Starts addressing #261, for the moment only for Machines, soon also for Samplers


gcarleo:
An alternative to backporting is to make `LogVal` and `DerLog` in 2.1 accept both 1D and 2D arrays, as implemented here. The issue is understanding if the internal conversions are fast enough, I guess @femtobit @twesterhout

gcarleo:
Closing this now, since 2.1 is "almost" compatible, and 2.0 will never be released beyond beta stage

PR #266, opened 2019-07-31
 Add convenience function for VMC estimation

femtobit:
This PR adds a convenience function for VMC sampling, which only requires callers to provide operator, machine, and MCResult.

For consistency in the interface, I have also changed the order of arguments for `operator.local_values`, so that is more closely matches the previous API and the new convenience function.


femtobit:
The changes in this PR have now been incorporated into #276, therefore I'm closing this one.

PR #267, opened 2019-07-31
Add Edgeless to Graph library and improve Pauli Operator (now PauliStrings)

gcarleo:


Sources/Graph/set.cc
```diff
@@ -0,0 +1,56 @@
...
+            std::end(automorphisms_.front()), 0);
+}
+
+int Set::Nsites() const noexcept { return n_sites_; }
+
+int Set::Size() const noexcept { return n_sites_; }
+
+bool Set::IsConnected() const noexcept { return false; }
+
+bool Set::IsBipartite() const noexcept { return false; }
```
femtobit: Strictly speaking, the edgeless graph is bipartite (it can be 2-colored -- in fact, it can even be 1-colored).

Sources/Graph/set.cc
```diff
@@ -0,0 +1,56 @@
...
+
+#include "set.hpp"
+
+#include <algorithm>
+#include <numeric>
+#include <sstream>
+
+namespace netket {
+
+Set::Set(int size) : n_sites_(size) {
```
femtobit: The name `Set` could be somewhat misleading. In graph theory, this structure is is sometimes called the _edgeless_ graph. `netket.graph.Edgeless` might sound a bit strange, but is it pretty unambiguous.

Sources/Operator/pauli_strings.hpp
```diff
@@ -46,12 +45,12 @@ class PauliOperator : public AbstractOperator {
...
   using VectorConstRefType = AbstractOperator::VectorConstRefType;
 
-  explicit PauliOperator(std::shared_ptr<const AbstractHilbert> hilbert,
-                         const std::vector<std::string> &ops,
-                         const std::vector<std::complex<double>> &opweights,
-                         double cutoff = 1e-10)
-      : AbstractOperator(hilbert),
-        nqubits_(hilbert->Size()),
+  explicit PauliStrings(const std::vector<std::string> &ops,
+                        const std::vector<std::complex<double>> &opweights,
```
femtobit: We can use the `Complex` type alias now.

gcarleo:
Alright, it should be good now!

PR #272, opened 2019-08-01
 Use shared_ptr for Operator and update plotting scripts and speed up PyRbm

gcarleo:


gcarleo:
`PyrRbm` is now much faster than the previous implementation, but still it's some 2x slower than `RbmSpinV2` on 20 sites and about 25% slower (only) than `RbmSpin`. The speed difference is most likely due to the gradient computation, that seems slower than the C++ implementations. Maybe it can be further improved only using numpy primitives/ better using views

gcarleo:
Can we merge this?

PR #273, opened 2019-08-01
Report durations for all tests cases when running tests on Travis

femtobit:
Calling [`pytest --durations=0`](http://doc.pytest.org/en/latest/usage.html#profiling-test-execution-duration) will print the runtime of all test cases, which I think is nice to have (primarily to identify which tests are slow).


gcarleo:
The problem however is that compilation/installation times take 2/3 of the total testing time (at least), not too much that some tests are long

femtobit:
True. Still interesting to see, how long the test cases take.

PR #274, opened 2019-08-02
WIP: A wrapper for jax.experimental.stax neural networks

twesterhout:
See `test_machine.py` for an example.

Still need to clean up the code a bit, but the PR is ready for feedback.


Test/Machine/test_machine.py
```diff
@@ -22,6 +27,30 @@ def merge_dicts(x, y):
...
+    return x
+  return init
+
+machines["JAX Machine"] = JAXMachine(hi,
+    jax.experimental.stax.serial(
+        jax.experimental.stax.Dense(16, glorot(), randn()),
+        jax.experimental.stax.Relu,
+        jax.experimental.stax.Dense(16, glorot(), randn()),
+        jax.experimental.stax.Relu,
+        jax.experimental.stax.Dense(2, glorot(), randn()),
```
gcarleo: we should check that the output has size 2 in the non holomorphic case

Examples/CustomMachine/jax_wrapper.py
```diff
@@ -0,0 +1,136 @@
...
+        # NOTE: The following call to __init__ is important!
+        super(JAXMachine, self).__init__(hilbert)
+        if seed is None:
+            # NOTE(twesterhout): I believe, jax uses 32-bit integers internally
+            # to represent the seed. Hence the limit
+            seed = random.randint(0, 2 ** 32 - 1)
+        init_fn, self._forward_fn = module
+        # Chosen randomly. It'll change if we forward propagate `v` with a
+        # different leading dimension.
+        batch_size = 64
```
gcarleo: randomly = 64

Examples/CustomMachine/jax_wrapper.py
```diff
@@ -0,0 +1,136 @@
...
+import random
+import numpy as _np
+import jax as _jax
+
+import netket
+
+__all__ = ["JAXMachine"]
+
+
+class JAXMachine(netket.machine.CxxMachine):
```
gcarleo: can we check that double precision is being actually used?

Test/Machine/test_machine.py
gcarleo: also worth checking what happens when initializing layers at random on multiple MPI cores...

Examples/CustomMachine/jax_wrapper.py
gcarleo: unfortunately the JAX machine is significantly slower (on my machine) than the bare numpy one (at least a factor of 10 or more)... we should look into this and see why this happens

Examples/CustomMachine/jax_wrapper.py
```diff
@@ -0,0 +1,168 @@
...
+    def _log_val(self, x, out):
+        """
+        Do not use this function directly! It's an implementation detail!
+
+        See `self.log_val`.
+        """
+        out[:] = (
+            np.asarray(self._forward_fn(self._params, x))
+            .view(dtype=self._complex_dtype)
+            .squeeze()
```
gcarleo: why do you need to squeeze this?

Examples/CustomMachine/jax_wrapper.py
```diff
@@ -0,0 +1,168 @@
...
+    def _der_log(self, x, out):
+        """
+        Do not use this function directly! It's an implementation detail!
+
+        See `self.log_val`.
+        """
+        J = self._jacobian(self._params, x)
+        batch_size = x.shape[0]
+        i = 0
+        for g in (g.reshape(batch_size, 2, -1) for layer in J for g in layer):
```
gcarleo: I don't understand what `this for in for in for in` loop does, maybe it's me... ahha

Test/Machine/jax_wrapper.py
```diff
@@ -0,0 +1 @@
+../../Examples/CustomMachine/jax_wrapper.py
```
gcarleo: maybe better to put instead a symbolic link to jax_wrapper here?

Examples/CustomMachine/jax_wrapper.py
```diff
@@ -0,0 +1,168 @@
+# Copyright 2019 The Simons Foundation, Inc. - All Rights Reserved.
```
gcarleo: it would be nice to be able to load this as a part of the netket module... 
Something like `netket.machine.Jax()`

setup.py
```diff
@@ -158,7 +158,7 @@ def _decode(x):
     url="http://github.com/netket/netket",
     author_email="netket@netket.org",
     license="Apache 2.0",
-    packages=["netket"],
+    packages=["netket", "netket.machine"],
```
gcarleo: why do we need netket.machine ?

PR #276, opened 2019-08-05
VMC driver and convenience functions in Python

femtobit:
This is a work-in-progress demonstration of how the VMC driver can be implemented in Python now.

The example driver added largely has the same functionality as the current C++ one. In addition to the NetKet optimizers, it is possible to pass optimizers from `jax.experimental.optimizers` or any lambda function `update` computing `p[i+1] = update(i, grad, p[i])`.


Examples/PyNetKet/ground_state_iter.py
```diff
@@ -17,50 +18,48 @@
...
+                energy.real,
                 sigma,
                 variance,
-                vsigma,
-                sa.acceptance[0],
-                vmc.last_rank,
-                w.min(),
-                w.max(),
+                rhat,
+                1,  # sa.acceptance,
```
gcarleo: MetropolisLocalV2 doesn't have acceptance?

netket/_driver.py
```diff
@@ -0,0 +1,265 @@
...
+        """
+        Performs a number of VMC optimization steps.
+
+        Args:
+            n_steps (int): Number of steps to perform.
+        """
+
+        def update_samples():
+            self._mc_data = _vmc.compute_samples(
+                self._sampler, self.n_samples, self.n_discard, der_logs="centered"
```
gcarleo: Yes, I really think we should remove der_logs from here! It would be much less error prone etc

netket/_driver.py
```diff
@@ -0,0 +1,265 @@
...
+
+        if not self._mc_data:
+            update_samples()
+
+        for _ in range(n_steps):
+            # Estimate energy
+            eloc, self._stats = self._get_mc_stats(self._ham)
+
+            # Estimate energy gradient
+            grad = _vmc.gradient_of_expectation(eloc, self._mc_data.der_logs)
```
gcarleo: I think the easiest is just to combine (if possible) these two functions in something that computes both expectations and gradients of expectations 

```python 
eloc, grad = _vmc.gradient_of_expectation(self._ham, self._mc_data) 
```

Examples/PyNetKet/ground_state_iter.py
femtobit: Right now it doesn't; that still needs to be implemented. (@twesterhout)

netket/_driver.py
femtobit: If having two separate calls does not impact performance, that would be the best solution, IMHO. Then one could call
```python
samples, logpsi = _vmc.compute_samples(...) # wrapping samples and logpsi into one object should not be needed anymore, so just return a tuple
logder = _vmc.log_derivatives(samples, centered=True)

eloc = _vmc.local_values(samples, logpsi, self.ham, self.machine)
grad_E = _vmc.gradient_of_expectation(eloc, logder)
```

Examples/PyNetKet/ground_state_iter.py
femtobit: Ok, see #278.

Sources/Optimizer/py_stochastic_reconfiguration.cc
```diff
@@ -3,31 +3,59 @@
...
+             const auto solver = SR::SolverFromString(solver_name);
+             NETKET_CHECK(solver.has_value(), InvalidInputError,
+                          "Invalid LSQ solver \"" << solver_name
+                                                  << "\" specified for SR");
+             return SR(solver.value(), diag_shift, use_iterative,
+                       is_holomorphic);
+           }),
+           py::arg("lsq_solver") = "LLT", py::arg("diag_shift") = 0.01,
+           py::arg("use_iterative") = false, py::arg("is_holomorphic") = true)
+      .def("compute_update", &SR::ComputeUpdate, py::arg("Oks").noconvert(),
```
twesterhout: "Oks" -> "der_logs" or whatever the rest of the codebase uses.

Sources/Optimizer/py_stochastic_reconfiguration.cc
```diff
@@ -3,31 +3,59 @@
...
 
-void AddSR(PyObject* m) {
-  detail::AddSR(py::module{py::reinterpret_borrow<py::object>(m)});
+            Args:
+                Oks: The matrix 𝕆 of centered log-derivatives,
+                   𝕆_ij = O_i(v_j) - ⟨O_i⟩.
+                grad: The vector of forces f.
+                out: Output array for the update ẋ.
+          )EOF")
+      .def_property("store_rank_enabled", &SR::StoreRankEnabled,
```
twesterhout: Why not simply `store_rank`?

Sources/Optimizer/stochastic_reconfiguration.cc
```diff
@@ -68,20 +68,34 @@ void SR::SetParameters(double diagshift, bool use_iterative, bool use_cholesky,
                 use_iterative, is_holomorphic);
 }
 
-std::string SR::GetInfoString() {
+std::string SR::LongDesc(Index depth) const {
+  auto indent = [&depth]() { return std::string(4 * depth, ' '); };
   std::stringstream str;
```
twesterhout: Should probably be `std::ostringstream`.

Sources/Optimizer/stochastic_reconfiguration.hpp
```diff
@@ -55,7 +55,7 @@ class SR {
         sr_diag_shift_(diagshift),
         use_iterative_(use_iterative),
         is_holomorphic_(is_holomorphic) {
-    InfoMessage() << GetInfoString();
+    InfoMessage() << LongDesc();
```
twesterhout: Can we, please, not call `InfoMessage` here since it will be printed every time someone constructs an SR object in Python. If the user wants to see the description, they can call `info`.

netket/_driver.py
```diff
@@ -0,0 +1,265 @@
...
+        desc = "{}{}".format(arg.__name__, sig)
+        return arg, desc
+    else:
+        raise ValueError(
+            "Expected netket.optimizer.Optimizer subclass, JAX optimizer, "
+            + " or callable f(i, grad, p); got {}".format(arg)
+        )
+
+
+class VmcDriver:
```
twesterhout: You should probably derive from `object` in Python 2.

Sources/Optimizer/py_stochastic_reconfiguration.cc
femtobit: `store_rank` is more of a function name than a property to me. The property is _whether storing the rank is enabled_.

Sources/Optimizer/stochastic_reconfiguration.hpp
femtobit: Agreed.

Maybe be should discuss this part a bit more:

What I did in thsi PR is to use `__repr__` to get a short representation of the object (useful, e.g., when testing things interactively in a shell or notebook).
In some cases, it can also be nice to have longer output describing an object (e.g., showing the Hamiltonian, Machine, SR, etc. for the driver). We should proably have a standard function to do that, which we can use for several classes. I would suggest calling it `.info()`, so that's what I did here.

What do you think?

netket/_driver.py
```diff
@@ -0,0 +1,265 @@
...
+        Args:
+            hamiltonian: The Hamiltonian of the system.
+            machine: The variational wavefunction.
+            sampler: The Monte Carlo sampler.
+            optimizer: Determines how optimization steps are performed given the
+                bare energy gradient. This parameter supports three different kinds of inputs,
+                which are described in the docs of `make_optimizer_fn`.
+            n_samples: Number of Markov Chain Monte Carlo sweeps to be
+                performed at each step of the optimization.
+            discarded_samples (int): Number of sweeps to be discarded at the
```
femtobit: Docstring still needs to be updated.

femtobit:
Some things that are left to do either in this PR or later:

- [ ] Update all docstrings
- [ ] Decide where we want to put the new driver class in the module hierarchy and move it there from its current location in `_driver`
- [ ] Move computation of `der_log` into a separate function from `compute_samples`

Sources/GroundState/py_variational_montecarlo.cc
```diff
@@ -0,0 +1,309 @@
...
+
+namespace detail {
+template <class T, int ExtraFlags>
+py::array_t<T, ExtraFlags> as_readonly(py::array_t<T, ExtraFlags> array) {
+  py::detail::array_proxy(array.ptr())->flags &=
+      ~py::detail::npy_api::NPY_ARRAY_WRITEABLE_;
+  return array;
+}
+
+auto TestCreateArray() -> py::array_t<Complex> {
```
gcarleo: To be removed?

Sources/GroundState/py_variational_montecarlo.cc
```diff
@@ -0,0 +1,309 @@
...
+            return detail::as_readonly(py::array_t<Complex, py::array::c_style>{
+                {self.log_values.rows() / self.n_chains, self.n_chains},
+                self.log_values.data(),
+                py::none()});
+          },
+          py::return_value_policy::reference_internal,
+          R"EOF(An array of `complex128` representing `Ψ(vᵢ)` for all
+                sampled visible configurations `vᵢ`.)EOF")
+      .def_property_readonly(
+          "der_logs",
```
gcarleo: der_logs no longer needed, just a reminder here

Sources/Stats/py_stats.cc
```diff
@@ -127,6 +129,77 @@ void AddStatsModule(py::module m) {
                     Rank-2 tensors should have shape `(N, M)` where `N` is the
                     number of samples in one Markov Chain and `M` is the number
                     of Markov Chains. Data should be in row major order.)EOF");
+
+  subm.def(
+      "covariance_sv",
```
gcarleo: I guess this function would be fast enough if written in pure python with numpy?

gcarleo:
Instead of completely removing `der_logs`, an alternative (and elegant, I guess) solution would be to just pass a callback-like object, say 

```python
samples, results = vmc.compute_samples(sampler,...., netket.log_der_callback(machine)) 
```
and results contains `log_values, der_logs` 

or one could instead want only log_values 

```python
samples, log_values = vmc.compute_samples(sampler,...., netket.log_callback(machine)) 
```

or nothing at all 
```python
samples = vmc.compute_samples(sampler,....) 
```
or even compute local energies on the fly 
```python
samples, e_locs= vmc.compute_samples(sampler,...., netket.local_ops_callback(machine,hamiltonian)) 
```

@femtobit @twesterhout what do you think?

Sources/GroundState/py_variational_montecarlo.cc
```diff
@@ -0,0 +1,309 @@
...
+          },
+          R"EOF(
+        Calculate and return the value of the operators stored as observables.
+
+        )EOF")
+      .def_property_readonly("vmc_data", &VariationalMonteCarlo::GetVmcData)
+      .def_property_readonly("sr", &VariationalMonteCarlo::GetSR);
+
+  m_vmc.def(
+      "compute_samples",
```
gcarleo: compute_samples should be in `netket.sampler.compute_samples` as the corresponding C++ function!

gcarleo:
Here is a minimal code prototype, let me know if this makes sense 

```python
samples, der_logs, e_locs= compute_samples(sampler,....,netket.vmc_callback(machine,hamiltonian))

grad=covariance(der_logs,e_locs) 

# use grad to perform updates... 

```

I think we really need two functions here: `compute_samples` and `covariance`, nothing more that that ?

Sources/GroundState/py_variational_montecarlo.cc
femtobit: done

Sources/Stats/py_stats.cc
femtobit: Since it performs an MPI reduction, I think it's better keep the C++ at the moment.

femtobit:
Alright, I removed the log-derivative computation from the `compute_samples` interface.
Aside from the MPI reduction and the centering, `machine.der_log` can do that part (with the batched interface, one can even pass in all the data from `compute_samples` at the same time).

Now we can write
```python
samples, logvals = compute_samples(sampler, 10000, 1000)
derlogs = log_derivatives(machine, samples, centered=True)
```

(It's still there in the C++ code because the C++ VMC driver class uses it and removing it there would take some time.)

Sources/Sampler/py_sampler.hpp
```diff
@@ -152,6 +161,94 @@ void AddSamplerModule(py::module& m) {
...
+                {self.log_values.rows() / self.n_chains, self.n_chains},
+                self.log_values.data(),
+                py::none()});
+          },
+          py::return_value_policy::reference_internal,
+          R"EOF(An array of `complex128` representing `Ψ(vᵢ)` for all
+                sampled visible configurations `vᵢ`.)EOF");
+
+  subm.def(
+      "compute_samples",
```
gcarleo: I think that ```compute_samples``` should just return a rank 2 tensor of visible, and be an easy and plain method defined in ```AbstractSampler``` so that one just does the easy and intuitive thing

```python
samples=sa.compute_samples(n_samples,n_discard)
```

Sources/Sampler/py_sampler.hpp
```diff
@@ -152,6 +161,94 @@ void AddSamplerModule(py::module& m) {
   AddExactSampler(subm);
   AddCustomSampler(subm);
   AddMetropolisHastings(subm);
+
+
+  py::class_<MCResult>(subm, "MCResult",
```
gcarleo: this ```MCResult``` type also seems redundant, see my previous reasoning

netket/_driver.py
```diff
@@ -0,0 +1,263 @@
...
+            if self._sr:
+                dp = _np.empty(self._npar, dtype=_np.complex128)
+                shp = self._derlogs.shape
+                der_logs = self._derlogs.reshape(shp[0] * shp[1], shp[2])
+                self._sr.compute_update(der_logs, grad, dp)
+            else:
+                dp = grad
+
+            self._machine.parameters = self._optimizer_step(
+                self.step_count, dp, self._machine.parameters
```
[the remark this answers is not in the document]
femtobit: This signature is for compatibility with JAX optimizers. (The step count is used, e.g., for decaying step size there.)

netket/variational.py
```diff
@@ -26,3 +27,78 @@ def _Vmc_iter(self, n_iter=None, step_size=1):
...
+    The estimate is based on a Markov chain of configurations obtained from
+    `netket.sampler.compute_samples`.
+
+    Args:
+        op: Linear operator
+        psi: Variational wavefunction
+        samples: A matrix (or a rank-3 tensor) of visible
+            configurations. If it is a matrix, each row of the matrix
+            must correspond to a visible configuration.  `samples` is a
+            rank-3 tensor, its shape should be `(N, M, #visible)` where
```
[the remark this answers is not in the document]
femtobit: Because then users would have to reshape the output of `compute_samples` before passing it to this  function.

Sources/Sampler/py_sampler.hpp
femtobit: What about the log-values?

Sources/Sampler/py_sampler.hpp
femtobit: Yeah, I think I kept it by accident. it is not actually used in Python anymore.

femtobit:
>  So in this sense I don't really see a role for log_derivatives here?
We could instead make `machine.der_log` accept the rank-3 samples object as well. (I do not think we should return "flattened" samples over all chains - then we would need to pass the number of chains separately for MC statistics and getting that wrong affects the diagnostics.)
Then we would still have to subtract the mean over MPI somewhere else.

femtobit:
Hm, only [one of the Travis configs](https://travis-ci.org/netket/netket/jobs/586600413) fails, because it seems to hang during a test (or take waay too long). I am not sure why this happens.

Sources/Sampler/py_sampler.hpp
gcarleo: Can we remove it then? :)

Sources/Sampler/py_sampler.hpp
gcarleo: they are not needed, if we want we can implement an internal caching mechanism, that is going to be even faster than just returning log_val

Sources/Sampler/py_sampler.hpp
```diff
@@ -152,6 +161,94 @@ void AddSamplerModule(py::module& m) {
...
+
+  subm.def(
+      "compute_samples",
+      [](AbstractSampler &sampler, Index n_samples, Index n_discard) {
+        // Helper types and lambda for writing the shapes in a clean way:
+        using Shape2 = std::array<std::size_t, 2>;
+        using Shape3 = std::array<std::size_t, 3>;
+        const auto _ = [](Index i) { return static_cast<std::size_t>(i); };
+
+        MCResult result =
```
gcarleo: MCResult is still used here though?

femtobit: As I said, it is not used in the Python interface. Removing it in C++ requires changing a lot of code without immediate benefit.

Sources/Sampler/py_sampler.hpp
femtobit: I wouldn't object to that. But I also do not see the need to do this here. (As long as the Python interface is reasonably stable, this can be changed under the hood later.)

Sources/Sampler/py_sampler.hpp
femtobit: That would require changing at least `nk.operator.local_values` as well (and, at least currently, recomputing all the log values there, if I am not mistaken).

netket/_driver.py
```diff
@@ -0,0 +1,263 @@
...
+            self._derlogs = _vmc.log_derivatives(
+                self._machine, self._samples, centered=True
+            )
+
+        if not self._mc_data:
+            update_samples()
+
+        for _ in range(n_steps):
+            # Estimate energy
+            eloc, self._stats = self._get_mc_stats(self._ham)
```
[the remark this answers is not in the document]
femtobit: It is also used for other observables.

netket/_driver.py
```diff
@@ -0,0 +1,263 @@
+import sys
```
gcarleo: can we call this _variational_driver.py or something to avoid confusion though?

femtobit: We probably should agree on a permanent place to put the new driver (at least until NetKet 3.0).
If we just follow the common scheme, we could just put it under `netket.variational.VmcDriver`.

netket/_driver.py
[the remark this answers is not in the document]
gcarleo: Ok, I guess we can leave it for now, and finalize the interface when we also know what we should pass to pytorch optimizers

netket/_driver.py
gcarleo: I don't recall exactly the outcome of our previous discussions, but I thought there was agreement about putting all the highest-level applications in the main namespace? something like `netket.Vmc` ? In any case, my comment was more about naming the python file something different than just `netket/_driver.py` that could cause ambiguity with other drivers

femtobit: Fair enough, I can rename it (after all, the name of that file is an implementation detail and can be changed without breaking the interface anyway) and import the driver directly in the `netket` module.

femtobit: Done. Let me know what you think of the interface now.
The new driver is `netket.Vmc`.

Also, I have changed the top-level `estimate_expectations` function, it can now be used like this:
https://github.com/netket/netket/blob/59e75baeb8476b875e2da15ca977351e5196b966/Test/GroundState/test_vmc.py#L73-L78
(It has an optional `compute_gradients` argument. Only downside is that it will compute the gradients for all observables. However, for more control it is always possible to use the lower-level functions like the driver does.)

femtobit:
There is a single Travis build that fails at the moment, but the error seems unrelated to this PR:
```
    ModuleNotFoundError: No module named 'skbuild'
```

Examples/PyNetKet/ground_state_iter.py
```diff
@@ -69,29 +68,35 @@ def output(vmc, step):
...
 
-def run_vmc(steps, step_size, diag_shift):
-    op = nk.optimizer.Sgd(step_size)
-    vmc = nk.variational.Vmc(
+def run_vmc(steps, step_size, diag_shift, n_samples):
+    opt = jaxopt.sgd(step_size)
+    # opt = nk.optimizer.Sgd(step_size)
+
+    sr = nk.optimizer.SR(lsq_solver="BDCSVD", diag_shift=diag_shift)
+    sr.store_rank_enabled = False  # not supported by BDCSVD
```
gcarleo: maybe this is not related to this PR, but why these options are not just 
```python 
sr.store_rank = True
sr.store_covariance_matrix = True
```

Sources/Sampler/py_sampler.hpp
```diff
@@ -152,6 +161,66 @@ void AddSamplerModule(py::module& m) {
...
+                      sampler: sampler to use for Monte Carlo sweeps.
+                      n_samples: number of samples to record.
+                      n_discard: number of sweeps to discard.
+                      der_logs: Whether to calculate gradients of the logarithm
+                          of the wave function. `None` means don't compute,
+                          "normal" means compute, and "centered" means compute
+                          and then center.
+
+                  Returns:
+                      A `MCResult` object with all the data obtained during sampling.)EOF");
```
gcarleo: Needs to be updated

Sources/Stats/py_stats.cc
```diff
@@ -127,6 +129,77 @@ void AddStatsModule(py::module m) {
...
+          default:
+            NETKET_CHECK(false, InvalidInputError,
+                         "s_values has wrong dimension: "
+                             << s_values.ndim() << "; expected either 1 or 2.");
+        }  // end switch
+      },
+      py::arg{"s"}.noconvert(), py::arg{"v"}.noconvert(),
+      py::arg{"center_s"} = true,
+      R"EOF(Computes the covariance of two random variables S, V from MCMC
+            data. Note that while S has to be a complex scalar variable,
```
gcarleo: would this work for real-valued S or would you get an error?

netket/variational.py
```diff
@@ -26,3 +26,58 @@ def _Vmc_iter(self, n_iter=None, step_size=1):
 
 
 Vmc.iter = _Vmc_iter
+
+
+# Higher-level VMC functions:
+
+
+def estimate_expectations(
```
gcarleo: this function is not used anymore, to be removed

Examples/PyNetKet/ground_state_iter.py
femtobit: We could rename these options, yes. Let's do that in a separate PR.

netket/variational.py
femtobit: I think it could be useful for end users. For instance, if you have a saved machine and want to compute expectation values in a separate script. As end user, you would have to call `compute_samples`, `local_values`, and `statistics` instead of just doing
```
ma.load("some_file.dat")
sampler = MetropolisLocal(ma)
stats = estimate_expectations([op1, op2], sampler, 50000)
```

Sources/Stats/py_stats.cc
femtobit: It would throw an error, like this:
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-19-586e4b786a84> in <module>
----> 1 x = nk.stats.covariance_sv(elocs.real, derlogs)

TypeError: covariance_sv(): incompatible function arguments. The following argument types are supported:
    1. (s: numpy.ndarray[complex128], v: numpy.ndarray[complex128], center_s: bool = True) -> numpy.ndarray[complex128[m, 1]]

```

gcarleo: OK, but the natural behavior would be to accept also floats and compute the standard covariance in that case... We can change this later I guess, but it would make sense

netket/variational.py
gcarleo: Ok, I agree good point

Examples/Ising1d/ising1d.py
```diff
@@ -34,13 +34,13 @@
 op = nk.optimizer.Sgd(learning_rate=0.1)
```
gcarleo: sorry, very last thing, can we revert this example?

femtobit: No problem. Should it use the old C++ driver or the new one correctly?

gcarleo: The new one I'd say... then at some point we should deprecate the old style things, but we need to support also old-style variational at least until the major release

femtobit: Agreed.

femtobit: Hm, I'd leave the example as it is (beside making it run again, of course) right now, since the driver does not have a `run` function like the old driver yet.
(Personally, I would not miss that function if stays absent from the new driver. But if necessary, we could still add it later.)

Examples/Ising1d/ising1d.py
```diff
@@ -35,12 +35,7 @@
 
 # Stochastic reconfiguration
 gs = nk.variational.Vmc(
-    hamiltonian=ha,
-    sampler=sa,
-    optimizer=op,
-    n_samples=1000,
-    diag_shift=0.1,
-    method="Sr",
+    hamiltonian=ha, sampler=sa, optimizer=op, n_samples=1000, diag_shift=0.1
```
gcarleo: method="Sr" is missing...

femtobit: Added it (its the default, though).

gcarleo:
Thanks @femtobit, we made it! I opened a few issues on the things still to be done, we can work on those later, let me know if there is some of those things you'd like to do yourself

PR #277, opened 2019-08-06
Add __repr__ and asdict() to Stats

femtobit:
Provides more useful `__repr__` output for `Stats` objects in Python and add `asdict` method:

```python
>>> stats = netket.stats.statistics(local_energy)
>>> stats
-8.321 ± 0.8524 (Im=-0.01043, var=58.23, R=1.0056)
>>> stats.asdict()
{'Mean': (-8.320763354546632-0.010434343835340771j), 'Sigma': 0.8524036308501588, 'Variance': 58.231913842857885, 'R': 1.0055981084528502, 'TauCorr': nan}
````

### DIFF
--- a/Sources/Stats/py_stats.cc
+++ b/Sources/Stats/py_stats.cc
@@ -21,6 +21,16 @@ py::dict GetItem(const ObsManager& self, const std::string& name) {
   return dict;
 }
 
+int GetPrecision(double /* value */, double error) {
+  const int loge = std::floor(std::log10(std::abs(error)));
+  return std::max(1 - loge, 0);
+}
+
+int GetPrecision(Complex value, double error) {
+  return GetPrecision(std::max(std::abs(value.real()), std::abs(value.imag())),
+                      error);
+}
+
 void AddStatsModule(py::module m) {
   auto subm = m.def_submodule("stats");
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: Sources/Stats/py_stats.cc [hunk 1]
```diff
@@ -46,12 +48,39 @@ void AddStatsModule(py::module m) {
...
       .def_readonly("variance", &Stats::variance)
       .def_readonly("tau_corr", &Stats::correlation)
-      .def_readonly("R", &Stats::R);
+      .def_readonly("R", &Stats::R)
+      .def("__repr__",
+           [](const Stats& self) {
+             std::ostringstream stream;
+             const double imag = self.mean.imag();
+             // clang-format off
+             stream << std::setprecision(4)
```
gcarleo: Can we set the number of significant digits according to the error in self.error_of_mean? (the error itself only needs 2 significant digits at most, I would say)

femtobit: Alright, let's try and do it properly then.

femtobit:
@gcarleo The test fail on MacOS during the installation of `mpi4py`, as far as I can see. If I am not missing something, this seems to be unrelated to this PR.

Does this branch work on your Mac?

femtobit:
Tests pass now; can this PR be merged?

Issue #268, opened 2019-07-31
2.1: Structure of generic MCMC samplers


gcarleo:
We should define a generic `TransitionKernel` - based sampling for MCMC, see for example this [nice implementation](https://www.tensorflow.org/probability/api_docs/python/tfp/mcmc) in Tensorflow. At the same time, we can remove all the duplicate parallel tempering samplers, as done in [`tfp.mcmc.ReplicaExchangeMC`](https://www.tensorflow.org/probability/api_docs/python/tfp/mcmc/ReplicaExchangeMC).

Notice that `tfp.mcmc.sample_chain` returns the states in the generated chain but doesn't store (as far as I can tell) neither the log of the sampled probabilities nor its derivatives

PR #279, opened 2019-08-07
Generic Metropolis-Hastings and parallel tempering versions

gcarleo:
This addresses #268 , introducing a generic `MetropolisHastings` sampler working with batches. All existing samplers have been rewritten in terms of this one, simplifying the codebase and removing duplicates. 

In addition to custom C++ `TransitionKernel`, it is also possible to write transition kernels in pure python if needed. 

For example 
```python

# Defining a custom kernel for MetropolisHastings
# Notice that this sampler exchanges two random sites
# thus preserving the total magnetization
# Also notice that it is not recommended to define custom kernels in python
# For speed reasons it is better to define exchange kernels using CustomSampler
def exchange_kernel(v, vnew, logprobcorr):

    vnew[:, :] = v[:, :]
    logprobcorr[:] = 0.0

    rands = np.random.randint(v.shape[1], size=(v.shape[0], 2))

    for i in range(v.shape[0]):
        iss = rands[i, 0]
        jss = rands[i, 1]

        vnew[i, iss], vnew[i, jss] = vnew[i, jss], vnew[i, iss]

sa = nk.sampler.MetropolisHastings(ma, exchange_kernel, batch_size=16, sweep_size=20)
```


gcarleo:
This is read for me, willing to merge it soon

gcarleo:
Alright, I am merging this today unless there are objections

gcarleo:
Thanks @femtobit

Sources/Sampler/exchange_kernel.hpp
```diff
@@ -0,0 +1,79 @@
+// Copyright 2018 The Simons Foundation, Inc. - All Rights Reserved.
```
twesterhout: 2019

Sources/Sampler/exchange_kernel.hpp
```diff
@@ -0,0 +1,79 @@
...
+
+// Kernel generating local random exchanges
+class ExchangeKernel {
+  // number of visible units
+  const int nv_;
+
+  // clusters to do updates
+  std::vector<std::array<Index, 2>> clusters_;
+
+  std::uniform_int_distribution<Index> distcl_;
```
twesterhout: Just like `std::uniform_real_distribution` `std::uniform_int_distribution` should be constructed on the fly rather than stored as a member.

Sources/Sampler/hop_kernel.hpp
```diff
@@ -0,0 +1,91 @@
...
+  const Index nv_;
+
+  // clusters to do updates
+  std::vector<std::vector<Index>> clusters_;
+
+  Index nstates_;
+  std::vector<double> localstates_;
+
+  std::uniform_int_distribution<Index> distcl_;
+  std::uniform_int_distribution<Index> diststate_;
```
twesterhout: Same here

Sources/Sampler/hop_kernel.hpp
```diff
@@ -0,0 +1,91 @@
...
+
+  void GenerateClusters(const AbstractGraph &graph, int dmax) {
+    auto dist = graph.AllDistances();
+
+    assert(Index(dist.size()) == nv_);
+
+    for (Index i = 0; i < nv_; i++) {
+      for (Index j = 0; j < nv_; j++) {
+        if (dist[i][j] <= dmax && i != j) {
+          clusters_.push_back({i, j});
```
twesterhout: Doesn't this include `{i, j}` twice? Is this intended behaviour?

Sources/Sampler/hop_kernel.hpp
```diff
@@ -0,0 +1,91 @@
...
+      : nv_(psi.GetHilbert().Size()) {
+    Init(psi, psi.GetHilbert().GetGraph(), dmax);
+  }
+
+  void Init(const AbstractMachine &psi, const AbstractGraph &graph,
+            Index dmax) {
+    nstates_ = psi.GetHilbert().LocalSize();
+    localstates_ = psi.GetHilbert().LocalStates();
+
+    GenerateClusters(graph, dmax);
```
twesterhout: Could you add a check that `dmax` is actually valid?

Sources/Sampler/hop_kernel.hpp
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  void GenerateClusters(const AbstractGraph &graph, int dmax) {
```
twesterhout: `int` -> `Index`

Sources/Sampler/hop_kernel.hpp
```diff
@@ -0,0 +1,91 @@
...
+  std::uniform_int_distribution<Index> distcl_;
+  std::uniform_int_distribution<Index> diststate_;
+
+ public:
+  HopKernel(const AbstractMachine &psi, Index dmax = 1)
+      : nv_(psi.GetHilbert().Size()) {
+    Init(psi, psi.GetHilbert().GetGraph(), dmax);
+  }
+
+  void Init(const AbstractMachine &psi, const AbstractGraph &graph,
```
twesterhout: Init should be private and defined in the `cpp` file.

Sources/Sampler/hop_kernel.hpp
twesterhout: `GenerateClusters` should also be private.

Sources/Sampler/local_kernel.hpp
```diff
@@ -0,0 +1,75 @@
...
+#include "Utils/random_utils.hpp"
+
+namespace netket {
+
+// Metropolis sampling generating local moves in hilbert space
+class LocalKernel {
+  std::vector<double> local_states_;
+
+  std::uniform_int_distribution<Index> dist_random_sites_;
+  std::uniform_int_distribution<Index> dist_random_states_;
```
twesterhout: Same here

Sources/Sampler/local_kernel.hpp
```diff
@@ -0,0 +1,75 @@
...
+
+ public:
+  explicit LocalKernel(const AbstractMachine& psi) { Init(psi); }
+
+  void Init(const AbstractMachine& psi) {
```
twesterhout: Should be private and defined in `.cc` file.

Sources/Sampler/local_kernel.hpp
```diff
@@ -0,0 +1,75 @@
...
+    if (!psi.GetHilbert().IsDiscrete()) {
```
twesterhout: Use `NETKET_CHECK` to get prettier error messages.

Sources/Sampler/metropolis_hastings.hpp
```diff
@@ -0,0 +1,85 @@
...
+#include "Utils/exceptions.hpp"
+#include "Utils/random_utils.hpp"
+
+namespace netket {
+
+class MetropolisHastings : public AbstractSampler {
+ public:
+  using TransitionKernel = std::function<void(
+      Eigen::Ref<const RowMatrix<double>>, Eigen::Ref<RowMatrix<double>>,
+      Eigen::Ref<Eigen::ArrayXd>)>;
```
twesterhout: Since `TransitionKernel` is a `std::function` (i.e. cannot be inlined), all transition kernels should be implemented in `.cc` files.

twesterhout:
@gcarleo can we revert this, please?

Sources/Sampler/metropolis_hastings_pt.hpp
```diff
@@ -0,0 +1,125 @@
...
+
+  TransitionKernel transition_kernel_;
+
+  Index n_replicas_;
+  Index sweep_size_;
+
+  // Index holding the position of beta=1
+  Index beta1_ind_;
+
+  Eigen::ArrayXd accepted_samples_;
```
twesterhout: Shouldn't it be an array of `Index`?

Sources/Sampler/hop_kernel.hpp
gcarleo: Yes it's fine, doesn't really matte

Sources/Sampler/custom_local_kernel.hpp
```diff
@@ -29,52 +29,52 @@
...
-  int nstates_;
+  Index nstates_;
+
   std::vector<double> localstates_;
 
-  int sweep_size_;
+  std::discrete_distribution<Index> disc_dist_;
 
-  LogValAccumulator log_val_accumulator_;
+  std::uniform_real_distribution<double> distu_;
```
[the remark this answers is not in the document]
gcarleo: yes but I don't like too much the asymmetry here, since we are storing as a member (for performance reason) ```std::discrete_distribution``` anyways

Sources/Sampler/hop_kernel.hpp
gcarleo: I can agree for distcl_, but diststate_ is called twice inside a loop, and I'd need to create the distribution twice... that's a bit ugly t.b.h., and no, I don't want to use a lambda inside a lambda :)

Sources/Sampler/metropolis_hastings_pt.hpp
gcarleo: It should,but then I have issues doing 
```c++
  accepted_samples_ / static_cast<double>(total_samples_);
```

Sources/Machine/abstract_machine.cc
```diff
@@ -28,23 +28,23 @@
...
 
-namespace {
-inline unsigned ReallyRandomSeed() {
-  std::random_device rd;
-  return rd();
-}
-}  // namespace
-
 void AbstractMachine::InitRandomPars(double sigma,
                                      nonstd::optional<unsigned> seed) {
```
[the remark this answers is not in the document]
gcarleo: Yes but this is going to change the existing interface and I don't like too much the fact that initializing the parameters becomes non-seedable... We can maybe add an overloaded method doing what you suggest, but that's gonna be for another PR!

PR #280, opened 2019-08-07
netket.machine.Torch

twesterhout:
Allows one to use a `torch.jit.ScriptModule` as a NetKet machine.

Backpropagation is not (yet) supported. The purpose of this PR is to access performance of PyTorch in the context of NQS and to decide whether it's worth it to depend on PyTorch by default.

Users who don't want to use PyTorch can simply pass `-DNETKET_USE_TORCH=off` to CMake. Compilation of `netket::PyTorchMachine` will be disabled, and, in Python, `torch.machine.Torch` will no longer be available.


gcarleo:
@twesterhout what was missing to make this PR work?

gcarleo:
Closing this now, see #335

PR #281, opened 2019-08-07
Add test for MCMC statistics

femtobit:
This PR adds a test that checks the MC statistics computed by `netket.stats.statistics` for the case of one MPI process and several different batch sizes (i.e., different numbers of Markov chains).

### DIFF
--- a/Test/Stats/test_stats.py
+++ b/Test/Stats/test_stats.py
@@ -20,7 +20,7 @@ def _setup():
 
 
 def _test_stats_mean_std(hi, ham, ma, batch_size):
-    sampler = nk.sampler.MetropolisLocalV2(ma, batch_size=batch_size)
+    sampler = nk.sampler.MetropolisLocal(ma, batch_size=batch_size)
 
     n_samples = 16000
     num_samples_per_chain = n_samples // batch_size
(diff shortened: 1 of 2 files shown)


Test/Stats/test_stats.py
```diff
@@ -0,0 +1,58 @@
...
+    ham = nk.operator.Heisenberg(hi)
+
+    ma = nk.machine.RbmSpinV2(hi, alpha=2)
+    ma.init_random_parameters()
+
+    return hi, ham, ma
+
+
+def _test_stats_mean_std(hi, ham, ma, batch_size):
+    sampler = nk.sampler.MetropolisLocalV2(ma, batch_size=batch_size)
```
gcarleo: please sure MetropolisLocal(...) since v2 is going to be removed in my PR

femtobit: Good point. Then we should probably merge that one first, so that `MetropolisLocal` supports `batch_size` as argument.

femtobit:
Updated to current `v2.1` after merging #279.

Test/Stats/test_stats.py
```diff
@@ -0,0 +1,59 @@
...
+    assert eloc.shape == (num_samples_per_chain, batch_size)
+
+    stats = statistics(eloc)
+
+    # These tests only work for one MPI process
+    assert nk.MPI.size() == 1
+
+    assert stats.mean == pytest.approx(np.mean(eloc))
+    if batch_size > 1:
+        # error of mean == stdev of sample mean between chains / sqrt(#chains)
```
gcarleo: Did you check this after I fixed the issue with the factor in error of mean in C++?

femtobit: Yes, I changed `ddof=0` (it was 1 previously because of the `m-1` denominator) and added the division by `sqrt(m)`. This matches the error of mean computed with your changes.

PR #283, opened 2019-08-13
V2.1 : Fix wrong factors in new McStats class

gcarleo:
The error on the mean was largely overestimated, because the factor 1/m was missing. I also fixed the calculation of the correlation time, that is now an integer by default. The estimate of the correlation time has large error bars +/-1 (at least) thus it makes sense to round it to just an integer

### DIFF
--- a/Sources/Stats/mc_stats.cc
+++ b/Sources/Stats/mc_stats.cc
@@ -153,7 +153,8 @@ Stats Statistics(Eigen::Ref<const Eigen::VectorXcd> values,
     const auto R =
         std::sqrt(static_cast<double>(n - 1) / static_cast<double>(n) +
                   var.first / var.second);
-    return {mean, std::sqrt(var.first), var.second, correlation, R};
+    return {mean, std::sqrt(var.first / static_cast<double>(m - 1)), var.second,
+            correlation, R};
   }
   return Stats{mean, NaN, NaN, NaN, NaN};
 }
--- a/Sources/Stats/mc_stats.cc
+++ b/Sources/Stats/mc_stats.cc
@@ -15,6 +15,7 @@
 #include "Stats/mc_stats.hpp"
 
 #include <mpi.h>
+#include <cmath>
 #include <nonstd/span.hpp>
 
 #include "Utils/exceptions.hpp"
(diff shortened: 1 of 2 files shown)


Sources/Stats/mc_stats.cc
```diff
@@ -106,6 +107,7 @@ Stats Statistics(Eigen::Ref<const Eigen::VectorXcd> values,
   NETKET_CHECK(values.size() >= local_number_chains, InvalidInputError,
                "not enough samples to compute statistics");
   constexpr auto NaN = std::numeric_limits<double>::quiet_NaN();
+  constexpr auto iNaN = std::numeric_limits<int>::quiet_NaN();
```
twesterhout: Even though this might be legal (i.e. compiles) there is no such thing as an integer NaN. See e.g. https://stackoverflow.com/a/3949459 if you don't trust me :) Also `iNaN == iNaN` will return `true` which should not be the case for NaN.

Sources/Stats/mc_stats.cc
```diff
@@ -141,21 +143,22 @@ Stats Statistics(Eigen::Ref<const Eigen::VectorXcd> values,
...
     const auto t = var.first / var.second;
-    auto correlation = 0.5 * (static_cast<double>(n) * t * t - 1.0);
-    if (correlation < 0.0) correlation = NaN;
+    auto correlation = 0.5 * (t * static_cast<double>(n) - 1.0);
+    if (correlation < 0.0) correlation = 0;
     const auto R =
         std::sqrt(static_cast<double>(n - 1) / static_cast<double>(n) +
                   var.first / var.second);
-    return {mean, std::sqrt(var.first), var.second, correlation, R};
+    return {mean, std::sqrt(var.first / static_cast<double>(m)), var.second,
```
twesterhout: @gcarleo you're now dividing by `m^2` (once during computation of `var` and the second time here). It seems wrong to me. Are you sure this is correct?

Sources/Stats/mc_stats.hpp
```diff
@@ -27,8 +27,8 @@ struct Stats {
   double error_of_mean;
   /// Average of in-chain variances of the observable over all Markov Chains.
   double variance;
-  /// Autocorrelation
-  double correlation;
+  /// Autocorrelation time
+  int correlation;
```
twesterhout: I suggest we stick to using a `double` here for the sake of having a NaN value. I don't have anything against rounding it though.

Sources/Stats/mc_stats.cc
gcarleo: yes, the second factor of m comes because we want the error on the mean, this is consistent... otherwise if you don't divide you would get the variance, not the error on the mean

gcarleo: the thing that is still slightly wrong is that global_var[2] should contain the variance over all samples (i.e. merging all chains on all cores) whereas now it contains the average variance

PR #285, opened 2019-08-15
Additional changes for #279

gcarleo:
See comments in #279


gcarleo:
@twesterhout, it seems that these tests here and in other PRs are failing because of Jax?

twesterhout:
> @twesterhout, it seems that these tests here and in other PRs are failing because of Jax?

Interesting... The error seems to be somewhere inside jax. Let's see if I can reproduce it locally

gcarleo:
Yes it's most likely an internal error of JAX. In any case, can we merge this PR for the moment?

PR #286, opened 2019-08-16
Update Machines to batched versions and start removing Lt

gcarleo:
This PR updates all machines to take advantage of the batched versions.
 
In this PR I have also removed most of the look-up tables, since they do not have any significant effect in speeding up the simulations, once batching is activated. 

A mechanism for internal caching in Machines  and also allowing for the use of `LogValDiff` in local expectation values will be the subject of future PRs, but it is not a top priority at the moment.


gcarleo:
Any comment on this? Can I merge it?

gcarleo:
I asssume there is consent by silent from the others, and merge this

PR #288, opened 2019-08-16
Update setup.py to 2.0

gcarleo:


gcarleo:
Yes but I actually just wanted to release 2.0 with a proper label etc. Since, as you said, there won't be any breaking changes in the APIs between 2.0 and 2.1 I think it's fine to release it at last

femtobit:
Alright. Let's wait for the updated binder config.

PR #290, opened 2019-08-16
Binder Build Bug in V2.0 Branch

jamesETsmith:
This PR addresses some of the binder build problems which involved missing dependencies. I've also tried to reduce the number of packages in `environment.yml` to minimize the build time on binder virtual machines.

Thanks to @femtobit for noticing!

Test things out [here](https://mybinder.org/v2/gh/jamesETsmith/netket/v2.0).


gcarleo:
Is the binder thing bound to a given branch? Can we make it bound to a certain version tag instead?

jamesETsmith:
@gcarleo you can link binder builds to commits, tags, or branches, e.g. check out the [v2.0b3 release](https://mybinder.org/v2/gh/netket/netket/v.2.0b3). If you go to the [mybinder.org](https://mybinder.org/) you can specify the repo url and then use a commit hash, tag, or branch name so we can make binder builds for pretty much any state of NetKet we want.

Issue #113, opened 2018-11-20
Compiling with USE_LAPACK in v2.0 is broken 


gcarleo:
Reported by @fabienalet

fabienalet:
I think it's related to [this](https://github.com/libigl/libigl/issues/651).
The simplest workaround I found seems to 
````
#undef I
````
before
````
#include "Dynamics/pydynamics.hpp"
````
in [pynetket.cc](https://github.com/netket/netket/blob/v2.0/NetKet/pynetket.cc)
It compiles but it's a bit ugly and I'm not sure if it won't break other things later.

gcarleo:
Oh interesting, I thought we had already solved this before! I personally changed all the I to I_ in a lot of parts in the code. Maybe there is some leftover. In what part of the code do we still have `I` ? Maybe it's easier just to rename those and not use the `#undef`

fabienalet:
It is in pybind11 actually. More precisely [here](https://github.com/pybind/pybind11/blob/a23996fce38ff6ccfbcdc09f1e63f2c4be5ea2ef/include/pybind11/detail/common.h#L473) and perhaps elsewhere

gcarleo:
Oh I see, then we should open an issue on pybind11

twesterhout:
@fabienalet and @gcarleo pybind11 is not the one to blame here! It's basically standard practice in template-heavy C++ code to use `Is` or `I` for integral parameter/parameter packs. Most style guides mandate that template parameters must be in CamelCase and normal code's `i` is thus transformed into `I`.
Not to sound harsh, but anyone who includes `<complex.h>` in C++ code is most likely doing something very wrong. Including `<complex.h>` is really just asking for trouble :)
So if anything, I'd submit a bug report to Eigen which includes it in one of the low-level cblas-related parts and doesn't undefine the macros afterwards.

gcarleo:
@twesterhout I wonder wether with separate compilation we can also fix this old issue ?

twesterhout:
I can give it a try.

PR #295, opened 2019-09-18
Set lapack off by default because of Eigen bug

gcarleo:
Eigen still has a bug with the Lapack interface, because of the infamous ``I`` used in templates. This will cause netket installation to fail, thus it is wise to set the lapack option OFF by default.


fabienalet:
There was a simple workaround for previous versions of netket which I reported here
https://github.com/netket/netket/issues/113#issuecomment-440323539
but it does not work anymore. I couldn't find a simple one along the same lines, even though it's certainly possible

Issue #299, opened 2019-09-26
2.1: Remove c++ driver classes for high-order methods


gcarleo:
As done in #276 , we should remove all the C++ high-level applications (variational, supervised, unsupervised), and change them to new python drivers. The new python drivers should accept as much as possible the same arguments of the old C++ drivers, to ease the transition.
Full compatibility with the old drivers should be guarantueed, for example using different namespaces for the new drivers.  The new drivers should be just in the ```netket``` main module (say ```netket.Vmc```, ```netket.Supervised``` etc)

Issue #297, opened 2019-09-26
2.1: Drop support for python 2.x 


gcarleo:
We should just drop support for older versions of python, since most mainstream libraries are doing just the same

PR #302, opened 2019-10-04
Simplify batching logic and improve python driver for vmc

### LINKED ISSUE #300: 2.1: Rename batch_size
The option `batch_size` in Markov Chain samplers should be rename to `n_chains`, since that's more descriptive and corresponds to its actual purpose

gcarleo:
1. This PR addresses #300 changing `batch_size` to `n_chains` and allowing samplers to further accept an (actual) `batch_size` argument that however does not change the number of chains, but only how the calls are distributed. This fixes #300 while giving more freedom in the choice of the batch size, independently on the number of chains. 

2. Python 2 machine are also removed from tests, as per #297. 

3. OpenMP is also deactivated by default, since in my experience it gives more trouble than benefits (especially on Mac), and it is actually faster in most cases to just use the purely serial version without the calls to OpenMP. 

4. The Python driver for vmc (`netket.Vmc`) has been put on par with the C++ driver. The previous version was slower, because the `GenerateSamples` written in C++ was (amazingly) slow. The reasons are not entirely clear to me, but most likely it was because of some implicit copies done by pybind11. The amazing thing is that the pure python version of `GenerateSamples` provided in this PR is (basically) as fast as the pure C++ version, thus there is no need to carry along the C++ version. 

5. A first step towards simplifying batched calls to LocalValues has been done. This is still not optimal, in the sense that batching in principle can happen on irregular sizes (as noticed by @twesterhout). However, the version in this PR is much more maintainable and there is no measurable slowing down with respect to the previous version. In the future, we will address the issue of providing batched versions of `LogValDiff` and similar. 

6. There is now a legacy class (`netket.variational.Vmc`) that is complaint with the legacy version and just calls internally the new `netket.Vmc` driver. This legacy class also has a method `run` identical to the old interface, producing json outputs compatibile with the legacy json output files. 

7. The C++ VMC driver and many other unnecessary classes have been completely removed, addressing part of issue #299.


Sources/Machine/rbm_spin.hpp
```diff
@@ -42,21 +42,39 @@ class RbmSpin : public AbstractMachine {
   VectorType GetParameters() final;
   void SetParameters(Eigen::Ref<const Eigen::VectorXcd> pars) final;
 
-  void LogVal(Eigen::Ref<const RowMatrix<double>> v,
-              Eigen::Ref<Eigen::VectorXcd> out, const any & /*unused*/) final;
+  void LogVal(Eigen::Ref<const RowMatrix<double>> x,
+              Eigen::Ref<Eigen::VectorXcd> out, const any &) final {
+    SubBatch<Eigen::Ref<const RowMatrix<double>>, Eigen::Ref<Eigen::VectorXcd>>(
+        std::bind(&RbmSpin::LogValImpl, this, std::placeholders::_1,
+                  std::placeholders::_2),
```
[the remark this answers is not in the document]
gcarleo: Ah great point, Indeed it makes also sense to use a template for F and a lambda here thanks!

Sources/Machine/abstract_machine.hpp
```diff
@@ -230,12 +237,41 @@ class AbstractMachine {
...
+
  protected:
   AbstractMachine(std::shared_ptr<const AbstractHilbert> hilbert)
-      : hilbert_(std::move(hilbert)) {}
+      : hilbert_(std::move(hilbert)), max_batch_size_(32) {}
+
+  // Function to dispatch sub batches call
+  // This is useful to make sure that batched calls on matrices are never called
+  // for more than max_batch_size_ rows at the time
+  template <class T1, class T2>
```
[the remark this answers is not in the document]
gcarleo: I change this a bit now, let me know if you think of a way to remove completely the templates...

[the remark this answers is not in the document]
twesterhout: Is there a reason why you'd like to get rid of templates here?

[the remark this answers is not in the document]
gcarleo: I had the impression that this was slowing down compilation a bit, but maybe it's because the subbatching should be implemented in the AbstractMachine only and not in all the individual machines (now this is implemented separately for RbmSpin, but that's because this is WIP)

gcarleo:
Concerning this: 

> 2\. Reimplementation of `LocalValues` has a) algorithmically worse performance and b) breaks the guarantee that batch size remains constant within a call to `LocalValues`. Let me illustrate with an example the core difference between the old and new implemetations. Suppose, we need to compute `LocalValues` for 3 spin configurations: s1, s2, and s3. `FindConn` returns 5 for `s1`, 2 for `s2`, and 7 for `s3`. For `max_batch_size=4`, the new implementation will make 5 calls to `LogValDiff` the following batch sizes: 4, 1, 2, 4, 3. The old implementation requires only 4 calls to `LogValDiff` and all of them with batches of 4 elements.

I agree, but that this can be improved. What I would like to do is to have a batched `LogValDiff` (and `FindConn`, I guess) that would manipulate tensors of shape (batch,n_connected,n_visible). In turn one can reshape these tensors into a (batch*n_connected,n_visible) matrix and call batched `LogVal` on this inside the default implementation of `LogValDiff`. The issue is finding a good tensor implementation in C++ that allows for copy-free reshaping and conversions to other types we use, but there are some good options.... I believe this will solve your concern on algorithmic complexity. Even though, right now, there is already no measurable performance penalty in this WIP implementation. Actually, in cases I could try, I found a slight speed up associated with the new version, but that's borderline for timing measurement. 

The advantage of this is that there might still be specific cases where individual machines implement optimized `LogValDiff`. There are wave functions for which there is an actual computational complexity advantage in providing optimized `LogValDiff`... This is true already for RBM, but the advantage shows up only on sufficiently large systems/alpha.

gcarleo:
Hi all, I believe that this PR is now in good shape to be merged in 2.1

Please let me know if you have comments/suggestions! 

Ideally, I'd like to merge this before the end of next week, so that I can continue addressing issue #299 and remove also the C++ drivers for Supervised and Unsupervised

Sources/Operator/abstract_operator.cc
```diff
@@ -44,7 +57,7 @@ void AbstractOperator::ForEachConn(VectorConstRefType v,
 }
 
 Eigen::VectorXcd LocalValues(Eigen::Ref<const RowMatrix<double>> samples,
-                             Eigen::Ref<const Eigen::VectorXcd> values,
+                             Eigen::Ref<const Eigen::VectorXcd> /*log_values*/,
```
femtobit: If we do not use the `log_values`, we should remove this parameter, otherwise it is confusing.

netket/_vmc_driver.py
```diff
@@ -125,33 +126,42 @@ def __init__(
...
-        if n_discard is not None and n_discard <= 0:
+        if n_discard is not None and n_discard < 0:
             raise ValueError(
                 "Invalid number of discarded samples: n_discard={}".format(n_discard)
             )
 
-        self.n_samples = n_samples
+        self._n_chains = sampler.n_chains
+
+        self._n_samples = int(self._n_chains * (n_samples // self._n_chains))
```
femtobit: Is there a reason `self._n_chains` is now marked as private with the underscore, while `self.n_discard` is not?

netket/_vmc_driver.py
```diff
@@ -259,3 +284,37 @@ def info(self, depth=0):
...
+            st = value.asdict()
+            st["Mean"] = st["Mean"].real
+            json_iter[key] = st
+
+    def _init_json_log(self):
+
+        self._json_out = {}
+        self._json_out["Output"] = []
+
+    def run(
```
femtobit: Nice to finally get rid of the C++ JSON output code.

netket/_vmc.py
```diff
@@ -79,13 +80,20 @@ def optimize_fn(_, grad, p):
...
 
     def __init__(
-        self, hamiltonian, sampler, optimizer, n_samples, n_discard=None, sr=None
+        self,
+        hamiltonian,
+        sampler,
+        optimizer,
+        n_samples,
+        n_discard=None,
+        discarded_samples_on_init=0,
```
femtobit: A name like `n_discard_on_init` would be more consistent.

netket/_vmc.py
```diff
@@ -241,9 +276,23 @@ def get_observable_stats(self, observables=None, include_energy=True):
...
+        for _ in range(self.discarded_samples_on_init):
+            self._sampler.sweep()
+
+    def _get_mc_stats(self, op):
+        loc = _local_values(op, self._machine, self._samples, self._logvals)
+        return loc, _statistics(loc)
+
     def __repr__(self):
         return "Vmc(step_count={}, n_samples={}, n_discard={})".format(
             self.step_count, self.n_samples, self.n_discard
```
femtobit: Should be `self._n_samples` here.

netket/_vmc.py
```diff
@@ -211,11 +243,14 @@ def iter(self, n_steps, step=1):
             self.advance(step)
             yield self.step_count
 
-    def add_observable(self, name, obs):
+    def add_observable(self, obs, name):
         """
         Add an observables to the set of observables that will be computed by default
         in get_obervable_stats.
         """
+        if type(name) != str:
```
femtobit: This ad-hoc type-checking is not very Pythonic. In any case, if we have to check the type, a better way would be to write `if isinstance(name, str): ...`.

gcarleo: Good point, the thing is whether we should use type annotations or instead do something else? Maybe just a runtime check of the input value?

gcarleo: Alright actually I agree that the dict should not be restricted to deal with str. I will remove this

netket/_vmc_driver.py
gcarleo: good catch, I have now moved n_samples and n_discard to be properties with dedicated setters. I think it's good to leave the freedom to be able to change the number of samples/discarded samples without the need to re-construct the vmc object!

netket/_vmc.py
femtobit: > Alright actually I agree that the dict should not be restricted to deal with str. I will remove this
Yes, I think not checking the key okay for now.

netket/_vmc_driver.py
femtobit: It's good to have those values as properties, agreed.

netket/variational.py
```diff
@@ -1,31 +1,131 @@
...
+    def run(
+        self, output_prefix, n_iter, step_size=1, save_params_every=50, write_every=50
+    ):
+        self._init_json_log()
+
+        for k in range(n_iter):
+            self.advance(step_size)
+
+            self._add_to_json_log(k)
+            if k % write_every == 0 or k == n_iter - 1:
```
femtobit: In the C++ `Vmc` class, `vmc.run(N, save_params_every=N)` has the counter-intuitive result of only storing the first step `i=0`, whereas `vmc.run(N+1, save_params_every=N)` would have been correct. Fixing this, as this class does by checking `k == n_iter - 1`, is indeed a good thing.

gcarleo: Indeed that was an issue!

Issue #306, opened 2019-10-24
Abstract Python class for samplers 


gcarleo:
In order to better integrate with other backends (pytorch, tensorflow, etc), we need to define abstract python classes for the common objects and functionalities used by the high-level python drivers of NetKet. This issue is to start a discussion about the APIs of the sampler. 
Here is a proposal, feedback is appreciated, especially in light of the requirements from other backends.  @noamwies @orsharir  

```python
from abc import ABC,property,abstractmethod

class Sampler(ABC):
    """
    Abstract base class for NetKet samplers.
    A `Sampler` generates quantum numbers $$ s = s_1\dots s_N $$ distributed
    according to the probability :

    $$P(s_1\dots s_N) = F(\Psi(s_1\dots s_N)),$$

    where F is an arbitrary function. By default F(X)=|X|^2.
    Samplers can generate multiple independent samples at once (see Sweep method).
    The current state of the sample (i.e. the value of s=s_1\dots s_N for all
    the chains) is stored in self.current_sample.
    """


    @abstractmethod
    def seed(self,base_seed):
        """Seeds the random number generator used by the Sampler."""
        pass

    @abstractmethod
    def reset(self,init_random):
        """Resets the state of the sampler, including the acceptance rate statistics
        and optionally initializing at random the visible units being sampled.

        Args:
        init_random: bool, optional
               If True the quantum numbers (visible units) are initialized at random,
               otherwise their value is preserved. (default is False)

        """
        pass

    @abstractmethod
    def sweep(self):
        """Performs a sampling sweep. Typically a single sweep
        consists of an extensive number of local moves.
        """
        pass

    @property
    @abstractmethod
    def current_sample(self):
        """Returns a 2d numpy array of currently sampled configurations. The size of
        the matrix is [N_c,N_v], where N_c is the number of independent chains being
        sampled and N_v is the size of visible vector being sampled.
        """
        pass

    @property
    @abstractmethod
    def current_state(self):
        """The current sampling state of the sampler. This contains a pair
        (visible,log_val) where log_val is the result of machine.log_val(visible)
        """
        pass

    @property
    @abstractmethod
    def n_chains(self):
        """Number of independent chains being sampled.
        """
        pass

    @property
    @abstractmethod
    def machine(self):
        """The machine (Psi(s)) used for the sampling.
        """
        pass

    @abstractmethod
    def machine_func(self):
        """function(complex): Sets and gets the function F to be used for sampling.
[...]

PR #308, opened 2019-11-04
Abstract python sampler

gcarleo:
This is PR defines an abstract python class (interface) for samplers (see discussion here #306 ). 

The abstract class is relatively simple and is found in netket/sampler/abstract_sampler.py . Basically it is implemented as a generator, as also suggested by @noamwies . 

There is also a pure-python implementation of samplers provided with this PR. Most notably, all MetropolisHastings sampler (non Parallel Tempering) and exact samplers have pure python versions. 
The backend for the sampler is chosen automatically, depending on the type of the machine to be sampled from.  This is similar to the mechanism implemented by @femtobit for optimizers, but it's hidden in the class implementations. 

A good side effect of this is that now one can use an arbitrary (pure python) MetropolisHastings sampler in case when the machine is not a C++ machine. The speed of the pure python MetropolisHastings is typically comparable to the pure C++ implementation, for machines that are not "too easy" (say, a RBM with at least alpha=4 for L>20 sites), otherwise it's a bit slower.


netket/__init__.py
```diff
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from __future__ import absolute_import
```
femtobit: Aren't we ending support for Python 2? This line has no effect in Python 3.

gcarleo:
I was just trying to understand why the import fails on Travis... Any idea?
It works on mac

femtobit:
> I was just trying to understand why the import fails on Travis... Any idea?

I think you need to add the sub-module to `setup.py`, as [it has been done for machine](https://github.com/netket/netket/commit/3d1452bdfe9720e6b14e5b17c1aa7b6a3eed485e), which also has its own subdirectory.

gcarleo:
> I think you need to add the sub-module to `setup.py`, as [it has been done for machine](https://github.com/netket/netket/commit/3d1452bdfe9720e6b14e5b17c1aa7b6a3eed485e), which also has its own subdirectory.

Thank you @femtobit this was very helpful!

gcarleo:
@noamwies is this PR ok for what concerns the python interface of samplers? Do you think we can merge this? Then we can think about the general interface of `machine` and see how to make it interact nicely with FlowKet

gcarleo:
Alright, I would then merge this, so that we can move forward with all the rest, including an abstract interface for machines in python. Anybody is against merging? :)

Issue #309, opened 2019-11-06
Can't install using pip install netket after installing the requirements


VolodyaCO:
I have installed the requirements successfully:

`sudo apt-get install cmake libopenmpi-dev openmpi-bin openmpi-doc`

Now, when I try installing it I get errors (which I will show at the end of this issue).

When I follow the instructions for [compiling the package](https://www.netket.org/getstarted/compiling/) I get:

> netket/build$ cmake ..
> -- The C compiler identification is GNU 7.4.0
> -- The CXX compiler identification is GNU 7.4.0
> -- Check for working C compiler: /usr/bin/cc
> -- Check for working C compiler: /usr/bin/cc -- works
> -- Detecting C compiler ABI info
> -- Detecting C compiler ABI info - done
> -- Detecting C compile features
> -- Detecting C compile features - done
> -- Check for working CXX compiler: /usr/bin/c++
> -- Check for working CXX compiler: /usr/bin/c++ -- works
> -- Detecting CXX compiler ABI info
> -- Detecting CXX compiler ABI info - done
> -- Detecting CXX compile features
> -- Detecting CXX compile features - done
> -- [NetKet] CMAKE_BUILD_TYPE not specified, setting it to Release. Use `-DCMAKE_BUILD_TYPE=...` to overwrite.
> -- Found PythonInterp: /home/vladimir/anaconda3/bin/python3.7 (found version "3.7.1") 
> -- Found PythonLibs: /home/vladimir/anaconda3/lib/libpython3.7m.so
> -- Found MPI_C: /home/vladimir/anaconda3/envs/mybinder-env/lib/libmpi.so (found version "3.1") 
> -- Found MPI_CXX: /home/vladimir/anaconda3/envs/mybinder-env/lib/libmpi_cxx.so (found version "3.1") 
> -- Found MPI: TRUE (found version "3.1")  
> -- Found OpenMP_C: -fopenmp (found version "4.5") 
> -- Found OpenMP_CXX: -fopenmp (found version "4.5") 
> -- Found OpenMP: TRUE (found version "4.5")  
> -- Looking for sgemm_
> -- Looking for sgemm_ - not found
> -- Looking for pthread.h
> -- Looking for pthread.h - found
> -- Performing Test CMAKE_HAVE_LIBC_PTHREAD
> -- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Failed
> -- Looking for pthread_create in pthreads
> -- Looking for pthread_create in pthreads - not found
[...]

PR #310, opened 2019-11-06
Version 2.0.1 addressing some common installation problems

gcarleo:
fixes for example the usual Lapack problem on Linux, see #309

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,7 +15,7 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 
 option(NETKET_BUILD_TESTING "Build unit tests." OFF)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -182,8 +182,14 @@ if(${CMAKE_GENERATOR} STREQUAL "Ninja")
     endif()
 endif()
 
+
 if(NETKET_USE_BLAS)
     find_package(BLAS REQUIRED)
+else()
```
femtobit: Isn't this a bit confusing? This way, even if I explicitly pass `NETKET_USE_BLAS=OFF`, it will still be used if found. (At least the name is misleading, with this change, the meaning is more like `NETKET_FAIL_IF_BLAS_NOT_FOUND`.)

gcarleo:
Superseded by v2.1

Issue #216, opened 2019-06-12
Is there a way to add internal links to the docs? 


twesterhout:
All the documentation generators I'm familiar with have some ways to create internal links between different parts of the docs. E.g. if I want to refer to another function in Sphinx, I can do ```py:func:`some_function` ```. What's the proper way to accomplish that in mydocstring?

gcarleo:
There is not such a functionality yet. We can add it I guess...

ooreilly:
If you know where the docs are for that function, it might be possible to use a markdown link. Example: `[some_function](functions.md#some_function)`.  
I would also recommend including some tool that can check for dead links.

twesterhout:
> If you know where the docs are for that function, it might be possible to use a markdown link. Example: [some_function](functions.md#some_function)

This kind of defeats the purpose of having a tool generate the docs, doesn't it?

@ooreilly how difficult would it be to add such functionality to mydocstring?

ooreilly:
@twesterhout  if you do not not like that solution, let us come up with something better. What should the syntax look like and what should you be able to achieve with it? How do you want dead links to be handled? Or do you prefer to reuse the Sphinx syntax, and if so what features should be supported?
The simplest solution I can think of would be to let the tool parse your custom link and convert it into a markdown link, which would then eventually get converted to html.

Here is one idea. We could use the markdown syntax for a link but leave the label empty. Example : `[](some_function)` which would then get converted into `[some_function](#some_function)` for `some_function` residing in the same module as the docstring containing the reference. Or if the thing that you are referencing is in some other module, then maybe you would write `[](netket.module.class.method)` which would get converted into `[netket.module.class.method](docs/module.md#method)`. If there are multiple things on the same page with the same name, then I believe they are referenced by `#method`, `#method-1`, `#method-2` in the order of apperance, at least that is how it works for github flavored markdown.  If the link is broken, then the documentation generator can emit a warning and render the link as plain-text instead.

twesterhout:
> What should the syntax look like and what should you be able to achieve with it?

Here's a quote from [standardese](https://github.com/foonathan/standardese) docs:
> To link to an entity, use the syntax `[link-text](<> "unique-name")` (a CommonMark link with empty URL and a title of `unique-name`). If you don't want a special `link-text`, this can be shortened to `[unique-name]()` (a CommonMark link with empty URL and the name of an entity as text).
[...]

PR #319, opened 2019-11-26
Update docstrings to conform sphinx rst 

gcarleo:
This PR addresses the issues and limitations related to maintaining a custom-based solution for documentation (see for example #216, as well as the complexity of the jekyll-based current website). 

We now switch to a sphinx-based system, where docstrings are still written (as before) in Google style. The only main difference is that math formulas are now to be written using sphinx's [math enviroment ](https://shimizukawa-sphinx.readthedocs.io/en/latest/ext/math.html). 

Docs for new classes however are very easy to add, using the sphinx facilities. The only requirement is to add entries to ```Docs/source/documentation.rst```. For example:

```rst
Operators
----------------

.. autosummary::
   :toctree: modules
   :nosignatures:

   netket.operator.BoseHubbard
   netket.operator.GraphOperator
   netket.operator.LocalOperator
   netket.operator.Ising
   netket.operator.Heisenberg
````

fully specifies documentations for operators. 

The web pages can be built going in ```Docs``` and doing ```make html``` (after having installed all the required sphinx extension). 

At present, there are some small inconsistencies in the web page rendering. Mostly in the main page (help is appreciated!!!).  This happens mostly because support of bootstrap in sphynx is not complete, and one needs to use raw html in most pages. 

**Notice that files in `source/modules/` are produced automatically by the autosummary extension and shouldn't be modified manually**


femtobit:
This is great! I think the move to sphinx will make maintaining the documentation a lot easier.

Some things I noticed (at least when I render the pages on my machine):

Code blocks are not always rendered correctly, e.g. here, under "Examples" (and in many other places):
image

Formulas in `RbmSpinSymm` look broken:
image

Method arguments are not rendered correctly everywhere, e.g., in `hilbert.Spin`:
image

gcarleo:
Thanks @femtobit , indeed there are still a few glitches. The main issue there is that the docstring format we were using now was taking $$ formula_here $$ for Latex, whereas sphinx accepts a [different syntax](https://shimizukawa-sphinx.readthedocs.io/en/latest/ext/math.html). This means that in principle we need to go through the docs one by one and change the formulas. I would like to avoid that, and I am still looking for an alternative

gcarleo:
Ah maybe that's what we need! https://www.sympy.org/sphinx-math-dollar/

gcarleo:
I have moved the Docs to the dedicated repository (netket-website), and it is now already online. We will still need to merge this branch for the changes done to some docstrings. 
The only thing I didn't do yet is to set up tests for the docs... It is less straightforward than I though with sphinx?

femtobit:
Looks quite good now.

One more thing I noticed: On the pages of individual classes (for example [`Edgeless`](https://www.netket.org/modules/netket.graph.Edgeless.html#netket.graph.Edgeless)), there is no navigation in the sidebar. This would be pretty helpful to navigate back to the summary page, however.

gcarleo:
> One more thing I noticed: On the pages of individual classes (for example Edgeless), there is no navigation in the sidebar. This would be pretty helpful to navigate back to the summary page, however.

Yes, that's something to improve, I also wanted to add that at some point. However at the moment I'd rather postpone this to another time also because to make it nice, it involves playing with templates/etc... 

I'd suggest we merge this now, so that future classes are documented with a consistent style!

PR #336, opened 2020-02-05
Add edge color property to AbstractGraph

femtobit:
This PR makes `AbstractGraph::EdgeColors` available in Python.


femtobit:
Fair enough. I've already had this lying around in my code for a while anyways as I've needed it in the past, so it can't hurt to add it for now, even if it's replaced later.

Issue #341, opened 2020-02-20
progress bar for interactive VMC run 


orialb:
Hi, 
First of all thanks for developing this very nice library!

This is a minor feature, but it would be nice to have a progress bar when playing with the library interactively. It is quite simple to implement it using the `tqdm` library, it would look something like this:
<img width="672" alt="image" src="https://user-images.githubusercontent.com/1208196/74938571-8d875580-53ee-11ea-8a54-4a8791df5ab2.png">

It could be enabled/disabled by a keyword argument `show_progress` (or something like this) for `Vmc.run` (I think it could also be made to work with MPI, but in case not it can be disabled by default and enabled by the user request). 

If there is interest I could make a PR for this.

orialb:
(actually maybe there is no problem at all with MPI, because it is used somewhere inside `Vmc.advance` as far as I understand and the progress bar only tracks the number of iterations in `Vmc.run`)

gcarleo:
Thank you @orialb for this proposal, it seems like a very nice idea! Yes please, it'd be great if you could make a PR (just remember to do it to the branch v2.1, not to master) thank you

orialb:
Great, I'll make the PR at some point during the following days.
Regarding the feature:
1. Are there any other statistics that are useful to print other than the mean energy?
2. Would you prefer the progress bar to be disabled or enabled by default?

gcarleo:
1. It is OK to put only the energy, but ideally you might want to do something a bit more general, say 
```python
show_progres=["Energy","SigmaX"...]
```
and it fetches the associated observables using the method get_observable_stats (take a look at netket/_vmc.py)

on the other hand, something like 
```python
show_progress=True 
```
or 
```python
show_progres=[]
```
would show the iteration progress only

2. I would say that by default show_progress=None

PR #344, opened 2020-02-22
Progress bar for VMC driver

orialb:
This PR implements a progress bar that can be displayed during the VMC run, as discussed in https://github.com/netket/netket/issues/341. The progress bar is implemented using [tqdm](https://github.com/tqdm/tqdm), which is now added as a dependency.

With this PR, one can call `vmc_driver.run` with an extra argument `show_progress` that is used to enable the progress bar. The options for `show_progress` are:
1. `None` (deafult) - progress bar is disabled.
2.  list/tuple containing names of observables (e.g. `["Energy", "Sx"]`)- the observables' mean values will be displayed in the progress bar. Observables which were not added to the driver using `add_obsrvable` will be ignored. 
3. `True`/`[]` or any other value not listed above- display the progress bar without any information about observables. 
(If wanted I could add a check that makes sure that `show_progress` is either `True`/`[]` or the values above and throws an error otherwise).

I also added an argument `show_progress_fmt` which allows controlling the formatting of the printed mean values (default is ".4f"). But if this is an overkill I can remove it. 

In a Jupyter notebook the progress bar looks like this: 
image

### DIFF
--- a/netket/variational.py
+++ b/netket/variational.py
@@ -5,6 +5,7 @@ from ._C_netket.optimizer import SR as _SR
 import json
 import warnings
 import numpy as _np
+from tqdm import trange
 
 
 class Vmc(object):
@@ -94,21 +95,31 @@ class Vmc(object):
         self._json_out["Output"] = []
 
     def run(
-        self, output_prefix, n_iter, step_size=1, save_params_every=50, write_every=50
+        self,
+        output_prefix,
+        n_iter,
+        step_size=1,
+        save_params_every=50,
+        write_every=50,
+        show_progress=True,
     ):
         self._init_json_log()
 
-        for k in range(n_iter):
-            self.advance(step_size)
+        with trange(n_iter, disable=not show_progress) as itr:
+            for k in itr:
+                self.advance(step_size)
 
-            self._add_to_json_log(k)
-            if k % write_every == 0 or k == n_iter - 1:
-                if self._mynode == 0:
-                    with open(output_prefix + ".log", "w") as outfile:
-                        json.dump(self._json_out, outfile)
-            if k % save_params_every == 0 or k == n_iter - 1:
-                if self._mynode == 0:
-                    self.machine.save(output_prefix + ".wf")
+                self._add_to_json_log(k)
+                if k % write_every == 0 or k == n_iter - 1:
+                    if self._mynode == 0:
+                        with open(output_prefix + ".log", "w") as outfile:
+                            json.dump(self._json_out, outfile)
+                if k % save_params_every == 0 or k == n_iter - 1:
+                    if self._mynode == 0:
+                        self.machine.save(output_prefix + ".wf")
+
+                # Update energy displayed in the progress bar
+                itr.set_postfix(Energy=(str(self._vmc._stats)))
 
     def iter(self, n_iter=None, step_size=1):
         """
--- a/Test/GroundState/test_vmc.py
+++ b/Test/GroundState/test_vmc.py
@@ -4,6 +4,10 @@ import numpy as np
 import netket as nk
 import netket.variational as vmc
 
+from io import StringIO
+from contextlib import redirect_stderr
+import tempfile
+import re
 
 SEED = 214748364
 nk.utils.seed(SEED)
@@ -94,3 +98,24 @@ def test_vmc_use_cholesky_compatibility():
         vmc = nk.variational.Vmc(
             ha, sampler, op, 1000, use_cholesky=True, sr_lsq_solver="BDCSVD"
         )
+
+
+def test_vmc_progress_bar():
+    ha, sx, ma, sampler, driver = _setup_vmc()
+    driver.add_observable(sx, "sx")
+    tempdir = tempfile.mkdtemp()
+    prefix = tempdir + "/vmc_progressbar_test"
+
+    f = StringIO()
+    with redirect_stderr(f):
+        driver.run(prefix, 5)
+    pbar = f.getvalue().split("\r")[-1]
+    assert re.match(r"100%\|#*\| (\d+)/\1", pbar)
+    assert re.search(r"Energy=\([-+]?[0-9]*\.?[0-9]*", pbar)
+    assert re.search(r"var=[-+]?[0-9]*\.?[0-9]*", pbar)
+
+    f = StringIO()
+    with redirect_stderr(f):
+        driver.run(prefix, 5, show_progress=None)
+    pbar = f.getvalue()
+    assert not len(pbar)
--- a/setup.py
+++ b/setup.py
@@ -167,5 +167,5 @@ setup(
          neural networks and machine learning techniques.""",
     cmdclass=dict(build_ext=CMakeBuild),
     zip_safe=False,
-    install_requires=["numpy>=1.16", "cmake>=3.12", "scipy>=1.2.1", "mpi4py>=3.0.1"],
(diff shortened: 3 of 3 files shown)


orialb:
I just realized that `get_observbale_stats` actually performs some computation and not just getting some stored values. 
With the code I added it is called twice, once in `_add_to_json_log`  and a new call in `_progress_bar_observables`, so this is inefficient and I should change that.

I see several options here:
1. Show only energy  in the progress bar to avoid extra complications.
2. Have `_add_to_json_log` return the `stats` dict and use it for the progress bar.
3. Get the `stats` dict in the main loop in `run` and pass it to `_add_to_json_log` and `_progress_bar_observables`.

gcarleo:
Thank you very much for this PR @orialb 
Yes, my preference is either 1 or 3, but I don't have a strong preference... as long any chosen method does not slow down the optimization loop!

gcarleo:
Also, if you decide for option 1, I would maybe also make the show_progress=True by default

orialb:
I will go with option 1 (display only energy), I think it is better to keep this feature as simple as possible. If a user wants to have a more detailed progress tracking they could generate some plots from the json logs (maybe this kind of feature could be implemented in the future, could use TensorBoard for example).

I partially updated the PR but I realized that there is currently no way to get only the energy using `vmc_driver.get_observable_stats`.  
Could I update `get_observable_stats` in `_vmc.py` such that `get_observable_stats([])` will return only the energy  (currently it will return also all the observables added to the driver), or just somehow expose `_vmc.Vmc._stats` to `variatioanl.Vmc` ?

orialb:
> I partially updated the PR but I realized that there is currently no way to get only the energy using vmc_driver.get_observable_stats.
Could I update get_observable_stats in _vmc.py such that get_observable_stats([]) will return only the energy (currently it will return also all the observables added to the driver), or just somehow expose _vmc.Vmc._stats to variatioanl.Vmc ?

As a suggestion I added a getter for `_vmc.Vmc._stats`, let me know what your think about this solution. 
I think that irregardless of this PR it might be nice for users have a way to get only the energy stats from `variational.Vmc`, but I don't have experience yet with actually using VMC for a real project so maybe I'm wrong about it.

netket/variational.py
```diff
@@ -94,21 +95,36 @@ def _init_json_log(self):
...
+                        with open(output_prefix + ".log", "w") as outfile:
+                            json.dump(self._json_out, outfile)
+                if k % save_params_every == 0 or k == n_iter - 1:
+                    if self._mynode == 0:
+                        self.machine.save(output_prefix + ".wf")
+
+                # Update energy displayed in the progress bar
+                itr.set_postfix(
+                    Energy=("{0:" + show_progress_fmt + "}").format(
+                        self._vmc.energy_stats.mean.real
```
femtobit: Instead of just the energy mean, I'd prefer if you would display `str(stats)`, which looks like `<mean> ± <error> [var=<var>, R=<R>]`, as MC error, variance and also the `R` diagnostic are very helpful to detect problems and assess convergence during a run. (We should document what `R` means somewhere, by the way. Short version: Its supposed to be close to one if the MCMC chains are converged and larger when they are not.)

As `str(stats)` is also supposed to adjust the number of decimal places displayed according to the current MC error, this further allows you to remove the additional `show_progress_fmt` argument, which I consider a good think in order to keep the interface tidy.

netket/_vmc.py
```diff
@@ -186,6 +186,10 @@ def n_discard(self, n_discard):
             else self._n_samples_node * self._batch_size // 10
         )
 
+    @property
+    def energy_stats(self):
```
femtobit: I don't really see the need for making the energy stats a public property at the moment. (In the internal code, you can use `self._stats` in its place of course.)

> I think that irregardless of this PR it might be nice for users have a way to get only the energy stats from variational.Vmc

I definitely agree. However, `Vmc.run` is primarily a convenience function that runs the VMC optimization loop without the user having to care about the output (which is also why I think including a progress bar makes a lot of sense).
If more control is wanted, the interface for that is `Vmc.iter`. If one does not want to get all observables but just the energy (and, say, the other observables every `n`th step), one could e.g. run
```python
other_obs = [SzSz, SxSx, SySy]
vmc = Vmc(...)
for step in vmc.iter(100):
    print(f"Energy = {vmc.get_observable_stats()}")
    if step % n == 0:
        print(f"Other observables: {vmc.get_observable_stats(other_obs, include_energy=False)}")
```
(For a progress bar in that setting, users of course can add a custom progress bar right now (`for step in tqdm(vmc.iter(100), total=100)`), as I've in fact been doing for a while in some scripts.
Using `add_observable` is only really necessary when one wants to use the simpler `Vmc.run` interface and in that case, having `Vmc.energy_stats` does not add too much value (as the other observables have already been computed anyway, when the users can access that property).

orialb: Thanks for the detailed comments.

> I don't really see the need for making the energy stats a public property at the moment. (In the internal code, you can use self._stats in its place of course.)

Well since `run` is a method of `variational.Vmc` and `_stats` belongs to `_vmc.Vmc` I was under the impression that by python convention I shouldn't access it directly in `variational.Vmc`.

It was indeed not so clear to me why there are two `Vmc` classes. But as I gather from your comment this is a temporary thing and the goal is to have everything inside `_vmc.Vmc` ? 

> Using add_observable is only really necessary when one wants to use the simpler Vmc.run interface and in that case, having Vmc.energy_stats does not add too much value (as the other observables have already been computed anyway, when the users can access that property).

I'm not sure I follow this comment. As far as I understand, with the current implementation `Vmc.get_observable_stats()` will compute the statistics for all observables which were added  using `add_observables`. What would be the way to access only the energy statistics (assuming that some observables were added) ?

femtobit: > I'm not sure I follow this comment. As far as I understand, with the current implementation Vmc.get_observable_stats() will compute the statistics for all observables which were added using add_observables. 

Indeed, that's true. What I mean is: Since `get_observable_stats(observables=[...])` allows passing in the observables to estimate, it is not necessary to use `add_observable` at all when using the `netket.Vmc.iter` interface (that doesn't have the `run` method altogether), it just provides the default. The only exception is when `Vmc.run` is used, as this of course directly saves the output without giving user code an option to call `get_observable_stats` every step.

The reason for its existence (and for the two separate Vmc classes) is mostly historical: Some time ago, `netket.variational.Vmc` was written in C++. In the process of porting larger parts of NetKet to pure Python, `netket.Vmc` was introduces, which is a pure Python driver. Due to some interface changes (such as introducing a separate `SR` class instead of passing all SR options in the constructor), we could not immediately replace the old driver (as that would be backwards-incompatible), so the current `netket.variational.Vmc` was written (also in Python) to provide the old interface but using the new driver underneath. In the future, it will be deprecated and `netket.Vmc` will be the only driver.

I'd argue that we should also remove `netket.Vmc.add_observable` at that point. (If we keep a `.run`-style method in the new class, it could just take the observables to output as a parameter instead of having those as state of the driver.)

For your PR, that doesn't make too much of a difference (sorry for the long text, but I thought it could help to provide some context). I'd just argue its okay here to use internal interface even in another Python class, as the existence of two classes is just for backwards compatibility and they are closely coupled anyway.

orialb:
I addressed @femtobit 's comments in a new commit. I could squash all the commits once a final version is approved. 

FWIW I think keeping `Vmc.run` around would be convenient.

orialb:
great, I squashed the commits.

PR #346, opened 2020-02-26
Add acceptance to markov chain in python samplers

gcarleo:
Thanks @kchoo1118 for pointing out that this was missing


gcarleo:
@kchoo1118 is it OK to merge this?

PR #358, opened 2020-03-25
Update code documentation

amitkumarj441:
It would make community people easier to understand/reproduce/enhance the code if documented properly. I hope to document more as I move along.

### DIFF
--- a/Sources/Machine/rbm_spin.cc
+++ b/Sources/Machine/rbm_spin.cc
@@ -18,28 +18,60 @@
 #include "Utils/messages.hpp"
 
 namespace netket {
-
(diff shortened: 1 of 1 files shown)


gcarleo:
Thank you for your first contribution @amitkumarj441 ! 

Could you please do this PR to the branch v2.1 instead of master? We are not accepting direct PR to master unless it's for new releases.  

Also, you will see that the there is a lot of changes going on in v2.1, since the tendency is to eventually remove all the C++ code with pybind11 bindings in favor of a pure python+numba approach.

amitkumarj441:
@gcarleo I've initiated a new PR #359. Closing this now.
Thanks!

PR #362, opened 2020-04-09
Port StochasticReconfiguration to Python

gcarleo:
Porting SR solver to pure python. 

It is accessible using the usual APIs: 

```python
sr = nk.optimizer.SR(diag_shift=0.1)
```

The full solver is implemented through scipy's [lstsq](https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.lstsq.html#scipy.linalg.lstsq) solver. **Performance is the same as the C++ solver for QR-based solvers**. The SVD solvers of scipy are **significantly faster** than the internal Eigen one (at least on conda...). [Using a conda installation one gets MKL automatically, thus also addressing the long-standing issues with Lapack #295]. 
The sparse solvers are implemented through scipy's minres (gmres) and conjugate gradient. They are typically faster than the C++ counterparts. 

(Minor) change in the interface:

- The ```LLT``` solver is not supported. Maybe we can still implement it using some other scipy method, but for the moment the default solution is a QR solver based on ```gelsy```'s LAPACK, which seems at least as fast as LLT. 
- ```store_rank_enabled``` , ```store_covariance_matrix_enabled``` etc are removed. When accessing the corresponding elements, None is returned if the rank (for example) is not being stored. 
- ```scale_invariant_regularization_enabled``` becomes a property ```scale_invariant_regularization``` that can be set True or False 
- added a is_holomorphic method to SR, that can (must!) be set by the drivers, otherwise in the previous interface we were relying on the user to pick the correct version of SR, leading to possible issues  

This will also later allow to remove the dependency on MPI in C++ code already as v2.1 (see #324).


fabienalet:
I'll take a look as soon as I can (simple working example would help me!)

Regarding LLT, why not using the [scipy implementation](https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.cholesky.html) of Cholesky decomposition ?

gcarleo:
@fabienalet yes thanks, it might be good to have also an LLT implementation! 

here is a minimal example 

```python
import netket as nk
import numpy as np
import time

n_par = 1000
n_samples = 40000

# Fake random derivatives
der_logs = np.random.normal(size=(n_samples, n_par)) + \
    1.0j * np.random.normal(size=(n_samples, n_par))

grad = np.random.normal(size=(n_par)) + \
    1.0j * np.random.normal(size=(n_par))


# New python class
sr = nk.optimizer.PySR(diag_shift=0.1, lsq_solver="QR")
dp = np.empty(n_par, dtype=np.complex128)

t0 = time.time_ns()
dp = sr.compute_update(der_logs, grad, dp)
t1 = time.time_ns()
print('Time (new): ', (t1 - t0) / 1.0e9)

# Old class
sr = nk.optimizer.SR(diag_shift=0.1, lsq_solver="LLT")
dp1 = np.empty(n_par, dtype=np.complex128)

t0 = time.time_ns()
dp1 = sr.compute_update(der_logs, grad, dp1)
t1 = time.time_ns()
print('Time (old): ', (t1 - t0) / 1.0e9)

assert(np.allclose(dp, dp1))
```

This on my computer gives 
```
Time (new):  4.315351
Time (old):  7.521601
``` 

if you try the SVD option (""BDCSVD"" in the old code, "SVD" or "BDCSVD" in the new one) it gives me 

```
Time (new):  4.538972
Time (old):  9.15326
```

Of course this is strongly dependent on the underlying LAPACK implementation, but the scipy version basically works out of the box...

fabienalet:
Well, I wasn't able to check what was the old code giving in terms of performances, but the new one works all fine and seems fast enough. Thus ready to merge.

I will take a look to see if Cholesky can easily be added.

Test/Optimizer/test_sr.py
```diff
@@ -18,36 +18,25 @@ def test_svd_threshold():
...
     Test SVD threshold option of BDCSVD
     """
-    with pytest.raises(
-        ValueError, match="svd_threshold option only available for BDCSVD solver"
-    ):
-        SR(svd_threshold=1e-3)
-    with pytest.raises(
-        ValueError, match="svd_threshold option only available for BDCSVD solver"
-    ):
+    with pytest.raises(ValueError):
```
femtobit, line 21: Could you match on the correct error message? Many things can throw a `ValueError`, we should check the error is the correct one.

Test/Optimizer/test_sr.py
```diff
@@ -18,36 +18,25 @@ def test_svd_threshold():
...
-        out = np.copy(b)
-        sr.compute_update(a1, b, out)
+        out = sr.compute_update(a1, b)
         return out
 
     sr = SR_with_threshold(1e-1)
     assert np.allclose(solve(sr, a, b), [1.0, 0.0, 0.0])
 
-    sr = SR_with_threshold(1e-3)
+    sr = SR_with_threshold(1e-4)
```
femtobit, line 38: Is this change due to a different interpretation of the `threshold` parameter in the new code?

Test/GroundState/test_groundstate.py
```diff
@@ -98,7 +99,8 @@ def test_vmc_iterator_iterative():
         for name in "Energy", "SigmaX":
             assert name in obs
             e = obs[name]
-            assert hasattr(e, "mean") and hasattr(e, "variance") and hasattr(e, "R")
+            assert hasattr(e, "mean") and hasattr(
+                e, "variance") and hasattr(e, "R")
```
femtobit, line 103: This is not the [black](https://github.com/psf/black) code style we [agreed on previously](https://github.com/netket/netket/pull/190), which is also the reason why this PR seems to contain so many changes that are purely formatting. Could you run `black` on all the files you changed? This should eliminate these unneeded changes.

Test/Optimizer/test_sr.py
gcarleo: if you are **exactly** at the threshold you basically get random behavior you should be slightly below it to make sure it is working properly

Test/GroundState/test_groundstate.py
femtobit: We don't have to do this in this PR though. I can also make a separate PR later, once we have the changes for 2.1 finalized.

gcarleo: sure I can do that... even though I find that having lines than are too long really hurts readability, so I am not sure black is doing a reasonable job there

Test/Optimizer/test_sr.py
gcarleo: Indeed in the old case you already had 
```python
    sr = SR_with_threshold(1e-1)
    assert np.allclose(solve(sr, a, b), [1.0, 0.0, 0.0])
```

i.e. the threshold was one order of magnitude smaller

Test/GroundState/test_groundstate.py
femtobit: The line length is something that can be changed. [Black uses 88 as default](https://github.com/psf/black#line-length) specifically, because the developers found that in the code they looked at there were often lines slightly over the 80 character limit.

But it is possible to change that default to 80 or maybe 82, if you prefer. Could be worth a try.

netket/optimizer/stochastic_reconfiguration.py
```diff
@@ -0,0 +1,303 @@
...
+            rep += "iterative (Conjugate Gradient)"
+        else:
+            rep += self._lsq_solver
+        rep += "\n"
+
+        return rep
+
+    def _linear_operator(self, oks, n_samp):
+        n_par = oks.shape[1]
+        shift = self._diag_shift
```
femtobit, line 268: This variable is unused.

netket/optimizer/stochastic_reconfiguration.py
```diff
@@ -0,0 +1,303 @@
...
+class SR:
+    r"""
+    Performs stochastic reconfiguration (SR) updates.
+    """
+
+    def __init__(
+        self,
+        lsq_solver=None,
+        diag_shift=0.01,
+        use_iterative=True,
```
femtobit, line 18: `use_iterative` is now determined solely by the `lsq_solver` option as it is overwritten below in any case. So it should probably not be a separate parameter.

netket/optimizer/stochastic_reconfiguration.py
```diff
@@ -0,0 +1,303 @@
...
+        is_holomorphic=True,
+        svd_threshold=None,
+        sparse_tol=None,
+        sparse_maxiter=None,
+    ):
+
+        self._lsq_solver = lsq_solver
+        self._diag_shift = diag_shift
+        self._use_iterative = use_iterative
```
femtobit, line 27: No need to set it here, as it is changed below anyway.

netket/optimizer/stochastic_reconfiguration.py
```diff
@@ -0,0 +1,303 @@
...
+        # Quantities for sparse solver
+        self.sparse_tol = 1.0e-5 if sparse_tol is None else sparse_tol
+        self.sparse_maxiter = sparse_maxiter
+        self._lsq_solver = lsq_solver
+        self._x0 = None
+        self._make_solver()
+
+        self._comm = MPI.COMM_WORLD
+
+    def _make_solver(self):
```
femtobit, line 47: Maybe call this `_init_solver`. `make_...` sounds like the method should return something, and not modify state of the object.

netket/_vmc.py
```diff
@@ -158,7 +160,7 @@ def _forward_and_backward(self):
 
             dp = _np.empty(self._npar, dtype=_np.complex128)
 
-            self._sr.compute_update(_der_logs, grad, dp)
+            dp = self._sr.compute_update(_der_logs, grad, dp)
```
femtobit, line 163: No need to assign `dp` here, its the same array that is passed in as `out` argument already.

netket/optimizer/stochastic_reconfiguration.py
femtobit: Previously, the default was to not use an iterative solver. Is there a reason for changing this default?

netket/optimizer/stochastic_reconfiguration.py
```diff
@@ -0,0 +1,303 @@
...
+
+class SR:
+    r"""
+    Performs stochastic reconfiguration (SR) updates.
+    """
+
+    def __init__(
+        self,
+        lsq_solver=None,
+        diag_shift=0.01,
```
femtobit, line 17: Since we seem to be changing defaults already: Isn't a shift of 0.01 a bit large as default? Maybe something like `1e-4` would be more appropriate?

gcarleo:
@femtobit ok to merge this now?

netket/optimizer/stochastic_reconfiguration.py
```diff
@@ -0,0 +1,303 @@
...
+        r"""bool: Whether to use the scale-invariant regularization as described by
+                    Becca and Sorella (2017), pp. 143-144.
+                    https://doi.org/10.1017/9781316417041
+        """
+        return self._scale_invariant_pc
+
+    @scale_invariant_regularization.setter
+    def scale_invariant_regularization(self, activate):
+        assert activate is True or activate is False
+        self._scale_invariant_pc = activate
```
femtobit, line 229: This line should be below the `if` block, so that the internal state of `sr` is not changed when the exception is thrown. (Otherwise, it could be confusing when someone uses SR interactively.)

femtobit:
I've had some more minor comments. Otherwise looks very good to me.

netket/optimizer/stochastic_reconfiguration.py
fabienalet: Perhaps it might be good to actually keep the `use_iterative` flag for the moment. I think sooner or later I shall be trying to make a PR on using the cholesky solver, which will likely imply a `use_cholesky` flag (as we did previously).

gcarleo: > `use_iterative` is now determined solely by the `lsq_solver` option as it is overwritten below in any case. So it should probably not be a separate parameter.

there is the case when lsq_solver=None (default value) and use_iterative=True. In that case use_iterative provided in the constructor is used... Basically the most common use pattern would be to call this without specifying lsq_solver but just use_iterative=True/False.

gcarleo: > Previously, the default was to not use an iterative solver. Is there a reason for changing this default?

I noticed that the sparse solver is faster anyways, even on small systems... but I didn't test this extensively, what's your feeling here?

netket/optimizer/stochastic_reconfiguration.py
gcarleo: Indeed it is not changed below unless lsq_solver is not None

netket/optimizer/stochastic_reconfiguration.py
gcarleo: > Perhaps it might be good to actually keep the `use_iterative` flag for the moment. I think sooner or later I shall be trying to make a PR on using the cholesky solver, which will likely imply a `use_cholesky` flag (as we did previously).

@fabienalet yes it makes sense to keep the use_iterative flag (see other comments!) For the cholesky, I would suggest to implement it as an optional ```lsq_solver='cholesky'``` in the constructor of SR or something like that, without introducing a new flag

netket/_vmc.py
gcarleo: Yes for some reason that I don't quite understand I need to do the explicit assignment otherwise dp is not modified... still looking into this

femtobit: Ah, that's because `out` is reassigned instead of written to in line 123 (and equivalent lines further below).

Probably it should be `out[:], ... = self._sparse_solver(...)`.

gcarleo: yeah thanks!

netket/optimizer/stochastic_reconfiguration.py
fabienalet: OK, well noted.

netket/optimizer/stochastic_reconfiguration.py
gcarleo: For most of the examples we have a shift of 1e-4 would be too small...

gcarleo:
> I've had some more minor comments. Otherwise looks very good to me.

should be good now

netket/optimizer/stochastic_reconfiguration.py
fabienalet: Trying to implement this, but I see the cholesky solver is very sensible to the shift and easily run into instabilities. I'm trying to solve this by comparing systematically to the QR solver. Quick basic question: what is the way to fix all random number generators to be the same in two different runs? I thought fixing the seed of the machine, as well as netket.random.seed would do it but it doesn't seem to be the case as(maybe I'm missing something obvious/did a mistake)

gcarleo: @fabienalet yes if you are using the c++ sampler backend for the moment you need to fix also ```netket.utils.seed``` ... this is going to be removed though

PR #374, opened 2020-04-15
Port Optimizers to Python

### LINKED ISSUE #368: Port optimizers to python
Define an AbstractOptimizer class in python and rewrite the c++ optimizers in python (maybe numba will be necessary, maybe not)

gcarleo:
Rewrite of optimizers in Python. 

In the process, I also a wrote a wrapper for PyTorch solvers:

```python
import netket as nk
from torch.optim import SGD
....

# RBM Spin Machine
ma = nk.machine.RbmReal(alpha=1, hilbert=hi)

# Use a Torch optimizer 
op = nk.optimizer.Torch(ma, SGD, lr=0.01)
```

The same thing can be done for JAX solvers, thus removing the need for make_opt etc. 

Closes #368


gcarleo:
@emilyjd , I don't know if you have some free time these days, but just in case it would be great if you could port some of your comprehensive tests for optimizers to Python!

femtobit:
I believe supporting JAX optimizers would require a change in the design, so that the optimizer wraps the machine, changes the parameters itself, and code using it just passes in the gradients.

This is necessary here because JAX requires first converting the initial parameters to its own internal state object and passing that to the update function which then returns an updated internal state. Therefore, operating on different parameters than the ones returned from the previous step would require a full reset of the optimizer, forgetting auxiliary information such as momentum.

Encapsulating this would look something like this:
```python
from .abstract_optimizer import AbstractOptimizer

class Jax:
    r"""Wrapper for JAX-like optimizers.
    """

    def __init__(self, machine, optimizer_triple):
        r"""
           Constructs a new ``Jax`` optimizer that can be used in NetKet drivers.

           Args:
               :optimizer_triple (tuple): A triple (3-tuple) of functions as defined
                    by JAX as part of the ``jax.experimental.optimizers`` module.
                    https://jax.readthedocs.io/en/latest/jax.experimental.optimizers.html
        """
        self._init_fun, self._update_fun, self._get_fun = optimizer_triple
        self._machine = machine
        self.reset()

    def update(self, grad):
        self._state = self._update_fun(self._step, grad, self._state)
        self._step += 1
        self._machine.parameters = self._get_fun(self._state)

    def reset(self):
        self._state = self._init_fun(self._machine.parameters)
        self._step = 0
```
Note how `_init_fun` needs to be called to convert the parameters to the internal form. Calling `_init_fun` every update would also reset internal state of the optimizer such as momentum etc. which would break the optimization scheme, so that is not really an option.

gcarleo:
Indeed @femtobit , this is something similar to what I had to do with PyTorch, but in that case I could get away without reinitializing the internal state. 
In any case, the main issue is that here we work under the assumption that machine.parameters are numpy arrays... we could maybe start relaxing that, and allowing the parameters to be whatever type the autograd engine wants, and then work with the associated optimizer for those types...

femtobit:
That is another reason why I think it would make sense to make our optimizers work specifically on a machine (and not generic NumPy arrays). We lose some generality if we ever wanted to optimize something that is not a machine. But we could incorporate for example the following logic: If a machine uses a PyTorch state internally (there should be some way to check this), then the Torch optimizer could just use that. Otherwise, it could create its own tensor and update it based on the machine parameters as your `Torch` class does now.

By the way, JAX optimizers would also work on the `state_dict`, because they do not require their arguments to be arrays, but any pytree of arrays, which includes dictionaries. This could help prevent redundant copies, as the `state_dict` elements reference the internal arrays, while `.parameters` always copies them to squeeze them into a single vector.

gcarleo:
The main issue is that you won't be able to use (say) SR with a Torch machine, unless SR is implemented on Torch types...

gcarleo:
> That is another reason why I think it would make sense to make our optimizers work specifically on a machine (and not generic NumPy arrays).

Anyways I completely agree that we should do this... it's just a matter of understanding what's the best strategy, and it's not yet clear to me... 

Any volunteer? :)

In any case, for the moment I would merge this, and understand that these concrete implementations of AbstractOptimizer objects are meant to work with numpy arrays only, and that there might be other implementations that work with other types ?

emilyjd:
Hi! I am in a time-crunch for the next month or so while I defend and write my thesis. I can work on it after that but I'm not sure how soon you need it.

gcarleo:
@emilyjd alright no problem, if you happen to have time before summer it would be good.. otherwise good luck with your phd defense!

gcarleo:
@femtobit I am merging this now , otherwise things become quickly  unmergeable in 3.0 these days... we should definitely come back to this point of making parameters untied to numpy though

PR #376, opened 2020-04-16
Adding Cholesky solver

fabienalet:
This PR adds the Cholesky (LLt) solver as a possible solver, by choosing lsq_solver="Cholesky"
(technically, this does not solve the least square problem but that doesn't matter).

 My experience on the Ising1d example with this version, and previous C++ versions, is that the Cholesky decomposition is always the faster exact solver. For intermediate system sizes, it is competitive with iterative solvers (but is no longer efficient for larger systems). There might be specific niche cases (e.g. with a lot of symmetries) where it is actually even faster and/or more stable than the iterative solver.

### DIFF
--- a/netket/optimizer/stochastic_reconfiguration.py
+++ b/netket/optimizer/stochastic_reconfiguration.py
@@ -139,7 +139,7 @@ class SR:
                 self._apply_preconditioning(grad)
 
                 if self._lsq_solver == "Cholesky":
-                    c, low = _cho_factor(self._S)
+                    c, low = _cho_factor(self._S, check_finite=False)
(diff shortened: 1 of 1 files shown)


gcarleo:
I guess that at this point we should just decide on the default solvers...

I think it still makes sense to keep the sparse solver as the default one, since it is the only one that is scalable while it is almost as fast as the fastest full solvers on small systems. What do you think?

Maybe we could change the default full solver to Cholesky though, which is what we were doing in the previous versions anyways?

fabienalet:
@gcarleo  Yes to both (default iterative solver, and Cholesky default exact solver)

PR #383, opened 2020-04-20
Add test for MPI Markov chain statistics and fix Rhat

femtobit:
Complementing the discussion in #365, these are tests that I believe the NetKet statistics should satisfy.

Note that `test_var` currently uses `ddof=0` and passes. I believe we should have a `ddof` parameter on the NetKet functions as well, to apply the denominator corresction just as `np.var` does.

The `test_mc_stats` currently fails, due to the discrepancy in the computation of variance and `Rhat` with `ddof=1` we are discussing in #365. While for `Rhat`, we should definitely use `ddof=1`, I am not sure about `stats.variance`, though I believe the correction is also appropriate there.

Also note that these tests use MPI. I haven't yet figured out how to conveniently run them together with the others using `pytest` (at least not without producing duplicate output, as all processes print data with `pytest`), but they can be run manually using, e.g, `mpirun -np 2 python3 -m pytest -v Test/Stats/mpi_test_stats.py`.

### DIFF
--- a/Test/Stats/mpi_test_stats.py
+++ b/Test/Stats/mpi_test_stats.py
@@ -82,11 +82,11 @@ def test_mean():
 
 
 def test_var():
-    data = np.random.rand(size, 10, 11, 12)
+    data = np.random.rand(size, 10, 11, 12, 13)
     data = comm.bcast(data)
     mydata = data[rank]
 
-    for axis in None, 0, 1, 2:
+    for axis in None, 0, 1, 2, 3:
         if axis is not None:
             # Merge first "MPI axis" with target axis
             refdata = np.moveaxis(data, axis + 1, 0)
--- a/Test/Stats/mpi_test_stats.py
+++ b/Test/Stats/mpi_test_stats.py
@@ -1,0 +1,83 @@
+import netket as nk
+import numpy as np
+from mpi4py import MPI
+
+import pytest
+from pytest import approx
+
+comm = MPI.COMM_WORLD
+size = comm.Get_size()
+rank = comm.Get_rank()
+
+
+def reference_stats(data):
+    """
+    Computes the statistics on the full data, without MPI,
+    for reference.
+    """
+    M_par, M_loc, N = data.shape
+    M_full = M_par * M_loc
+    data_ = data.reshape(M_full, N)
+
+    chain_means = np.mean(data_, axis=1)
+    chain_vars = np.var(data_, axis=1, ddof=1)
+
+    mean_full = np.mean(data)
+    var_full = np.var(data, ddof=1)
+
+    var_mean = np.mean(chain_vars)
+    var_between = N * np.var(chain_means, ddof=1)
+
+    Rhat = np.sqrt((N - 1) / N + var_between / (N * var_mean))
+
+    return mean_full, var_full, Rhat
+
+
+def test_mc_stats():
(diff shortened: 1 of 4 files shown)


femtobit:
@gcarleo This fixes the computation of the `stats.Rhat` statistic, at least up to the `ddof=0/1` question; I have now mostly used the `ddof=0` version. Maybe that we should also adapt this in the `Rhat` formula by replacing the `N-1 / N` term.

Note that I changed `_batch_variance` to return what is essentially (up to the `ddof=1` correction) the between-chain variance, so that it can be used when computing Rhat. I've added the additional division back for the computation of `error_of_mean`; I am not sure though whether this is correct that way, as `stats.mean` already does this division internally once.

femtobit:
For reference, I have also compared the results our statistics implementation to another package that implements MCMC statistics, namely [`arviz`](https://github.com/arviz-devs/arviz).

The code is here:
```python
# coding: utf-8
import netket as nk
import numpy as np
import arviz

g = nk.graph.Hypercube(3, 2)
hi = nk.hilbert.PySpin(g, 0.5)
ham = nk.operator.Heisenberg(hi)
ma = nk.machine.RbmSpin(hi, alpha=2)
sampler = nk.sampler.MetropolisLocal(ma)


def compare():
    vs = sampler.generate_samples(500, True)
    lv = nk.operator.local_values(ham, ma, vs.reshape(-1, hi.size))
    lv = lv.reshape(500, sampler.n_chains)
    lvr = lv.real

    stats = nk.stats.statistics(lvr)
    print(stats)
    print(
        """arviz
    rhat={}
    ESS={} (n_samples={})
    MCSE={}
        """.format(
            arviz.rhat(lvr.T, method="identity"),
            np.round(arviz.ess(lvr.T)),
            lvr.size,
            arviz.mcse(lvr.T),
        )
    )
    print()


ma.init_random_parameters()

compare()

opt = nk.optimizer.Sgd(0.01)
sr = nk.optimizer.SR()
vmc = nk.Vmc(ham, sampler, opt, 1000, sr=sr)
vmc.run("Rhat_temp", 100)

print()
compare()
```

Example run:
```
17.993 ± 0.0025 [var=4.8e-02, R_hat=0.9989]
arviz
    rhat=1.0006477893285546
    ESS=6703.0 (n_samples=8000)
    MCSE=[0.00267666]
        

100%|██████████████████████████████| 100/100 [00:21<00:00,  4.61it/s, Energy=-14.75+0.01j ± 0.079 [var=6.3e+00, R_hat=1.0019]]

-14.61 ± 0.019 [var=3.0e+00, R_hat=1.0017]
arviz
    rhat=1.0010231656875923
    ESS=5761.0 (n_samples=8000)
    MCSE=[0.02092234]
```
Our statistics are not completely equivalent, but seem to more or less agree, especially on the MCSE (the MC standard error, which should be approximately equal to our `stats.error_of_mean`) and for larger sample sizes. There are still differences, but we use different estimates for the standard error (we never compute the effective sample size [ESS]), so that is to be expected, I think. In any case, this might be a useful comparison if we, e.g., decide to use the ESS or implement split-Rhat in the future.

gcarleo:
Thank you @femtobit this comparison is very interesting! 
We actually do compute the effective number of samples, it's just related to the correlation times by N/tau_corr, so you might want to see if also that ones agrees with them? Otherwise this looks good to me, I guess we can merge it

PR #386, opened 2020-04-21
complex jax wavefunctions

kchoo1118:


netket/machine/jax.py
```diff
@@ -39,131 +37,140 @@ def __init__(self, hilbert, module, seed=None):
...
-
-    def save(self, filename):
-        with open(filename, "wb") as output:
-            dump(self.state_dict(), output)
-
-    def load(self, filename):
-        with open(filename, "rb") as input:
-            self.load_state_dict(load(input))
+    @property
+    def parameters(self):
```
gcarleo, line 159: These are the default methods in the abstract class? you can remove them from here I think

PR #393, opened 2020-04-26
Stats/binning.hpp: fix not actually adding the value after merging bins

jmellak:
...which meant every 32th value in the sum was disgarded.

This bug is noticeable f.e. when summing over a specific set of states, where the result might be quite different if there are summands missing.

### DIFF
--- a/Sources/Stats/binning.hpp
+++ b/Sources/Stats/binning.hpp
@@ -104,6 +104,12 @@ class Binning {
         }
       } else {
         Merge();
+        // finally, add the value
+        nb1_ = bins1_[0].N();
+        bins2_[last2_] << val;
+        if (bins2_[last2_].N() == nb1_) {
+          last2_++;
+        }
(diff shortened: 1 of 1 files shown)


gcarleo:
Thank you @jmellak, looks like this was indeed a bug in this binning class! 

Are you still using this functionality though? As per v2.1 this is not used anymore, and in v3.0 we are removing all C++ code (in the last version statistics are done [in here](https://github.com/netket/netket/blob/1b288b5082290e6c404ff30ac3e7d7339d49e580/netket/stats/mc_stats.py#L85)) 

If you still depend on this for your project, it would be good if you can make a separate repository to keep it alive and maintained, otherwise I would encourage you to use v2.1 and not writing c++ extensions for future applications of netket!

gcarleo:
I am closing this now, please let us know if you have further feedback on the python parts!

PR #394, opened 2020-04-27
Merge statistics updates into v3.0

### LINKED ISSUE #384: Rename `netket.stats.Stats.R` to `Rhat`?
As we have seen in the discussion on #365, the `stats.R` parameter is ambiguously named; it is sometimes called "Gelman-Rubin" statistic or "potential scale reduction", but I think the most common way to refer to it is $\hat R$ in LaTeX math or "Rhat" in words (this name is used, e.g., by the [stan](https://mc-stan.org/rstan/reference/Rhat.html) framework). So I propose we also call it `Rhat` starting in 3.0.

What do you think?

femtobit:
It appears that for some reason (probably related to [this](https://github.com/netket/netket/pull/365#issuecomment-615359023)?) the changes to the statistics of #383 are not all present in the current `v3.0` branch.

So let's try this again.

(This also resolves #384.)

### DIFF
--- a/Examples/BoseHubbard1d/bosehubbard1d.py
+++ b/Examples/BoseHubbard1d/bosehubbard1d.py
@@ -44,4 +44,4 @@ vmc = nk.variational.Vmc(
     method="Sr",
 )
 
-vmc.run(output_prefix="test", n_iter=4000)
+vmc.run(n_iter=4000, out="test")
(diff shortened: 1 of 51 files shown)


gcarleo:
Thanks @femtobit ! Is there some residual issue with the naming in the json log?

femtobit:
Good point, should be fixed now too.

gcarleo:
Looks like the Rhat are still giving some issues :)

PR #395, opened 2020-04-27
Add MPI-enabled test cases for statistics

femtobit:
In order to make sure our statistics functions work, I have added a couple of tests to be run with MPI.
Adding those tests also uncovered a bug I had overlooked in the MPI stats functions that is now fixed.

The tests are in a separate `Test_MPI` folder so that they are not automatically run with the other tests. As a side effect, this also provides a quick way to check that one's own MPI environment is correctly set up.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,7 @@ matrix:
             - 'ninja-build'
             - 'g++-7'
             - 'libopenmpi-dev'
+            - 'openmpi-bin'
 
     # Modern toolsets
     - os: linux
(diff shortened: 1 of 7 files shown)


femtobit:
@gcarleo This PR is ready for review now. As far as I can see, the remaining test failures are present in `v3.0` as well (in fact, this PR fixes several warnings that take up lots of space in the Travis output, to the point that the Travis run fails, as can be seen in the current `v3.0` test runs).

femtobit:
In particular, the MPI tests run correctly on Travis now. (Maybe for future PRs we could think about whether there are more test cases where we'd like to know that they run properly with MPI enabled too.)

gcarleo:
Thanks @femtobit ! This indeed solves several issues... the remaining ones are silly numerical precision problems that hopefully will get fixed soon too

PR #402, opened 2020-05-02
Black lint

gcarleo:
This is a style-related PR that adds pre-commit hooks and CI checks for Black code style.
I also updated a bit the CONTRIBUTING.md file etc...


setup.py
```diff
@@ -186,4 +186,5 @@ def _decode(x):
         "numba>=0.48.0",
     ],
     python_requires=">=3.6",
+    extras_require={"dev": ["pytest", "pre-commit", "black"],},
```
femtobit, line 189: We could add `networkx` and `python-igraph` as `dev` requirements here. The tests fail to run if those modules are not installed, but they are (at the moment) not needed otherwise.

gcarleo, line 189:
```suggestion
    extras_require={"dev": ["pytest", "networkx","python-igraph","pre-commit", "black"],},
```

gcarleo: Yes I agree!

PR #405, opened 2020-05-04
Port graphs to Python

VolodyaCO:
Rewrite of graphs in Python. At the moment these are supported:

- CustomGraph (now with the class name as PyCustomGraph)
- Hypercube (with the class name of PyHypercube)

I am only adding the folder with those .py files (as well as an PyAbstractGraph that supports the previous).

Also, I am not sure how to include automorphisms for coloured graphs.


netket/graph/custom_graph.py
```diff
@@ -0,0 +1,99 @@
...
+		return [list(self.graph.neighbors(node)) for node in self.graph.nodes]
+
+	@property
+	def is_connected(self):
+		# TODO: how to check if a multigraph is connected?
+		return _nx.is_connected(self.graph)
+
+	@property
+	def edges(self):
+		return [e for e in self.graph.edges()]
```
gcarleo, line 65: can't you just return self.graph.edges ?

netket/graph/custom_graph.py
```diff
@@ -0,0 +1,99 @@
...
+	r"""A custom graph, specified by a list of edges and optionally colors."""
+
+	def __init__(self, edges=[]):
+		"""
+		Constructs a new graph given a list of edges.
+
+		Args:
+		    edges: If `edges` has elements of type `Tuple[int, int]` it is treated
+		        as a list of edges. Then each element `(i, j)` means a connection
+		        between sites `i` and `j`. It is assumed that `0 <= i <= j`. Also,
```
gcarleo, line 16: does networkx assume that i <= j ?

netket/graph/custom_graph.py
```diff
@@ -0,0 +1,99 @@
...
+
+		edges_array = _np.array(edges, dtype=_np.int32)
+		if edges_array.ndim != 2:
+			raise ValueError("edges must be a list of lists or tuples of the same length (2 or 3)")
+
+		if not(edges_array.shape[1] == 2 or edges_array.shape[1] == 3):
+			raise ValueError("edges must be a list of lists or tuples of the same length (2 or 3), where the third column indicates the color")
+
+		self.graph = _nx.MultiGraph()
+		self.graph.add_edges_from(edges_array)
```
gcarleo, line 46: does this work when you have colors ? shouldn't it be 
```python
self.graph.add_edge(i, j, color=k)
```
or something like that?

netket/graph/custom_graph.py
```diff
@@ -0,0 +1,99 @@
...
+
+		super().__init__()
+		
+	def get_edges_from_color(self, color):
```
gcarleo, line 50: yes this would be nice to have

netket/graph/custom_graph.py
```diff
@@ -0,0 +1,99 @@
+from abstract_graph import PyAbstractGraph
+
+import numpy as _np
+import networkx as _nx
+
+class PyCustomGraph(PyAbstractGraph):
+	r"""A custom graph, specified by a list of edges and optionally colors."""
+
+	def __init__(self, edges=[]):
+		"""
```
gcarleo, line 10: please have a look at CONTRIBUTING.md in the branch 3.0 (in order to have consistent coding style in all the codebase we adopt black coding style)

netket/graph/custom_graph.py
```diff
@@ -0,0 +1,99 @@
...
+	def edge_colors(self):
+		return NotImplementedError
+
+	@property
+	def edges_and_colors(self):
+		return NotImplementedError
+
+	@property
+	def automorphisms(self):
+		# TODO: check if this is better than using the VF2 networkx algorithm
```
gcarleo, line 93: yes  I am not sure this is faster, maybe VF2 might be better... especially because by default we run this with symmetry=False

netket/graph/custom_graph.py
```diff
@@ -0,0 +1,99 @@
+from abstract_graph import PyAbstractGraph
+
+import numpy as _np
+import networkx as _nx
+
+class PyCustomGraph(PyAbstractGraph):
```
gcarleo, line 6: I would call this just ```Graph``` (or ```PyGraph```, for the time being)

netket/graph/custom_graph.py
gcarleo, line 93: if we don't implement the case with colors, we should have a check that in that case when one calls autormophisms an error is raised

netket/graph/custom_graph.py
```diff
@@ -0,0 +1,99 @@
...
+			raise ValueError("edges must be a list of lists or tuples")
+
+		edges_array = _np.array(edges, dtype=_np.int32)
+		if edges_array.ndim != 2:
+			raise ValueError("edges must be a list of lists or tuples of the same length (2 or 3)")
+
+		if not(edges_array.shape[1] == 2 or edges_array.shape[1] == 3):
+			raise ValueError("edges must be a list of lists or tuples of the same length (2 or 3), where the third column indicates the color")
+
+		self.graph = _nx.MultiGraph()
```
gcarleo, line 45: with MultiGraph and not Graph? in the c++ case we don't allow duplicate edges

netket/graph/custom_graph.py
VolodyaCO: Yes, but as this is a MultiGraph, the list that is returned only contains pairs. self.graph.edges contains triplets (with the respective colour for each edge).

netket/graph/custom_graph.py
VolodyaCO: It doesn't, but it orders edges lexicographically. so that `g = nx.MultiGraph([[0,1],[2,1]])` is the same as `g = nx.MultiGraph([[0,1],[1,2]]) `

netket/graph/custom_graph.py
VolodyaCO: Yes, it could be like that, but MultiGraph already handles "colours". So that a third element can be passed. In any case, if I add the edges with an Nx2 array, MultiGraph defaults the colour to 0, whereas if I add edges with an Nx3 array (with colours), MultiGraph adds the corresponding colours to the edges.

netket/graph/custom_graph.py
VolodyaCO: Yep, I left this unfinished because I don't know if it is actually helpful elsewhere.

netket/graph/custom_graph.py
VolodyaCO: I'll do that

netket/graph/custom_graph.py
VolodyaCO: Alright, I will change it

VolodyaCO: Ok, this is a good solution for the moment.

netket/graph/custom_graph.py
VolodyaCO: Because of the colours. Maybe one could have the case where the edges (0,1,0) and (0,1,1) exist because one operator acts on the 0 colour, whereas another operator acts on the 1 colour.

netket/graph/hypercube.py
```diff
@@ -0,0 +1,54 @@
+from custom_graph import PyCustomGraph
+
+import numpy as _np
+import networkx as _nx
+
+class PyHypercube(PyCustomGraph):
+	r"""A hypercube lattice of side L in d dimensions. 
+	Periodic boundary conditions can also be imposed"""
+
+	def __init__(self, length, n_dim=1, pbc=False):
```
gcarleo, line 10: I think that default value used to be pbc=True, I would keep that for consistency

netket/graph/abstract_graph.py
```diff
@@ -0,0 +1,56 @@
...
+import numpy as _np
+
+class PyAbstractGraph(abc.ABC):
+	"""Abstract class for NetKet graph objects"""
+
+	@property
+	@abc.abstractmethod
+	def n_sites(self):
+		r"""int: The number of vertices in the graph"""
+		return NotImplementedError
```
femtobit, line 11: This method is a duplicate of `size`. We should decide on one version.

netket/graph/custom_graph.py
femtobit: Then we should remove the comment about that assumption, if we do not actually require sorted pairs. (Which can be slightly annoying with PBC.)

netket/graph/custom_graph.py
```diff
@@ -0,0 +1,99 @@
...
+		    A 10-site one-dimensional lattice with periodic boundary conditions can be
+		    constructed specifying the edges as follows:
+
+		    >>> import netket
+		    >>> g=netket.graph.PyCustomGraph([[i, (i + 1) % 10] for i in range(10)])
+		    >>> print(g.n_sites)
+		    10
+		"""
+
+		assert isinstance(edges, list)
```
femtobit, line 33: This raises an `AssertionError`, which is different fro the `ValueError` raised by the checks below.

netket/graph/custom_graph.py
femtobit: The `networkx` interface is... interesting here:
```python
>>> g = nx.MultiGraph()
>>> g.add_edges_from([(0, 1, 0), (1, 2, 1), (2, 3, 0), (3, 0, 1)])
>>> [e for e in g.edges]
[(0, 1, 0), (0, 3, 1), (1, 2, 1), (2, 3, 0)]
>>> [e for e in g.edges()]
[(0, 1), (0, 3), (1, 2), (2, 3)]
```
Iterating over `.edges` gives triplets, iterating over `.edges()` gives pairs (unless called as `.edges(keys=True)`, the it also includes the colors/keys).

Maybe we should actually just include the colors as well when returning the edges. Having a separate `edge_colors` property (which actually does not return just the colors, so the name is not perfect) might be unnecessary. What do you think?

netket/graph/custom_graph.py
femtobit: This should be a method not a property, calling `.automorphisms` actually performs a potentially more expensive computation. Then, we could also make the method configurable via an optional argument.

netket/graph/hypercube.py
```diff
@@ -0,0 +1,54 @@
+from custom_graph import PyCustomGraph
+
+import numpy as _np
+import networkx as _nx
+
+class PyHypercube(PyCustomGraph):
```
femtobit, line 6: Since through `networkx.generators.lattice.grid_graph` this class now supports non-square grids by passing, e.g., `length=(10, 5)`, maybe it should be renamed (`Hypergrid` or just `Grid`?).

netket/graph/abstract_graph.py
gcarleo: yes I would remove n_sites and keep only size

netket/graph/hypercube.py
gcarleo: Yes, I would call it Grid but I would also leave at least a conveniency function creating Hypercube from one of these more general cases! Also, it might good to have small functions for the most used cases or for people like me who don't remember specific names without looking at the documentation (e.g.):

```python
def Square(L,pbc=True):
    return Grid(length=(L,L),pbc=pbc)
```
or something like that... maybe also for ```Cube``` and ```Line```.

gcarleo:
As raised by @femtobit , handling colors in our graphs is a bit of a nuisance, and in most cases we don't even use this functionality... I suggest to remove ```edge_colors``` and ```edge_and_colors``` and just make ```edges``` accept a keyword as a filter, such that 

```python
graph.edges(color=1)
```

would return edges with that given color, for example. This should be easy to implement using networkx + it's what one typically needs to define operators. 

What do you think @VolodyaCO @femtobit ?

netket/graph/hypercube.py
femtobit: I fully agree. A couple of creation functions like `Hypercube`,  `Line`, and `Square` like in your example would be convenient.

femtobit:
@gcarleo I agree. The `networkx` interface, where `.edges` can be used both like a property and a method is too much magic for my taste, so let's just stick with a function.

How about
```python
g.edges() # <- returns all edges without colors, [(0, 1), (1, 2), ...]
g.edges(color=True) # <- returns all edges with colors, [(0, 1, 0), (1, 2, 1), ...]
g.edges(color=1) # <- returns edges with color 1, [(1, 2, 1), ...]
```

gcarleo:
Yes that makes sense!

netket/graph/abstract_graph.py
VolodyaCO: I noticed that you use size to refer to the number of nodes, but in mathematics the size of the graph is normally the number of edges. What notation would you like to keep?

netket/graph/custom_graph.py
VolodyaCO: Which one should I use? Should I only raise ValueErrors?

netket/graph/hypercube.py
VolodyaCO: Shall I make these separate classes that inherit the Grid class properties?

VolodyaCO:
Good, I'll make the changes then.

femtobit:
Note also that the graph isomorphisms of the hypercube are not the same as the translation symmetries, which is what the Hypercube class provides in NetKet v2. This is apparent already for, e.g., `Hypercube(3, n_dim=1)` where we have
```python
>>> g2 = nk.graph.Hypercube(3, n_dim=1, pbc=True)
>>> g3 = nk.graph.PyHypercube(3, n_dim=1, pbc=True)
>>> g2.automorphisms
[[0, 1, 2], [1, 2, 0], [2, 0, 1]]
>>> g3.automorphisms
[[0, 1, 2], [0, 2, 1], [1, 0, 2], [2, 0, 1], [1, 2, 0], [2, 1, 0]]
```
We need to decide what we actually want to return here.

gcarleo:
> Note also that the graph isomorphisms of the hypercube are not the same as the translation

@femtobit indeed that was a limitation of the old version. Strictly speaking the isomorphisms contain all symmetries, including reflections, not only translations. The nice thing about networkx is that now it allows to filter on the basis of symmetries (as far as I understand there is a way to put together isomorphisms that correspond to the same generator), so maybe in the future we can add a keyword to get only _some_ automorphisms. For the moment I would keep this simple and just return all automorphisms

netket/graph/abstract_graph.py
gcarleo: yes that's a good point, networkx.size() actually returns the number of edges... Strictly speaking to be more consistent maybe we should do size -> order and/or size -> n_vertices , but I wouldn't leave the size method to return the number of edges, since this would cause an infinite number of bugs (at least on my side) because of this ambiguity (also it seems that there are multiple mathematical definitions for size).

netket/graph/custom_graph.py
VolodyaCO: I will call it `Graph`

netket/graph/custom_graph.py
femtobit: Hm, strictly speaking it should probably a `TypeError` for this specific case, see https://docs.python.org/3/library/exceptions.html#TypeError.

VolodyaCO:
Sorry, I am a newbie to PRs. Do I only push my work to my fork repo, and these changes are automatically seen here? Or should I do something else? Also, I am waiting for you to decide if I include the Line, Square and Hypercube cases as separate classes in separate files, or as separate classes in the same file. @gcarleo

femtobit:
> Do I only push my work to my fork repo, and these changes are automatically seen here?

Yes, the source branch of your PR here is the `v3.0` branch in your fork, https://github.com/VolodyaCO/netket/tree/v3.0 (if in doubt, this is shown at the top of the PR page below the title). If you push commits to that branch, they will appear in this PR right away.

gcarleo:
Yes if you push on your repo this PR will be updated automatically. 
No need to create separate files for those classes, just define them in the same file where you define Grid I guess

netket/graph/abstract_graph.py
VolodyaCO: Alright, I will leave `size` for n_sites, then.

VolodyaCO:
Ok, I'll push all the changes once I finish them, I'll book some time off work to complete this today.

VolodyaCO:
> @gcarleo I agree. The `networkx` interface, where `.edges` can be used both like a property and a method is too much magic for my taste, so let's just stick with a function.
> 
> How about
> 
> ```python
> g.edges() # <- returns all edges without colors, [(0, 1), (1, 2), ...]
> g.edges(color=True) # <- returns all edges with colors, [(0, 1, 0), (1, 2, 1), ...]
> g.edges(color=1) # <- returns edges with color 1, [(1, 2, 1), ...]
> ```

I am worried that edges is being intensively used elsewhere, and that not having the result stored in a class property will slow things down. I implemented already edges as a method, which give the correct results, as it is expected. I am testing if everything is working well. I will push in 5 minutes.

femtobit:
@VolodyaCO
> I am worried that edges is being intensively used elsewhere, and that not having the result stored in a class property will slow things down.

This is not how properties work in Python: When you implement edges as
```
@property
def edges():
    return ...
```
this will still result in a method call every time `.edges` is accessed. All the `@property` decorator does here is hide the function call a bit. Performance wise, not using `@property` at all would make no difference.

In `Graph`, you are calling underlying `networkx` edges method every time anyway, so there are several function calls every time `g.edges` is accessed. I would be very surprised if that would be a performance issue though, the graph has never been a bottleneck as far as I remember. (Unless it were to affect MC sampling, which is a bottleneck. But afaik none of our samplers access the graph, except maybe once for initialization.)

netket/graph/abstract_graph.py
```diff
@@ -0,0 +1,51 @@
...
+
+    @property
+    @abc.abstractmethod
+    def edges(self):
+        r"""list: List containing the edges of the graph"""
+        return NotImplementedError
+
+    @property
+    @abc.abstractmethod
+    def distances(self):
```
femtobit, line 28: For the same reason we discussed regarding `automorphisms`, I think `distances` should not be a property but a method, so that it is more clear to callers that this can perform expensive computation.

netket/graph/abstract_graph.py
```diff
@@ -0,0 +1,51 @@
...
+        r"""list[list]: List containing the distances between the nodes.
+		The fact that some node may not be reachable from another is represented by -1"""
+        return NotImplementedError
+
+    @property
```
femtobit, line 33: This `property` decorator should also be removed.

netket/graph/grid.py
```diff
@@ -0,0 +1,55 @@
...
+                condition = [isinstance(x, int) and x >= 1 for x in length]
+                if sum(condition) != len(length):
+                    raise TypeError
+            except TypeError:
+                raise TypeError("Components of length must be integers greater than 1")
+
+        graph = _nx.generators.lattice.grid_graph(length, periodic=pbc)
+        newnames = {old: new for new, old in enumerate(graph.nodes)}
+        graph = _nx.relabel_nodes(graph, newnames)
+        super().__init__(list(graph.edges))
```
femtobit, line 55: This process seems to change the order of vertices again after you rename them, when initializing `Graph` in the base class constructor with just the edge list, as done in line 55:
```python
>>> grid = nk.graph.PyHypercube(4, 1, pbc=True)
>>> grid.graph.nodes
NodeView((0, 1, 3, 2))
>>> grid.automorphisms
[[0, 1, 3, 2],
 [0, 3, 1, 2],
 [1, 0, 2, 3],
 [3, 0, 2, 1],
 [2, 1, 3, 0],
 [2, 3, 1, 0],
 [1, 2, 0, 3],
 [3, 2, 0, 1]]
```
Note that the identity permutation is not in the list (or, actually, it is because its `[0, 1, 3, 2]`, but that's pretty counter-intuitive).

I think it would be better, to initialize the `Graph` directly with the networx graph created by of `grid_graph`. That way, there is no reordering of nodes.

netket/graph/hypergrid.py
```diff
@@ -0,0 +1,146 @@
+from graph import Graph
+
+import numpy as _np
+import networkx as _nx
+
+
+class Hypergrid(Graph):
```
femtobit, line 7: This PR now seems to include `HyperGrid` and `Hypergrid` classes in `grid.py` and `hypergrid.py`, respectively. That's probably a leftover from previous changes, right?

gcarleo:
Concerning the performance issue of calling edges many times: it is true that this could be a bottleneck @VolodyaCO, and we should be careful there. However at the moment all algorithms we have store the graph edges at the beginning and then reuse it, so in practice the method is called only once. A way to maybe speed up the call to edges would be to store (in the graph) something like 

```python
self._edges=None 
```
and when we call the edges method we check if these have been already computed or not, avoiding recomputing them if already stored. However I think that this is not worth at the moment, let's stick to an easier setup where we just call the underlying networkx method! 

Concerning methods versus properties: it is true that it makes more sense in the new version to transform all these (old) properties into methods, since some of them accept or will accept options + they perform calculations that can be expensive. At this point, I would just transform all the old properties into methods (including n_nodes), to avoid confusion and also to stick to what networkx internally does... 

@VolodyaCO yes, we also need Edgeless (or Empty) since in some cases we need empty graphs with given number of sites... it's a bit weird but I would live it there for the moment. 

For the reordering, I didn't look into that yet, but yes it would be nice to have an ordering where the identity shows up as range(n_nodes) if possible. 

n_nodes seems to me a better name than size at this point, as pointed out by @VolodyaCO size is not what people mean to refer to the number of nodes, unfortunately...

VolodyaCO:
Ok, I'll try to devote some time to solve these issues today

netket/graph/grid.py
VolodyaCO: The thing is that `grid_graph` creates hypergrids where nodes are not integers but tuples, which is why I renamed the nodes.

netket/graph/hypergrid.py
VolodyaCO: In my `hypergrid.py` file I just have `Hypergrid`, `Hypercube`, `Square` and `Chain`. I don't know why `HyperGrid` is appearing.

gcarleo: maybe you didn't commit it? try git status to see what's going on!

VolodyaCO: > On branch v3.0
> Your branch is up-to-date with 'origin/v3.0'.
> 
> nothing to commit, working tree clean

It looks good

netket/graph/grid.py
gcarleo: I think that if we want to stick to the old automorphisms we need to input the edges by hand, following the same convention used in old ```Hypercube``` or, we can just stick with this strange ordering that ```grid_graph``` gives... but it will require some reshuffling of the autormophism to make sure that it is correctly sorted, as @femtobit pointed out.

gcarleo: In practice, you'd need to first sort the dict given by the networkx function with something like
```python
isos=nx.isomorphism.GraphMatcher(grid.graph, grid.graph).isomorphisms_iter()

for iso in isos:
      sorted_d = sorted(iso.items(), key=lambda x: x[1]
      iso = sorted_d
```

to be checked it this actually works... !

VolodyaCO: I found that the creation of the graph is a bit weird for the case pointed out by @femtobit. We can create an hypergrid (in this case a 1D chain) like this:

```
nx.generators.lattice.grid_graph([4], periodic=True).nodes             
Out[26]: NodeView((0, 1, 2, 3))
```
Here, the nodes are nicely ordered, and the edges are given by:
```
nx.generators.lattice.grid_graph([4], periodic=True).edges             
Out[27]: EdgeView([(0, 1), (0, 3), (1, 2), (2, 3)])
```
Now, this graph is created at `hypergrid.py`, and then a list of its edges is passed to the general-purpose Graph in `graph.py`. But this messes the order of the nodes:
```
nx.MultiGraph(nx.generators.lattice.grid_graph([4], periodic=True).edges).nodes                                                               
Out[30]: NodeView((0, 1, 3, 2))
```
The same happens if one uses `nx.Graph` instead of `nx.MultiGraph`. Basically, if one passes the list of edges `[(0, 1), (0, 3), (1, 2), (2, 3)]` one gets the ordering problem, but if one passes a list of edges `[(0, 1), (1, 2), (2, 3), (3, 0)]`, then the ordering of the nodes is correct. In both cases the edges are the same.

netket/graph/custom_graph.py
VolodyaCO: @femtobit I have implemented these errors as `TypeError`

netket/graph/grid.py
VolodyaCO: I found that the only solution was to pass to the graph creator first the nodes, and then the edges, so that the nodes are created first and in an ordered fashion. I just pushed the changes.

femtobit: Sounds good.

netket/graph/edgeless.py
```diff
@@ -0,0 +1,28 @@
...
+
+		    >>> import netket
+		    >>> g=netket.graph.Edgeless(10)
+		    >>> print(g.n_nodes)
+		    10
+    	"""
+        super().__init__([(i, i) for i in range(n_nodes)])
+        edgelessgraph = _nx.MultiGraph()
+        edgelessgraph.add_nodes_from(self.graph)
+        self.graph = edgelessgraph
```
femtobit, line 28: i don't see why this `__init__` function calls the base class constructor with one self-loop for each node in the first line and then immediately discards the `networkx` graph that `Graph.__init__` created from that edge list and assigns a different one.

That seems like it could be simplified, for example by allowing the construction of `Graph` directly from a `networkx` graph. That is in the end what the `Graph` class is now: A wrapper around a `networkx` graph (while `AbstractGraph` is the more general interface). So making it possible to directly create it that way seems logical to me.

VolodyaCO: @femtobit yes, I was thinking the same to solve the automorphisms ordering problem. However, the translation from a `Graph` to a `MultiGraph` is always necessary (to be able to represent colours). As it stands right now, the `Graph` class could have `nodes` and `edges` as arguments so that the `Edgeless` graph can be constructed by passing only `nodes`. Also, users could create graphs from `edges` only, and `hypergrid` constructors can create graphs from either `edges` or `nodes` or both. What do you think?

gcarleo: yes it seems to me that we should just have a way to convert a generic networkx graph into a ```netket.Graph``` object, then all the other cases are special instantations of this wrapper

VolodyaCO: @gcarleo Ok, so, do I write `Graph` to accept a `networkx` graph and convert it to a `MultiGraph` object to handle colours? If this is true, do I create a `CustomGraph` class that can receive `nodes` and/or `edges` for people that want to enter them explicitly?

gcarleo: @VolodyaCO : The way I see it, you should implement a single class (call it ```netket.graph.NetworkX``` or something, that takes a networkx object and wraps it into something that satisfies the ```AbstractGraph``` requirements and methods. 

Then, ```Graph``` is another class that should work accepting nodes and or edges, and basically have the same interface that it has now ```CustomGraph```. Notice however that internally you can just implement ```Graph``` as something that does two thinks:

1. Takes the edges and or nodes and creates a networkx object (call it ```n_graph```)
2. Returns ```netket.graph.NetWorkx(n_graph)```, so that all methods are automatically implemented...

gcarleo: In this sense, ```netket.graph.Graph``` is not even a class but it's just a pure function

VolodyaCO: Good, I'll do that today.

VolodyaCO: I implemented the `Graph` as a function, `Edgeless` as a function, `Hypergrid` as a class. @gcarleo , do you want me to treat `Hypercube`, `Square` and `Chain` as functions as well? Implementing them as a class is also not needed in my opinion...

gcarleo: yes it makes sens

netket/graph/graph.py
```diff
@@ -0,0 +1,166 @@
...
+
+        # For the moment, if there are colors, the method returns a NotImplementedError:
+        if self.edges():
+            colors = _np.unique(_np.array(self.edges(color=True))[:, 2])
+        else:
+            colors = _np.array([])
+        if colors.size >= 2:
+            return NotImplementedError
+
+        if self._automorphisms:
```
gcarleo, line 77:
```suggestion
        if self._automorphisms is not None:
```

netket/graph/abstract_graph.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+class AbstractGraph(abc.ABC):
+    """Abstract class for NetKet graph objects"""
+
+    @property
+    @abc.abstractmethod
+    def is_connected(self):
+        r"""bool: True if the graph is connected"""
+        return NotImplementedError
```
gcarleo, line 12: I think these should be raise NotImplementedError, also in the other files you have written

netket/graph/graph.py
```diff
@@ -0,0 +1,165 @@
...
+    def distances(self):
+        return _nx.floyd_warshall_numpy(self.graph)
+
+    @property
+    def is_bipartite(self):
+        return _nx.is_bipartite(self.graph)
+
+    @property
+    def size(self):
+        print("DeprecationWarning: size is deprecated. Use n_nodes from now on.")
```
gcarleo, line 58: there is a standard way to write deprecations, @femtobit maybe can tell you more about it

netket/graph/graph.py
```diff
@@ -0,0 +1,165 @@
...
+        # TODO: check how to compute these when we have a coloured graph where there could
+        #       be a duplicated edge with two different colors.
+
+        # For the moment, if there are colors, the method returns a NotImplementedError:
+        if self.edges():
+            colors = _np.unique(_np.array(self.edges(color=True))[:, 2])
+        else:
+            colors = _np.array([])
+        if colors.size >= 2:
+            return NotImplementedError
```
gcarleo, line 75: raise

netket/graph/abstract_graph.py
VolodyaCO: > Thanks! we are getting very close! There is a few microscopic changes + please also try to install an automatic black linter for the code otherwise the tests won't pass!
> 
> ```
> 2.86s$ python -m black --check ../netket/
> would reformat /home/travis/build/netket/netket/netket/graph/hypergrid.py
> would reformat /home/travis/build/netket/netket/netket/graph/graph.py
> Oh no! 💥 💔 💥
> 2 files would be reformatted, 75 files would be left unchanged.
> ```

Oh!, I forgot about it completely, sorry. I have installed black already!

netket/graph/graph.py
femtobit: Is there a need to deprecate it, though? `graph.size` did not exist in NetKet v2 if I recall correctly, so we can just remove it.

VolodyaCO: If I understand correctly, size is being used elsewhere in the code many times. I only added the warning so that existing code runs but points out where the size property is being used. But sure... one can just replace size by n_nodes in the other parts of the code.

gcarleo: ok then just remove size, I will then do another PR making sure that size is not used in other places when dealing with the new graph objects

gcarleo:
@femtobit further changes needed?

netket/graph/graph.py
```diff
@@ -0,0 +1,166 @@
...
+            >>> nk_g = netket.graph.NetworkX(nx_g)
+            >>> print(nk_g.n_nodes)
+            3
+        """
+        assert isinstance(graph, _nx.classes.graph.Graph) or isinstance(
+            graph, _nx.classes.multigraph.MultiGraph
+        )
+
+        if isinstance(graph, _nx.classes.graph.Graph):
+            self.graph = _nx.MultiGraph(graph)
```
femtobit, line 30: What if `graph` is a multi-graph? Then `self.graph` is never assigned, isn't it?

netket/graph/graph.py
```diff
@@ -0,0 +1,166 @@
...
+        super().__init__()
+
+    @property
+    def adjacency_list(self):
+        return [list(self.graph.neighbors(node)) for node in self.graph.nodes]
+
+    @property
+    def is_connected(self):
+        # TODO: how to check if a multigraph is connected?
+        return _nx.is_connected(self.graph)
```
femtobit, line 43: Is this TODO resolved?

netket/graph/graph.py
```diff
@@ -0,0 +1,166 @@
...
+        # TODO: check how to compute these when we have a coloured graph where there could
+        #       be a duplicated edge with two different colors.
+
+        # For the moment, if there are colors, the method returns a NotImplementedError:
+        if self.edges():
+            colors = _np.unique(_np.array(self.edges(color=True))[:, 2])
+        else:
+            colors = _np.array([])
+        if colors.size >= 2:
+            raise NotImplementedError
```
femtobit, line 74: I think this should have an error message.
```suggestion
            raise NotImplementedError("automorphisms is not yet implemented for colored edges")
```

netket/graph/graph.py
```diff
@@ -0,0 +1,166 @@
...
+    edges_array = _np.array(edges, dtype=_np.int32)
+    if edges_array.ndim != 2:
+        raise TypeError(
+            "edges must be a list of lists or tuples of the same length (2 or 3)"
+        )
+
+    if not (edges_array.shape[1] == 2 or edges_array.shape[1] == 3):
+        raise TypeError(
+            "edges must be a list of lists or tuples of the same length (2 or 3), where the third column indicates the color"
+        )
```
femtobit, line 134: I think these should be `ValueError`s.

netket/graph/hypergrid.py
```diff
@@ -0,0 +1,132 @@
...
+            assert isinstance(length, list)
+        except AssertionError:
+            raise TypeError("length must be a list of integers >= 1")
+
+        try:
+            condition = [isinstance(x, int) and x >= 1 for x in length]
+            if sum(condition) != len(length):
+                raise TypeError
+        except TypeError:
+            raise TypeError("Components of length must be integers greater than 1")
```
femtobit, line 45: Same as above.
This doesn't even change the type of the error. Just raise the exception below the if, not need to catch and re-raise. Also, the contents of the lists being wrong should bey a `ValueError`.

netket/graph/hypergrid.py
```diff
@@ -0,0 +1,132 @@
...
+            Also, a 2x2x3 lattice with open boundary conditions can be constructed as follows:
+            >>> g=netket.graph.Hypergrid(length=[2,2,3], pbc=False)
+            >>> print(g.n_nodes)
+            12
+        """
+
+        try:
+            assert isinstance(length, list)
+        except AssertionError:
+            raise TypeError("length must be a list of integers >= 1")
```
femtobit, line 38: This is a slightly strange way of writing the check. How about
```suggestion
    if not isinstance(length, list):
        raise TypeError("length must be a list of integers")
```

gcarleo:
> What about the unit tests? They still test the C++ graphs as far as I can see, since netket/graph.py is still the what is imported and used, not the files in netket/graph/.

Yes I am going to remove the c++ graphs at a later stage and rename these removing the Py prexif, so that testing will be OK

netket/graph/hypergrid.py
```diff
@@ -0,0 +1,132 @@
+from graph import NetworkX
+
+import numpy as _np
+import networkx as _nx
+
+
+class Hypergrid(NetworkX):
```
femtobit, line 7: Can we just call it `Grid`? I think there is no need to add "hyper", a grid can be any dimension. (For `Hypercube` it makes sense, as cube by itself is 3d.)

femtobit:
Ah, and there are still separate `grid.py` and `hypergrid.py` files in the PR.

netket/graph/graph.py
VolodyaCO: Yes

netket/graph/hypergrid.py
VolodyaCO: Sorry, forgot to change those.

netket/graph/graph.py
gcarleo: can you remove the TODO then? :)

netket/graph/hypergrid.py
VolodyaCO: I named it `Grid` now

gcarleo:
@femtobit are we good to go? Still for the moment this is not replacing our graphs, I will do a further PR removing our graphs and adding these new versions to the tests

PR #408, opened 2020-05-06
Stochastic Reconfiguration in Jax

chrisrothUT:
Hi everyone,

I fixed the code so that stochastic reconfiguration works with Jax. Right now the only SR solver that is implemented with Jax types is "jaxcg" which does conjugate gradient with Jax. I'll post a speed comparison shortly.

Unfortunately, the exact solvers in Jax are causing SIGBUS errors for systems of more than 10 electrons, so I have left their SciPy implementations currently. 

Now that Jax is working on CPUs, I will try to get it working on a GPU. Soon we may be able to train deep networks with stochastic reconfiguration.

### DIFF
--- a/netket/_vmc.py
+++ b/netket/_vmc.py
@@ -11,6 +11,7 @@ from netket.stats import (
 
 from netket.vmc_common import info, tree_map
 from netket.abstract_variational_driver import AbstractVariationalDriver
+import jax
 
 
 class Vmc(AbstractVariationalDriver):
@@ -145,12 +146,19 @@ class Vmc(AbstractVariationalDriver):
             self._grads, self._jac = self._machine.vector_jacobian_prod(
                 samples_r, eloc_r / self._n_samples, self._grads, return_jacobian=True
             )
-            _sum_inplace(self._grads)
+            
+            self._grads = tree_map(_sum_inplace, self._grads)
+            
+            self._grads = self._machine.jax_flatten(self._grads)
+            self._jac = self._machine.jacobian_flatten(self._jac)
 
             # Center the log derivatives
-            self._jac -= _mean(self._jac, axis=0)
+            # self._jac -= _mean(self._jac, axis=0)
+            self._jac -= jax.numpy.mean(self._jac,axis=0)
 
             self._dp = self._sr.compute_update(self._jac, self._grads, self._dp)
+
+            self._dp = self._machine.unflatten(self._dp, self._machine._params)
 
         else:
             # Computing updates using the simple gradient
--- a/netket/_vmc.py
+++ b/netket/_vmc.py
@@ -12,9 +12,6 @@ from netket.stats import (
 from netket.vmc_common import info, tree_map
 from netket.abstract_variational_driver import AbstractVariationalDriver
 
-from netket.utils import jax_available
-if jax_available:
-    from netket.vmc_common import shape_for_sr, shape_for_update
 
 class Vmc(AbstractVariationalDriver):
     """
@@ -65,6 +62,7 @@ class Vmc(AbstractVariationalDriver):
         self._sr = sr
         if sr is not None:
             self._sr.is_holomorphic = sampler.machine.is_holomorphic
+            self._sr.machine = sampler.machine
 
         self._npar = self._machine.n_par
 
@@ -150,11 +148,8 @@ class Vmc(AbstractVariationalDriver):
             )
             
             self._grads = tree_map(_sum_inplace, self._grads)
-            self._grads, self._jac = shape_for_sr(self._grads,self._jac)
-            
-            self._dp = self._sr.compute_update(self._jac, self._grads, self._dp)
 
-            self._dp = shape_for_update(self._dp,self._machine.parameters) 
+            self._dp = self._sr.compute_update(self._jac, self._grads, self._dp)
 
         else:
             # Computing updates using the simple gradient
--- a/netket/_vmc.py
+++ b/netket/_vmc.py
@@ -146,7 +146,7 @@ class Vmc(AbstractVariationalDriver):
             self._grads, self._jac = self._machine.vector_jacobian_prod(
                 samples_r, eloc_r / self._n_samples, self._grads, return_jacobian=True
             )
-            
+
             self._grads = tree_map(_sum_inplace, self._grads)
 
             self._dp = self._sr.compute_update(self._jac, self._grads, self._dp)
@@ -156,7 +156,7 @@ class Vmc(AbstractVariationalDriver):
             self._grads = self._machine.vector_jacobian_prod(
                 samples_r, eloc_r / self._n_samples, self._grads
             )
-            
+
             self._grads = tree_map(_sum_inplace, self._grads)
 
             self._dp = self._grads
--- a/netket/_vmc.py
+++ b/netket/_vmc.py
@@ -11,8 +11,10 @@ from netket.stats import (
 
 from netket.vmc_common import info, tree_map
 from netket.abstract_variational_driver import AbstractVariationalDriver
-import jax
 
+from netket.utils import jax_available
+if jax_available:
+    from netket.vmc_common import shape_for_sr, shape_for_update
 
 class Vmc(AbstractVariationalDriver):
     """
@@ -148,24 +150,18 @@ class Vmc(AbstractVariationalDriver):
             )
             
             self._grads = tree_map(_sum_inplace, self._grads)
+            self._grads, self._jac = shape_for_sr(self._grads,self._jac)
             
-            self._grads = self._machine.jax_flatten(self._grads)
-            self._jac = self._machine.jacobian_flatten(self._jac)
-
-            # Center the log derivatives
-            # self._jac -= _mean(self._jac, axis=0)
-            self._jac -= jax.numpy.mean(self._jac,axis=0)
-
             self._dp = self._sr.compute_update(self._jac, self._grads, self._dp)
 
-            self._dp = self._machine.unflatten(self._dp, self._machine._params)
+            self._dp = shape_for_update(self._dp,self._machine.parameters) 
 
         else:
             # Computing updates using the simple gradient
             self._grads = self._machine.vector_jacobian_prod(
                 samples_r, eloc_r / self._n_samples, self._grads
             )
-
+            
             self._grads = tree_map(_sum_inplace, self._grads)
 
             self._dp = self._grads
--- a/netket/machine/jax.py
+++ b/netket/machine/jax.py
@@ -55,7 +55,7 @@ class Jax(AbstractMachine):
 
         forward_scalar = jax.jit(lambda pars, x: self._forward_fn(pars, x).reshape(()))
         grad_fun = jax.jit(jax.grad(forward_scalar, holomorphic=self.is_holomorphic))
-        self._perex_grads = jax.jit(jax.vmap(grad_fun, in_axes=(None, 0)))
+        self._perex_grads = jax.jit(jax.vmap(grad_fun, in_axes=(None,0)))
 
         self.init_random_parameters()
 
@@ -149,14 +149,19 @@ class Jax(AbstractMachine):
                 out = tree_map(jax.numpy.conjugate, pout[0])
 
             return out
+
         else:
-            if conjugate and self._dtype is complex:
-                prodj = lambda j: jax.np.tensordot(vec, j.conjugate(), axes=vec.ndim)
-            else:
-                prodj = lambda j: jax.np.tensordot(vec.conjugate(), j, axes=vec.ndim)
+
+            vals, f_jvp = jax.vjp(
+                self._forward_fn, self._params, x.reshape((-1, x.shape[-1]))
+            )
+
+            pout = f_jvp(vec.reshape(vals.shape).conjugate())
+            out = self.jax_flatten(pout[0])
 
             jacobian = self._perex_grads(self._params, x)
-            out = tree_map(prodj, jacobian)
+            jacobian = self.jacobian_flatten(jacobian)
+
             return out, jacobian
 
     @property
@@ -182,6 +187,24 @@ class Jax(AbstractMachine):
 
         assert npar == self._npar
 
+    def jacobian_flatten(self,jac):
+        r"""Flattens the jacobian to be shape [-1,n_params] 
+
+        Args:
+             data: a (possibly non-flat) structure containing jax arrays.
+
+        Returns:
+             jax.numpy.ndarray: a 2-dimensional array corresponding to the
+             Jacobian of the examples w.r.t. the parameters
+        """
+
+        return jax.numpy.concatenate(tuple(fd.reshape(len(fd),-1) for fd in tree_flatten(jac)[0]),-1)
+
+
+
+    def jax_flatten(self,data):
+        return jax.numpy.concatenate(tuple(fd.reshape(-1) for fd in tree_flatten(data)[0]))
+
     def numpy_flatten(self, data):
         r"""Returns a flattened numpy array representing the given data.
             This is typically used to serialize parameters and gradients.
@@ -192,9 +215,10 @@ class Jax(AbstractMachine):
         Returns:
              numpy.ndarray: a one-dimensional array containing a copy of data
         """
+
         return _np.concatenate(tuple(fd.reshape(-1) for fd in tree_flatten(data)[0]))
 
-    def numpy_unflatten(self, data, shape_like):
+    def unflatten(self, data, shape_like):
         r"""Attempts a deserialization of the given numpy data.
             This is typically used to deserialize parameters and gradients.
 
--- a/netket/machine/jax.py
+++ b/netket/machine/jax.py
@@ -157,10 +157,11 @@ class Jax(AbstractMachine):
             )
 
             pout = f_jvp(vec.reshape(vals.shape).conjugate())
-            out = self.jax_flatten(pout[0])
+
+            if conjugate and self._dtype is complex:
+                out = tree_map(jax.numpy.conjugate, pout[0])
 
             jacobian = self._perex_grads(self._params, x)
-            jacobian = self.jacobian_flatten(jacobian)
 
             return out, jacobian
 
@@ -187,24 +188,6 @@ class Jax(AbstractMachine):
 
         assert npar == self._npar
 
-    def jacobian_flatten(self,jac):
-        r"""Flattens the jacobian to be shape [-1,n_params] 
-
-        Args:
-             data: a (possibly non-flat) structure containing jax arrays.
-
-        Returns:
-             jax.numpy.ndarray: a 2-dimensional array corresponding to the
-             Jacobian of the examples w.r.t. the parameters
-        """
-
-        return jax.numpy.concatenate(tuple(fd.reshape(len(fd),-1) for fd in tree_flatten(jac)[0]),-1)
-
-
-
-    def jax_flatten(self,data):
-        return jax.numpy.concatenate(tuple(fd.reshape(-1) for fd in tree_flatten(data)[0]))
-
     def numpy_flatten(self, data):
         r"""Returns a flattened numpy array representing the given data.
             This is typically used to serialize parameters and gradients.
@@ -218,7 +201,7 @@ class Jax(AbstractMachine):
 
         return _np.concatenate(tuple(fd.reshape(-1) for fd in tree_flatten(data)[0]))
 
-    def unflatten(self, data, shape_like):
+    def numpy_unflatten(self, data, shape_like):
         r"""Attempts a deserialization of the given numpy data.
             This is typically used to deserialize parameters and gradients.
 
@@ -241,7 +224,6 @@ class Jax(AbstractMachine):
             k += size
 
         return tree_unflatten(tree, datalist)
-
 
 from jax.experimental import stax
 from jax.experimental.stax import Dense
@@ -267,10 +249,10 @@ def logcosh(x):
 
 LogCoshLayer = stax.elementwise(logcosh)
 
-
 def JaxRbm(hilbert, alpha, dtype=complex):
     return Jax(
         hilbert,
-        stax.serial(stax.Dense(alpha * hilbert.size), LogCoshLayer, SumLayer()),
+        stax.serial(stax.Dense(alpha * hilbert.size),LogCoshLayer, SumLayer()),
         dtype=dtype,
     )
+
--- a/netket/machine/jax.py
+++ b/netket/machine/jax.py
@@ -22,7 +22,6 @@ import numpy as _np
 from netket.random import randint as _randint
 from jax.tree_util import tree_flatten, tree_unflatten, tree_map
 
-
 class Jax(AbstractMachine):
     def __init__(self, hilbert, module, dtype=complex):
         """
@@ -152,18 +151,16 @@ class Jax(AbstractMachine):
 
         else:
 
-            vals, f_jvp = jax.vjp(
-                self._forward_fn, self._params, x.reshape((-1, x.shape[-1]))
-            )
-
-            pout = f_jvp(vec.reshape(vals.shape).conjugate())
-
             if conjugate and self._dtype is complex:
-                out = tree_map(jax.numpy.conjugate, pout[0])
+                prodj = lambda j: jax.np.tensordot(vec.transpose(), j.conjugate(), axes=1)
+            else:
+                prodj = lambda j: jax.np.tensordot(vec.transpose().conjugate(), j, axes=1)
 
             jacobian = self._perex_grads(self._params, x)
-
+            out = tree_map(prodj, jacobian)
+            
             return out, jacobian
+
 
     @property
     def is_holomorphic(self):
--- a/netket/machine/jax.py
+++ b/netket/machine/jax.py
@@ -22,6 +22,7 @@ import numpy as _np
 from netket.random import randint as _randint
 from jax.tree_util import tree_flatten, tree_unflatten, tree_map
 
+
 class Jax(AbstractMachine):
     def __init__(self, hilbert, module, dtype=complex):
         """
@@ -54,7 +55,7 @@ class Jax(AbstractMachine):
 
         forward_scalar = jax.jit(lambda pars, x: self._forward_fn(pars, x).reshape(()))
         grad_fun = jax.jit(jax.grad(forward_scalar, holomorphic=self.is_holomorphic))
-        self._perex_grads = jax.jit(jax.vmap(grad_fun, in_axes=(None,0)))
+        self._perex_grads = jax.jit(jax.vmap(grad_fun, in_axes=(None, 0)))
 
         self.init_random_parameters()
 
@@ -152,15 +153,18 @@ class Jax(AbstractMachine):
         else:
 
             if conjugate and self._dtype is complex:
-                prodj = lambda j: jax.np.tensordot(vec.transpose(), j.conjugate(), axes=1)
+                prodj = lambda j: jax.np.tensordot(
+                    vec.transpose(), j.conjugate(), axes=1
+                )
             else:
-                prodj = lambda j: jax.np.tensordot(vec.transpose().conjugate(), j, axes=1)
+                prodj = lambda j: jax.np.tensordot(
+                    vec.transpose().conjugate(), j, axes=1
+                )
 
             jacobian = self._perex_grads(self._params, x)
             out = tree_map(prodj, jacobian)
-            
-            return out, jacobian
 
+            return out, jacobian
 
     @property
     def is_holomorphic(self):
@@ -222,6 +226,7 @@ class Jax(AbstractMachine):
 
         return tree_unflatten(tree, datalist)
 
+
 from jax.experimental import stax
 from jax.experimental.stax import Dense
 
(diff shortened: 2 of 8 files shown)


netket/machine/jax.py
```diff
@@ -182,6 +187,24 @@ def parameters(self, p):
...
+        return jax.numpy.concatenate(tuple(fd.reshape(len(fd),-1) for fd in tree_flatten(jac)[0]),-1)
+
+
+
+    def jax_flatten(self,data):
+        return jax.numpy.concatenate(tuple(fd.reshape(-1) for fd in tree_flatten(data)[0]))
+
     def numpy_flatten(self, data):
         r"""Returns a flattened numpy array representing the given data.
             This is typically used to serialize parameters and gradients.
```
femtobit, line 210: Looking at all the flatten/unflatten functions here: Is there a reason they a part of the machine class? They do not depend on any state of the machine (as they don't even access self) as far as I can see, and thus would make a lot more sense as free-standing functions (which is also more easily reusable, no need to create machine just to use them). Or am I missing something?

netket/_vmc.py
```diff
@@ -11,6 +11,7 @@
 
 from netket.vmc_common import info, tree_map
 from netket.abstract_variational_driver import AbstractVariationalDriver
+import jax
```
gcarleo, line 14: jax should never be imported directly but only if available, see how it is imported in other modules through netket.utils.has_jax

netket/_vmc.py
```diff
@@ -145,13 +146,20 @@ def _forward_and_backward(self):
             self._grads, self._jac = self._machine.vector_jacobian_prod(
                 samples_r, eloc_r / self._n_samples, self._grads, return_jacobian=True
             )
-            _sum_inplace(self._grads)
+            
+            self._grads = tree_map(_sum_inplace, self._grads)
```
gcarleo, line 150: this breaks all other cases in which we are not using jax machines, is that correct?

netket/machine/jax.py
```diff
@@ -149,14 +149,19 @@ def vector_jacobian_prod(
                 out = tree_map(jax.numpy.conjugate, pout[0])
 
             return out
+
         else:
-            if conjugate and self._dtype is complex:
```
gcarleo, line 153: what happened to the conjugate option?

netket/machine/jax.py
gcarleo: those are methods that each machine is supposed to use to convert their internal types to flattened numpy array, so in principle each machine can have a different implementation... it's not optimal, but I was using that only for testing basically

netket/optimizer/stochastic_reconfiguration.py
```diff
@@ -8,6 +8,9 @@
 from scipy.sparse.linalg import cg, gmres, minres
 from mpi4py import MPI
 
+import jax
```
gcarleo, line 11: jax direct importing to be avoided

netket/optimizer/stochastic_reconfiguration.py
```diff
@@ -69,6 +72,8 @@ def _init_solver(self):
                         "minres can be used only for real-valued parameters."
                     )
                 self._sparse_solver = minres
+            elif lsq_solver == "jaxcg":
+                self._is_holomorphic = True
```
gcarleo, line 76: why? what if it's not holomorphic?

netket/machine/jax.py
chrisrothUT: No reason at all, just my general coding inexperience :). I agree that these functions can be more general

netket/_vmc.py
chrisrothUT: Yes, I will fix that and make sure other cases work with the updated code

netket/machine/jax.py
chrisrothUT: Oops, will fix that

chrisrothUT:
Comparison between CG using Jax vs Scipy (time in s)

Params  Jax  Scipy 
420 0.1751 0.0234
2100 0.2465 0.1035 
5500 0.4656 0.2353

As you can see the Jax version is much slower for small systems but catches up as system size gets larger.

netket/machine/jax.py
femtobit: Yeah, for testing this is fine of course. I think we should do something like 1) extract the `jax_flatten` and others as utility functions to some other place, as they are probably generally useful, not just in a context where we have a machine accessible, 2) leave a pair of abstract `flatten`/`unflatten` methods on the machine, which call the appropriate utility functions for the machine type.

netket/optimizer/stochastic_reconfiguration.py
```diff
@@ -199,6 +208,39 @@ def compute_update(self, oks, grad, out=None):
...
+    def _jax_cg_solve(self,oks,grad,n_samp):
+        """
+        Solves the SR flow equation using the conjugate gradient method 
+        """
+
+        n_par = grad.shape[0]
+        if self._x0 is None:
+            self._x0 = _np.zeros(n_par, dtype=_np.complex128)
+
+        cov_op = self._jax_linear_function(oks,n_samp)
```
gcarleo, line 220: @chrisrothUT have you tried to jit this function to see if you have performance improvements? 
say
```python
cov_op = jax.jit(self._jax_linear_function(oks,n_samp))
```

chrisrothUT:
Okay I have figured out part of the issue. It looks like both params and grads are stored as 1D arrays for numpy machines where as they are stored as trees for Jax machines. So we only need to transform from tree to array in jax. 

Since VMC should be type agnostic perhaps we should have a machine property that tells the array type (numpy, jax, torch), then we can call the appropriate helper functions to reshape for SR. What do you guys think?

netket/optimizer/stochastic_reconfiguration.py
chrisrothUT: I'll give that a try

gcarleo:
> Since _vmc should be type agnostic perhaps we should add a machine property that tells the array type (numpy, jax, torch), then if we need to restructure the data for SR we can call the appropriate helper function. What do you guys think?

@chrisrothUT yes you touched upon a very good point here.
This issue is a design decision we have been discussing for some time, and it's unclear what the best solution is. For the moment, we have decided not to include the array type directly as a machine property, but rather have all the reduction functions needed in the VMC (and I think also in SR) work with all types by means of an internal automatic dispatch system. 

If you look at the current implementation of the vmc (the part without SR) this operation here: 
```python
self._grads = tree_map(_sum_inplace, self._grads)
```

is intrinsically type agnostic, because we have implemented _sum_inplace to work with both numpy and jax types. In practice, I would put the same call also in the SR branch, and maybe move this 
```python
# Center the log derivatives
self._jac -= _mean(self._jac, axis=0)
```

directly into the SR driver (I.e. I would pass the bare jacobian to the SR driver, and let it subtract the mean internally, using its own types and/or using type agnostic functions such as _mean (I am not sure _mean is already type agnostic though, but we can make it...).

chrisrothUT:
@gcarleo Yes that makes sense. So in _vmc I can do two things.

1. Only do the reshaping for SR gradients if jax_available
2. Define a function in vmc_common that returns the identity for numpy and returns the reshaped objects for jax

Obviously 2 calls an unnecessary function for numpy but it seems more in the style of what you recommended

netket/_vmc.py
chrisrothUT: I was wrong, this doesn't things break things because the function tree_map returns _sum_inplace(self._grads) if self._grads is an array

femtobit: Indeed. `tree_map` operates on a jax-like pytree structure where the leaves are arrays (or jax DeviceArrays etc.). A single leave is a valid input, so `tree_map` should work with it, otherwise that'd be a bug in `tree_map`.

netket/optimizer/stochastic_reconfiguration.py
chrisrothUT: For some reason that actually makes performance way worse (by a factor of 2/3). Maybe it is compiling every time?

gcarleo: yes, that's possible... you should try to jit this only once at construction times, but that might have some side effects if not done properly uhm

chrisrothUT:
Okay I've made some edits:

1. Now JaxSR is a separate optimizer from SR. JaxSR consists of all of the SciPy optimizers plus "jaxcg" which is the default option. 

2. Added two functions to vmc_common which take the grad, jacobian from tree->array and take the SR update back to a tree. For numpy these functions just return their inputs. These are called in _vmc

3. Moved the centering of the jacobian from _vmc to stochastic_reconfiguration

I've checked that SR and SGD work for jax and numpy networks.

gcarleo:
@chrisrothUT if you look at the tests, you see that there is some issue with indentation 
```
E     File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/netket/optimizer/stochastic_reconfiguration.py", line 110
E       oks -= _mean(oks,axis=0)
E                              ^
E   TabError: inconsistent use of tabs and spaces in indentation
```

can you please have a look at our guidelines [on coding style](https://github.com/netket/netket/blob/v3.0/CONTRIBUTING.md#python-coding-style)?

netket/_vmc.py
```diff
@@ -145,19 +148,20 @@ def _forward_and_backward(self):
...
                 samples_r, eloc_r / self._n_samples, self._grads, return_jacobian=True
             )
-            _sum_inplace(self._grads)
-
-            # Center the log derivatives
-            self._jac -= _mean(self._jac, axis=0)
-
+            
+            self._grads = tree_map(_sum_inplace, self._grads)
+            self._grads, self._jac = shape_for_sr(self._grads,self._jac)
```
gcarleo, line 153: Would it be possible to adsorb these two function calls ```shape_for_sr``` and ```shape_for_update``` directly into the Jax SR solver? 

This way, you could also remove any reference to Jax in this file, i.e. the lines 
```python
from netket.utils import jax_available
if jax_available:
    from netket.vmc_common import shape_for_sr, shape_for_update
```

netket/machine/jax.py
```diff
@@ -149,14 +149,20 @@ def vector_jacobian_prod(
...
         else:
+
+            vals, f_jvp = jax.vjp(
+                self._forward_fn, self._params, x.reshape((-1, x.shape[-1]))
+            )
+
+            pout = f_jvp(vec.reshape(vals.shape).conjugate())
+
             if conjugate and self._dtype is complex:
-                prodj = lambda j: jax.np.tensordot(vec, j.conjugate(), axes=vec.ndim)
```
gcarleo, line 154: I put these because I didn't want to call f_jvp here explicitly, when the jacobian is passed to the function (otherwise you do more work). It'd be good to revert the changes to this function

chrisrothUT: Okay true, I can write it this way

netket/_vmc.py
chrisrothUT: This would be tricky because SR doesn't have access to the shape of the parameter tree. I guess I could have flatten return the shape and use it for unflatten

gcarleo: but in the Jax SR you can just use the native ```flatten``` and ```unflatten``` from Jax that should allow you to pack and unpack at will! Actually I was just looking at what Jax does with optimizers, and basically it seems that they convert these trees into flat arrays, if I understand correctly [this](https://github.com/google/jax/blob/f60184e12e279c4602c77f91be19a0c6e1eb7083/jax/experimental/optimizers.py#L100)

gcarleo: the native things are 
```python
from jax.tree_util import (tree_map, tree_flatten, tree_unflatten)
```

chrisrothUT: Yeah, for some reason tree_flatten doesn't actually return a flat arrays, all it seems to do is concatenate all of the parameters into a list instead of nested lists (i.e. before flatten the params of each layer are in their own branch). That's why numpy_flatten in "netket/machine/jax.py" still has to iterate even after calling tree_flatten. I was also confused by this

gcarleo: But it seems to me that you can still easily move these two 

```python 
shape_for_sr(self._grads,self._jac)

shape_for_update(self._dp,self._machine.parameters) 
```

in the JaxSr class. In general the rationale is that the JaxSr class is supposed to work with Jax types, and that the Sr class we have now is supposed to work only with numpy types. Later on we will think about an automatic dispatch system...

netket/optimizer/jax_stochastic_reconfiguration.py
```diff
@@ -0,0 +1,370 @@
...
+from scipy.linalg import cho_factor as _cho_factor
+from scipy.linalg import cho_solve as _cho_solve
+from scipy.sparse.linalg import LinearOperator
+from netket.stats import sum_inplace as _sum_inplace
+from scipy.sparse.linalg import cg, gmres, minres
+from mpi4py import MPI
+
+from netket.utils import jax_available
+
+if jax_available:
```
gcarleo, line 13: I would check if jax is available in the __init__.py of the optimizer module, and then if available import jax_stochastic_reconfiguration, so that in this file you are already working with the assumption that jax is there and don't need to check again

netket/optimizer/jax_stochastic_reconfiguration.py
```diff
@@ -0,0 +1,370 @@
...
+        O_i(v_j) = ∂/∂x_i log Ψ(v_j) and f is a generalized force (the loss
+        gradient).
+
+        Args:
+            oks: The matrix of log-derivatives,
+		O_i(v_j)
+            grad: The vector of forces f.
+            out: Output array for the update ẋ.
+        """
+
```
gcarleo, line 114: basicaly, you can put here the first call you have now in the vmc that prepares the oks and the grad to the right format

netket/optimizer/jax_stochastic_reconfiguration.py
```diff
@@ -0,0 +1,370 @@
...
+                        x0=self._x0,
+                        tol=self.sparse_tol,
+                        maxiter=self.sparse_maxiter,
+                    )
+                    if info < 0:
+                        raise RuntimeError("SR sparse solver did not converge.")
+
+                self._x0 = out
+            else:
+                self._S = _np.matmul(oks.conj().T, oks, self._S)
```
gcarleo, line 151: shouldn't we use jax.numpy here (and in all the other cases) instead of standard numpy ?

netket/optimizer/jax_stochastic_reconfiguration.py
```diff
@@ -0,0 +1,370 @@
...
+
+        if out is None:
+            out = _np.zeros(n_par, dtype=_np.complex128)
+
+        if self._is_holomorphic:
+            if self._use_iterative:
+                if self._lsq_solver == "jaxcg":
+                    out = self._jax_cg_solve(oks,grad,n_samp)
+                else:
+                    op = self._linear_operator(oks, n_samp)
```
gcarleo, line 134: I think we can remove this branch that doesn't use jax solvers

netket/optimizer/jax_stochastic_reconfiguration.py
```diff
@@ -0,0 +1,370 @@
...
+
+                    if self._x0 is None:
+                        self._x0 = _np.zeros(n_par, dtype=_np.complex128)
```
gcarleo, line 137: also here pure numpy types should not be used

netket/optimizer/jax_stochastic_reconfiguration.py
```diff
@@ -0,0 +1,370 @@
...
+                    )
+
+                self._revert_preconditioning(out)
+
+        else:
+            if self._use_iterative:
+                if self._lsq_solver == "jaxcg":
+                    out.real = self._jax_cg_solve(oks,grad.real,n_samp)
+                else:
+                    op = self._linear_operator(oks, n_samp)
```
gcarleo, line 176: here the non-jax branch is not really nedeed anymore

netket/optimizer/jax_stochastic_reconfiguration.py
```diff
@@ -0,0 +1,370 @@
...
+        if self._x0 is None:
+            self._x0 = _np.zeros(n_par, dtype=_np.complex128)
+
+        cov_op = jax.jit(self._jax_linear_function(oks,n_samp))
+        
+        out, _ = jcg(cov_op,grad,x0=self._x0,tol=self.sparse_tol,maxiter=self.sparse_maxiter)
+
+        return out
+
+    def _jax_linear_function(self,oks,n_samp):
```
gcarleo, line 231: To avoid recompiling at each call of the SR, I think that the jit decorator should be put here, and not when definining cov_op above. 

Basically 

```python
@partial(jax.jit, static_argnums=(0, 2))
def _jax_linear_function(self,oks,n_samp)
```

netket/_vmc.py
chrisrothUT: Okay, I did this. Since SR doesn't have access to "self._machine.parameters", shape_for_sr now returns a copy of grads in order to shape back to Pytree. Since grads consumes. memory 1/batch size compared to Jac this didn't seem like a terrible solution but let me know if there's something better. I would return just the shapes, but tree_unflatten needs an object "shape_like"

netket/optimizer/jax_stochastic_reconfiguration.py
chrisrothUT: Right, is it okay to call the raw jacobian oks as well? I don't want to use two names to avoid copying memory.

netket/optimizer/jax_stochastic_reconfiguration.py
chrisrothUT: Okay, do we want to keep the scipy exact solvers though? Right now the Jax ones don't appear to be working

netket/optimizer/jax_stochastic_reconfiguration.py
gcarleo: yes it's fine

netket/_vmc.py
gcarleo: yes I see now the issue
Then the solution is just to construct Sr passing the machine explicitly in its __init__ or in vmc there is these lines 
```python
        if sr is not None:
            self._sr.is_holomorphic = sampler.machine.is_holomorphic
```
where you can add something like 

```python
        if sr is not None:
            self._sr.is_holomorphic = sampler.machine.is_holomorphic
            self._sr.machine=sampler.machine 
```

netket/optimizer/jax_stochastic_reconfiguration.py
chrisrothUT: _jax_linear_function returns the function matvec(cov,x) so you can't jit it. I tried jitting the matvec function inside and it's still slower (also no difference between step 1 and 2 so I think its compiling every time). 

I'll keep thinking about it

chrisrothUT: There is also this thing: 

https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.custom_linear_solve.html#jax.lax.custom_linear_solve

It looks like this leverages jax's autodiff capabilities.

chrisrothUT: Also here's the issue where they're discussing the linear solvers
https://github.com/google/jax/issues/1531

gcarleo: but we don't really need a differentiable solver, we just need a solver that works with jax types (thus possibly fast on a gpu) without converting to numpy! In any case we can leave the performance issue aside for the moment, we can fix this later

netket/optimizer/jax_stochastic_reconfiguration.py
gcarleo: is there a bug in their non-sparse solvers? Can you reproduce it with a simple example?

chrisrothUT:
Okay, I'm going to change JaxSR to only work with Jax types and Jax solvers. If non-jax solvers end up being faster for something, SR will work with Jax types as soon as the _mean is fixed to work with Jax

netket/_vmc.py
chrisrothUT: Okay great, this works.

netket/optimizer/jax_stochastic_reconfiguration.py
chrisrothUT: Yes its a bug with their non-sparse solvers. As an example for cho_factor, this script returns a bus error

import numpy as np
import jax.numpy as jnp
from jax.scipy.linalg import cho_factor

vec = jnp.array(np.random.normal(0.,0.1,[20,400]) + 1j*np.random.normal(0.,0.1,[20,400]))

matrix = jnp.matmul(vec.transpose(),vec) + jnp.array(np.identity(400))

print(cho_factor(matrix))

chrisrothUT: The threshold for breaking is slightly different for all of the solvers but a 400x400 matrix breaks all of them so they're useless at the moment

gcarleo: But this specific case breaks also on scipy, as far as I can tell 
```
 raise LinAlgError("%d-th leading minor of the array is not positive "
numpy.linalg.LinAlgError: 29-th leading minor of the array is not positive definite
```

are you sure there isn't something else going wrong?

gcarleo: I mean the issue is that you didn't put the conjugate, so that's not even a positive-definite matrix

```
matrix = jnp.matmul(vec.transpose().conjugate(),vec) + jnp.array(np.identity(400))
```
this works

chrisrothUT: Sorry my mistake. This code breaks in jax but not numpy

import numpy as np
import jax.numpy as jnp
from jax.scipy.linalg import cho_factor

vec = jnp.array(np.random.normal(0.,0.1,[20,400]) + 1j*np.random.normal(0.,0.1,[20,400]))

matrix = jnp.matmul(vec.transpose().conjugate(),vec) + jnp.array(np.identity(400))

print(cho_factor(matrix))

chrisrothUT: Numpy code that works:

import numpy as np
from scipy.linalg import cho_solve, cho_factor

vec = np.random.normal(0.,0.1,[20,400]) + 1j*np.random.normal(0.,0.1,[20,400])

matrix = np.matmul(vec.transpose().conjugate(),vec) + np.identity(400)

print(cho_factor(matrix))

chrisrothUT: i had forgotten the conjugate

gcarleo: This works fine on my computer
```
import numpy as np
import jax.numpy as jnp
from jax.scipy.linalg import cho_factor

n=1000

vec = jnp.array(np.random.normal(0.,0.1,[20,n]) + 1j*np.random.normal(0.,0.1,[20,n]))

matrix = jnp.matmul(vec.transpose().conjugate(),vec) + jnp.array(np.identity(n))

print(cho_factor(matrix))
```

chrisrothUT: Yes, that script also breaks on my computer

gcarleo: Yes, I confirm that everything seems to work fine on my computer, also the solve function

```python
import numpy as np
import jax.numpy as jnp
from jax.scipy.linalg import cho_factor,cho_solve

n=1000

vec = jnp.array(np.random.normal(0.,0.1,[20,n]) + 1j*np.random.normal(0.,0.1,[20,n]))
b=jnp.array(np.random.normal(size=n)+1.0j*np.random.normal(size=n))
matrix = jnp.matmul(vec.transpose().conjugate(),vec) + jnp.array(np.identity(n))

fac=cho_factor(matrix)
sol=cho_solve(fac,np.asarray(b))

vec1=np.asarray(vec)
matrix1=np.asarray(matrix)


import scipy
from scipy.linalg import cho_factor
fac=cho_factor(matrix1)

sol1=scipy.linalg.cho_solve(fac,np.asarray(b))

assert(np.linalg.norm(sol-sol1,ord=np.inf)<1.0e-5)
```

chrisrothUT: I have a mac, I remember having similar bizarre issues with pytorch as well, so maybe i'll just implement and test with small arrays

chrisrothUT: import numpy as np
import jax.numpy as jnp
from jax.scipy.linalg import cho_factor,cho_solve,qr,solve_triangular,svd

n=30

vec = jnp.array(np.random.normal(0.,0.1,[20,n]) + 1j*np.random.normal(0.,0.1,[20,n]))
b=jnp.array(np.random.normal(size=n)+1.0j*np.random.normal(size=n))
matrix = jnp.matmul(vec.transpose().conjugate(),vec) + jnp.array(np.identity(n))

#cho
fac=cho_factor(matrix)
sol=cho_solve(fac,np.asarray(b))
assert(np.linalg.norm(jnp.matmul(matrix,sol)-b,ord=np.inf)<1.0e-5)

#qr
Q, R = qr(matrix)
c = jnp.matmul(Q.transpose().conjugate(),b)
sol = solve_triangular(R,c)
assert(np.linalg.norm(jnp.matmul(matrix,sol)-b,ord=np.inf)<1.0e-5)

#SVD
U, S, V = svd(matrix)
c = jnp.matmul(U.transpose().conjugate(),b)/S
sol = jnp.matmul(V.transpose().conjugate(),c)
assert(np.linalg.norm(jnp.matmul(matrix,sol)-b,ord=np.inf)<1.0e-5)

Here's the code I used to check the Jax solvers were working correctly. Unfortunately this breaks for larger "n" so it would be nice if someone gave it a shot on their computers.

gcarleo: It works fine (I tried up to n=1000). For n=1000 you might need to use double precision to satisfy the assertion though
```python
import os
os.environ["JAX_ENABLE_X64"] = "1"

import numpy as np
import jax.numpy as jnp
from jax.scipy.linalg import cho_factor,cho_solve,qr,solve_triangular,svd
```

gcarleo: So most likely it's an issue with your installation (I have mac too, and I am using this on conda)

netket/vmc_common.py
```diff
@@ -16,3 +16,52 @@ def tree_map(fun, tree):
...
         return fun(tree)
+
+import numpy as np
+from netket.utils import jax_available
+if jax_available:
+    from jax.tree_util import tree_flatten, tree_unflatten, tree_map
+    import jax.numpy as jnp
+
+
+def shape_for_sr(grads,jac):
```
gcarleo, line 27: this and the next function should be indented under if jax_available, right?

netket/machine/jax.py
```diff
@@ -149,16 +148,20 @@ def vector_jacobian_prod(
...
             return out
+
         else:
+
             if conjugate and self._dtype is complex:
-                prodj = lambda j: jax.np.tensordot(vec, j.conjugate(), axes=vec.ndim)
+                prodj = lambda j: jax.np.tensordot(vec.transpose(), j.conjugate(), axes=1)
             else:
-                prodj = lambda j: jax.np.tensordot(vec.conjugate(), j, axes=vec.ndim)
+                prodj = lambda j: jax.np.tensordot(vec.transpose().conjugate(), j, axes=1)
```
gcarleo, line 157: was the old version with axes=vec.ndim wrong?

chrisrothUT:
I've made all of the requested changes,  and removed all numpy references in the file jax_stochastic_reconfiguration. 

One potential issue is the parameter svd_threshold. Since jax does not have lstsq implemented I'm unsure how to enforce it.

gcarleo:
@chrisrothUT this last version is almost a GO, thanks! I just have a doubt on the changes you made to ```netket.machine.Jax```

gcarleo:
> One potential issue is the parameter svd_threshold. Since jax does not have lstsq implemented I'm unsure how to enforce it.

It's ok... for the moment we can just use it as it is and then add it when least square solvers become available in jax

netket/vmc_common.py
chrisrothUT: Yes, I'll fix that now

gcarleo:
Other minor thing, please don't forge to run the code formatter or use an automatic check in Atom or other editors you are using 

```
python -m black --check ../netket/
would reformat /home/travis/build/netket/netket/netket/_vmc.py
would reformat /home/travis/build/netket/netket/netket/machine/jax.py
would reformat /home/travis/build/netket/netket/netket/optimizer/jax_stochastic_reconfiguration.py
would reformat /home/travis/build/netket/netket/netket/optimizer/stochastic_reconfiguration.py
would reformat /home/travis/build/netket/netket/netket/vmc_common.py
Oh no! 💥 💔 💥
5 files would be reformatted, 69 files would be left unchanged.
The command "python -m black --check ../netket/" exited with 1.
```

netket/machine/jax.py
chrisrothUT: Yes, it didn't work. Axes are the dimensions to contract but "1" contracts the last index of the first arg with the first index of the second arg. Since j takes on different shapes, this seemed like the most straightforward solution

chrisrothUT:
Ohh yeah I will run black and repush

gcarleo:
Thank you @chrisrothUT I think we can merge this for now! 
I will later try to see if I can manage to speed up the jax solver

gcarleo:
Sorry @chrisrothUT but also this is not working... 
```python 
sr = nk.optimizer.JaxSR(diag_shift=0.1, use_iterative=False)
```

ValueError: All input arrays must have the same shape.

chrisrothUT:
Okay, fixed those two bugs, hopefully everything works!

chrisrothUT:
Looks like there's still a failure related to doing SVD with a threshold. This is the problem I alluded to earlier

gcarleo:
It's ok, I just remove that test for the SVD threshold, since it is not a valid test anymore. In principle we can make it valid again, considering that now we subtract the mean of the jacobian inside the SR, but it's a bit tricky

PR #411, opened 2020-05-09
Update der_log function (jacobian) in torch.py

ChenAo-Phys:
Add an optional way to calculate the Jacobian faster using backpack. It's activated automatically whenever the pack is installed and the network structure satisfies the requirements of the pack. 

The original method based on for-loops is also kept. It will be used when backpack doesn't work. It's not updated to the new pytorch Jacobian API, because the API is not for network calculation and doesn't accelerate the calculation.


netket/machine/torch.py
```diff
@@ -18,12 +24,27 @@ class Torch(AbstractMachine):
     def __init__(self, module, hilbert):
         self._module = _torch.jit.load(module) if isinstance(module, str) else module
         self._module.double()
+
+        if _backpack_imported:
+            self._use_backpack = True
+            for module in self._module.children():
+                if not isinstance(module, (_torch.nn.Conv2d, _torch.nn.Linear)) \
+                   and sum(p.numel() for p in module.parameters()):
```
twesterhout, line 32: That doesn't look right...

First of all, Backpack expects a module to be `torch.nn.Sequential` (see documentation). I.e. you should do `isinstance(module, torch.nn.Sequential)` for normal modules and `module.original_name == Sequential` (or something similar) for `torch.jit.ScriptModule`.

Secondly, not all parameter-less layers are supported by Backpack (see documentation). You should probably use an explicit list of supported layers or just wrap `extend` into a `try: ... except: ...` block.

Finally, there's the requirement that `forward` function should not be overridden (see documentation). This is non-trivial to check automatically. E.g. you could check whether the module *is* `torch.nn.Sequential` is or derived from it etc. But this doesn't cover the case when the user simply assigns a `forward` method dynamically... This will also break for `ScriptModule`s.

All in all, I'd suggest dropping the idea of automatically detecting when Backpack can be used and simply introduce a flag `use_backpack` and delegate the decision to the user who knows the network architecture.

ChenAo-Phys: Thanks for your reply. 

These constraints are necessary for backpack to work in both 1st-order extensions (jacobian included) and some other 2nd-order extensions. In my tests backpack works even we write a forward function or add some parameter-less layers created by ourselves or don't write it as a sequence. Otherwise some necessary manipulations like periodic boundaries will be impossible.

But I agree that the use of backpack in this way is not 100% safe. Change it to a manual opt-in may a better option.

gcarleo: OK, for me the important thing is that this is optional and that we don't use it when it doesn't work ! If it is hard to check automatically whether it is safe to use bacpack, then I agree with @twesterhout that the best solution is just to put a ```use_backpack``` argument in the ```__init__``` of the Torch class, and leave the user the burden to check if this makes sense for the model chosen

gcarleo:
@ChenAo-Phys small issue: it seems you have style problems in your file, please have a look at [our coding guidelines](https://github.com/netket/netket/blob/v3.0/CONTRIBUTING.md#python-coding-style)

gcarleo:
It might be worth looking also into [this trick](https://colab.research.google.com/drive/1tcm7Lvdv0krpPdaYHtWe7NA2bDbEQ0uj#scrollTo=ue-EqGLwhFoH)  (what is called there Reverse mode with repeat trick) that is essentially loop free

ChenAo-Phys:
> It might be worth looking also into [this trick](https://colab.research.google.com/drive/1tcm7Lvdv0krpPdaYHtWe7NA2bDbEQ0uj#scrollTo=ue-EqGLwhFoH) (what is called there Reverse mode with repeat trick) that is essentially loop free

The repeat trick uses a larger batch size to replace the for-loop in backward pass. For example, if we have 1000 samples, the batch size it uses will be 1000*1000. I have seen this trick before. It works well in very small batches, but it's a pity that this trick is not good in general cases. For CPU a larger size will cause an obvious slow down, while for GPU such a huge batch is too memory demanding.

gcarleo:
Hi @ChenAo-Phys , this would be a great time to pull the latest changes from v3.0, since we have finally stabilized the interface for the DerLog ! If you have time, it would be great...!

ChenAo-Phys:
> Hi @ChenAo-Phys , this would be a great time to pull the latest changes from v3.0, since we have finally stabilized the interface for the DerLog ! If you have time, it would be great...!

OK I will finish this in these days.

netket/machine/torch.py
```diff
@@ -15,12 +21,15 @@ def _get_differentiable_parameters(m):
...
-    def __init__(self, module, hilbert, dtype, outdtype=None):
+    def __init__(self, module, hilbert, dtype, outdtype=None, use_backpack=False):
         self._module = _torch.jit.load(module) if isinstance(module, str) else module
         self._module.double()
         self._n_par = _get_number_parameters(self._module)
         self._parameters = list(_get_differentiable_parameters(self._module))
-
+        self._use_backpack = use_backpack
+        if self._use_backpack:
+            assert "backpack" in sys.modules, "BackPACK was not successfully imported"
```
twesterhout, line 31: The standard way to do such checks is to introduce a global variable, say `_BACKPACK_AVAILABLE` (caps because it's global and underscore because it's private) which you set to `True` if `import backpack` succeeds and to `False` otherwise. Then here you can simply ensure that `_BACKPACK_AVAILABLE` is true. Also this should definitely be a raise instead of an assert.

netket/machine/torch.py
```diff
@@ -76,50 +85,70 @@ def n_par(self):
...
-            _torch.cat([dw.flatten() for dw in dws_real], out=out[i, 0, ...])
-            _torch.cat([dw.flatten() for dw in dws_imag], out=out[i, 1, ...])
-
-        out_complex = _np.zeros((out.size(0), out.size(2)), dtype=_np.complex128)
-        out_complex = out[:, 0, :].numpy() + 1.0j * out[:, 1, :].numpy()
-
-        return out_complex.reshape(
-            tuple(list(batch_shape) + list(out_complex.shape[-1:]))
-        )
+        if self._use_backpack:
```
twesterhout, line 109: Could you, please, split this into two separate functions? It'd be easier to read. There's currently too much stuff happening within a single function.

netket/machine/torch.py
```diff
@@ -76,50 +85,70 @@ def n_par(self):
...
+                i = 0
+                for gb in (
+                    p.grad_batch.flatten(start_dim=1)
+                    for p in self._module.parameters()
+                    if p.requires_grad
+                ):
+                    dst[:, i : i + gb.shape[1]].copy_(gb)
+                    i += gb.shape[1]
+
+            m_sum_real = self._module(x)[:, 0].sum(dim=0)
```
twesterhout, line 122: It'd be nice to check what is more efficient: to do the reduction to scalar here or to provide `grad` argument to `backward`. It doesn't have to happen now though, just create an issue if you don't have time to check it yourself.

netket/machine/torch.py
gcarleo: The approach we have been following so far is to add a ```is_available``` variable in ```netket.utils```, for example you can see in there that we already have ```netket.utils.torch_available``` etc. So you could either add ```netket.utils.backpack_available``` for consistency with the other cases, or follow @twesterhout suggestion and localize it instead in the Torch file... Maybe I have a preference for the first solution, but it's not a big deal

netket/machine/torch.py
```diff
@@ -76,50 +85,70 @@ def n_par(self):
...
+                )
+                dws_imag = _torch.autograd.grad(
+                    m[i, 1], self._parameters, retain_graph=True
+                )
+                out[i, ...].real = _torch.cat([dw.flatten() for dw in dws_real]).numpy()
+                out[i, ...].imag = _torch.cat([dw.flatten() for dw in dws_imag]).numpy()
+            self._module.zero_grad()
+        return out.reshape(tuple(list(batch_shape) + list(out.shape[-1:])))
 
     def vector_jacobian_prod(self, x, vec, out=None):
```
gcarleo, line 148: I think that in the final interface we have that this function also accepts ```return_jacobian``` as an option

netket/machine/torch.py
ChenAo-Phys: Inside pytorch when a grad is provided to backward the input is also reduced to a scalar first, so I think it's just a problem of different coding styles. I prefer to write the sum explicitly to make it easier to read.

ChenAo-Phys:
The checks of my new updates failed, but it seems not the problem of my code. test_hilbert.py and test_groundstate.py failed, but the torch machine is not even used in these files...

The problem seems to come from the reshape manipulation of non-contiguous numpy arrays. Could you give it a check @gcarleo ? Please tell me if it's really my mistake.

ChenAo-Phys:
Thanks for the suggestions from @twesterhout and @gcarleo . I'm sorry for the slow update. Some changes this time:

1. Use utils.backpack_available as @gcarleo suggests. I'm not sure whether it should be a `RuntimeError` when `use_backpack==True and backpack_available==False`.

2. Split der_log function into two parts.

3. `vector_jacobian_prod` now accepts a `return_jacobian` argument.

4. Some changes to make the code of `der_log` and `vector_jacobian_prod` easier to read.

There are some problems which exist before my updates. I didn't correct it but list them here for future contributors:

1. The Torch machine doesn't support GPU calculation. Many parts still need to be modified. For example, `torch.Tensor.numpy()` should be changed to `torch.Tensor.cpu().numpy()`. Some further tests are also needed.

2. The gradient graph is not freed after the calculation of jacobian if backpack is not used. It may be a problem if the memory usage approaches the limit.

gcarleo:
Thank you @ChenAo-Phys , the test failure looks like an issue with the latest version of numba, maybe for the purpose of this PR could you please force numba to use version 0.49 (in our setup.py)?

gcarleo:
Concerning GPU support, we should do the same thing we have done with the JAX backend, namely never convert things to numpy. However this needs some extra work, that should be done in another PR

gcarleo:
Thanks @PhilipVinc for fixing the numba issue, now also the tests in this PR pass!

PR #418, opened 2020-05-20
Python orthorhombic lattices

VolodyaCO:
I am adding an implementation of the Lattice graph class in Python, which handles orthorhombic lattices.

Some things are initially missing, but the main structure is already implemented. Among the things missing:

- [x] site_to_vector
- [x] vector_to_coord
- [ ] vector_to_site
- [x] basis_vectors
- [x] checking orthogonality


gcarleo:
You can now import this directly into ```netket/graph/__init__.py``` (and remove the line importing Lattice from the C_netket library)

VolodyaCO:
@gcarleo I am not sure that `vector_to_site` is useful at all. When one has two atoms or more per cell, one cannot be sure which atom shall this method refers to.

netket/graph/lattice.py
```diff
@@ -0,0 +1,146 @@
+from graph import NetworkX
```
gcarleo, line 1:
```suggestion
from .graph import NetworkX
```

gcarleo:
Thanks @VolodyaCO ! Interestingly enough, it seems that most of the existing tests in ```../Test/Graph/test_graph.py``` are for non orthorhombic lattices, that's why the Graphs tests are not passing here... !

For the non-orthorhombic case, can't we just do a coordinate transformation ?

VolodyaCO:
Yes, something like [this](https://scicomp.stackexchange.com/a/21857). I'll do it tomorrow

gcarleo:
Yes, this should work! (unless there are weird effects at the borders with pbcs) 
https://en.wikipedia.org/wiki/Fractional_coordinates

gcarleo:
Also [these notes](https://pdfs.semanticscholar.org/a5bd/f949911c0d7333ffec488b0589e88cd722b6.pdf) are interesting, and discuss the issues with pbcs and fractional coordinates

VolodyaCO:
The Euclidean distance of fractional coordinates does not match the Euclidean distance of cartesian coordinates. I have been looking for an efficient neighbour search method with periodic boundary conditions and a custom metric. My idea is to use fractional coordinates as if they were euclidean so that the periodic boundary conditions can be imposed as in a box (this is easily solved), but the distance is measured by transforming pairs of points to the real space (cartesian coordinates) and measuring the Euclidean distance in that space. I found this recent [project](https://grispy.readthedocs.io/en/latest/tutorial.html) called [GriSPy](https://arxiv.org/pdf/1912.09585.pdf) which implements quick neighbour search with cubic periodic boundary conditions and custom distance. I haven't found any KDTree implementation with both a custom metric and periodic boundary conditions. Is there any problem with trying GriSPy? @gcarleo   or do you want me to implement the method exposed by Smith in the notes you provided?

VolodyaCO:
There's also this old script of [CoverTree](https://github.com/patvarilly/CoverTree) which also handles periodic boundary conditions and custom distances. But GriSPy looks quite robust.

gcarleo:
I think it's easier if you just do the brute force thing: basically just compute all the ~N^2 distances in the cartesian metric (after rotating the coordinates), then convert them back to the original metric. At that point just pick the neighbors row by row of this matrix. This is not optimal for MD simulations, but in our case we do this operation only once (and for N~1000 at max), so we really don't care too much if this is a bit too slow!

VolodyaCO:
Ok. It looks like this distance for periodic boundary conditions should work:

```python
def distance(s0, s1, extent, basis_vectors):
    x0 = np.matmul(basis_vectors.T, s0)
    x1 = np.matmul(basis_vectors.T, s1)
    
    deltax = np.abs(x0 - x1)
    deltas = np.abs(s0 - s1)
    
    dimensions = extent
    deltas = np.where(deltas > 0.5 * dimensions, deltas - dimensions, deltas)
    deltax = np.matmul(basis_vectors.T, deltas)
    return np.sqrt((deltax ** 2).sum(axis=-1))
```
for `s0, s1` the coordinates of two atoms w.r.t. the basis vectors.

VolodyaCO:
Ok, I'm a bit uncomfortable with the minimum image convention, as it might fail for some cases. So I decided to add a "shell" of cells surrounding the real parallelepiped that is formed by the "true" cells. If any edges are found (through KDTree neighbour search) between the nodes from the "true" cells and the "shell" cells, then the corresponding image of the node in the "shell" cell is found within a "true" cell. This seems to be working... (I'll clean the code a bit before pushing changes). Maybe this is not an elegant solution, but it is certainly always correct. I've visualised it in two dimensions and it seems alright for this simple example with 2 atoms per cell, an extent of 2,2 (blue region, with a shell of 1 cell of thickness, which is used to find periodic neighbours):

image

VolodyaCO:
image
It seems to be working fine (this is with a skewed cell with basis vectors [1,1] and [1,0])... I'll put my code into the Lattice class, and will push tomorrow.

gcarleo:
Yes this approach is basically what is already implemented in our C++, so it should be fine. The issue is that sometimes a single shell is not enough, especially if you have very skewed unit cells?

gcarleo:
Thank you! Looks like you have an issue with PBC in tests? (you can run tests also locally just doing:
```
python -m pytest Test/Graph --verbose
```

VolodyaCO:
I'm getting an error:
```bash
==================================== ERRORS ====================================
__________________ ERROR collecting Test/Graph/test_graph.py ___________________
ImportError while importing test module '/home/vladimir/Documents/Projects/physics/netket-dev/netket/Test/Graph/test_graph.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
Test/Graph/test_graph.py:1: in <module>
    import netket as nk
netket/__init__.py:31: in <module>
    from . import (
E   ImportError: cannot import name '_C_netket' from 'netket' (/home/vladimir/Documents/Projects/physics/netket-dev/netket/netket/__init__.py)
```
I will use the log shown here in github to try and see what is happening.

PhilipVinc:
you have to build the c module first.
The easiest thing is for you to do `pip install .` from the netket folder.
and then run the tests from within the Test folder, otherwise you pickup the local netket, not the one installed through pip

VolodyaCO:
Ok, the problem is that I am checking that the input of pbc are boolean values, but the test is made with pbc=[0,0] instead of pbc=[False, False]

gcarleo:
Just change the test then to use bool

VolodyaCO:
Ok.

btw, I couldn't `pip install .` I get this error:

> warning: CMakeBuild: CMake Error: Error: generator : Ninja
>   Does not match the generator used previously: Unix Makefiles
>   Either remove the CMakeCache.txt file and CMakeFiles directory or choose a different binary directory.

I always installed it via `python setup.py install`, but the tests don't run for me.

VolodyaCO:
nvm, I deleted my build folder and everything worked as Philip indicated

VolodyaCO:
Ok, I changed the tests... there's something that you need to decide. I raise an error whenever the atom coordinate is outside [0,1). so [1, 0] cannot be an atom coordinate. This is to avoid that an atom from one cell overlaps with an atom from another cell. So I changed the atom coordinates in the test from [1,0] to [one, 0], where `one = 1 - eps, eps=1e-9`. I will push.

gcarleo:
@VolodyaCO I am not sure I understand this issue with 1-eps 

If you take this case 

```python 
        basis_vectors=[[1.5, math.sqrt(3) / 2.0], [0, math.sqrt(3)]],
        extent=[3, 5],

```

why you can't have 

```python
    atoms_coord=[[0, 0], [1, 0]],
```

that is well inside the unit cell

VolodyaCO:
Oh... I am assuming `atoms_coord` are fractional coordinates, so that the second atom would be on `[1.5, math.sqrt(3) / 2.0]`

gcarleo:
An no they are not supposed to be fractional coordinates...

VolodyaCO:
:rofl: Ok, let me do the changes then. I'll try to do them tonight.

VolodyaCO:
But, do I still require that the fractional coordinates are in [0,1)?

gcarleo:
Yes, fractional coordinates should be in there yes, otherwise there is ambiguity, as you were saying before! It's just that the way Lattice was designed (already before) was that everything is in cartesian coordinates, we never used fractional coordinates (of course you can use them internally, but that's an implementation detail)

VolodyaCO:
Ok, the test fails because there is a test example that is wrong:
```python
    Lattice(
        basis_vectors=[[1.5, math.sqrt(3) / 2.0], [0, math.sqrt(3)]],
        extent=[3, 5],
        atoms_coord=[[0, 0], [1, 0]])
```
This produces the unit cell depicted in the following image, and the atoms coordinates are 0,0, which is within the cell (point C), but 1,0 is not within the cell (point E):
image

Do I correct the tests and push them @gcarleo ?

gcarleo:
Ah interesting, yes please correct that....

PR #454, opened 2020-07-30
Rename implementation files in netket.operator

femtobit:
As brought up in issue #445, it can be confusing (at least with auto-completion) if `netket.operator` contains both `LocalOperator` directly (which users are supposed to use) and the submodule `local_operator.py` where that class is implemented. Thus, this implementation file is renamed. And for consistency, so are all the others to which the same argument applies.

Since none of the current PRs modify any operator code, now seems to be as good as time as any to do this with minimal conflicts.


PhilipVinc:
This will wreck havoc when rebasing, so let's do it now and forget about it

PhilipVinc:
Actually, before doing this, can we just check what is the customary way to do this in python packages?

It seems weird to me that all file names should start with underscores...

femtobit:
@PhilipVinc I think this is relatively common, see e.g. [PyTorch](https://github.com/pytorch/pytorch/tree/master/torch) which has lots of underscore-prefixed files. Jax does it too a lesser degree, but see. e.g, [`_util.py` here](https://github.com/google/jax/tree/master/jax/numpy).

I think other packages often (a) put more code in a single file or (b) do not care too much about this "namespace cleanness". Since for the purpose of supporting `from package import *` you can define `__all__`, this is mostly just an issue with autocompletion.

PhilipVinc:
Yes, you are right. I was just checking QuTip and indeed they don't care about this.
But we are perfectionists, so... let's merge ahead ⛵!

gcarleo:
Cool go ahead! I won't be able to review this soon sorry

femtobit:
There are some other places were we should also cleanup the file structure, but since there are some open PRs touching several of those, I think it's best if we wait until v3.0 is ready for release otherwise.

PhilipVinc:
agreed

PR #481, opened 2020-09-29
Add options for mixed PBC/OBC and direction-colored edges to Grid

femtobit:
This PR extends `Grid` to support mixed boundary conditions by passing a list of booleans to `pbc=`. It also adds an option to color the edges by direction in order to make it easier to define anisotropic models.

```python
Grid(length=[8, 8], pbc=[True, False])  # cylinder
Grid(length=[8, 8], color_edges=True)   # differently colored x/y edges
```

Under the hood, this PR also changes how the color is stored in `NetworkX` graphs: Previously, it was stored as the _key_ of an edge, which is a `networkx.MultiGraph`-feature meant to distinguish multiple edges between the same node. Now, I use an extra attribute `"color"` on each edge. This makes the implementation easier (attributes can be added and changed more easily than edge keys) and could allow us to use normal graphs where a multigraph is not needed in the future.

### DIFF
--- a/Test/Graph/test_graph.py
+++ b/Test/Graph/test_graph.py
@@ -5,11 +5,15 @@ import math
 
 from netket.graph import *
 
+import pytest
+
 nxg = nx.star_graph(10)
 graphs = [
     Hypercube(length=10, n_dim=1, pbc=True),
     Hypercube(length=4, n_dim=2, pbc=True),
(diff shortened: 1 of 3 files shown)


PhilipVinc (approved):

gcarleo:
thanks @femtobit good idea !

PR #517, opened 2020-12-01
WIP stochastic reconfiguration with jvp and vjp

inailuig:
Adds the option to do stochastic reconfiguration where the gradients are never stored, but (re-)calculated using  jvp and vjp for every matrix-vector product in the iterative solver.

There are several advantages of this approach:
- works for large numbers of samples and parameters where its impossible to store the gradients
- non-homogeneous parameters can be treated more easily
- can be faster for certain kinds of machines

I guess the integration in the existing SR class still needs some tweaking...
Also see the test for a version of the existing code which can deal with non-homogeneous parameters

### DIFF
--- a/Test/Optimizer/test_sr_onthefly.py
+++ b/Test/Optimizer/test_sr_onthefly.py
@@ -1,7 +1,9 @@
 import pytest
-from jax.config import config
+import netket as nk
 
-config.update("jax_enable_x64", True)
+if not nk.utils.jax_available:
+    pytest.skip("skipping jax-only SR-onthefly tests", allow_module_level=True)
+
 import jax
 import jax.numpy as jnp
 import jax.flatten_util
--- a/Test/Optimizer/test_sr_onthefly.py
+++ b/Test/Optimizer/test_sr_onthefly.py
@@ -54,10 +54,8 @@ def flatten(x):
 
 def toreal(x):
     if jnp.iscomplexobj(x):
-        # workaround for
-        # NotImplementedError: Transpose rule (for reverse-mode differentiation) for 'imag' not implemented
         return jnp.array(
-            [x.real, (-1j * x).real]
+            [x.real, x.imag]
         )  # need to use sth which jax thinks its a leaf
     else:
         return x
--- a/Test/Optimizer/test_sr_onthefly.py
+++ b/Test/Optimizer/test_sr_onthefly.py
@@ -12,8 +12,6 @@ from jax.scipy.sparse.linalg import cg
 from netket.optimizer.jax._sr_onthefly import *
 from netket.optimizer.jax.stochastic_reconfiguration import _jax_cg_solve_onthefly
 
-# TODO more sophisitcated example?
-
 
 @partial(jax.vmap, in_axes=(None, 0))
 def f(params, x):
@@ -73,7 +71,6 @@ def tree_toreal_flat(x):
 def reassemble_complex(x, fun=tree_toreal_flat, target=params):
     # target: a tree with the expected shape and types of the result
     (res,) = jax.linear_transpose(fun, target)(x)
-    # TODO ...
     res = tree_conj(res)
     # fix the dtypes:
     return tree_cast(res, target)
@@ -116,20 +113,18 @@ def test_reassemble_complex():
 
 def test_vjp():
     actual = O_vjp(samples, params, v_tilde, f)
-    # TODO ...
     expected = tree_conj(reassemble_complex((v_tilde @ ok_real).real))
     assert tree_allclose(actual, expected)
 
 
 def test_mean():
     actual = O_mean(samples, params, f)
-    # TODO ...
     expected = tree_conj(reassemble_complex(okmean_real.real))
     assert tree_allclose(actual, expected)
 
 
-def test_Odagger_w():
-    actual = Odagger_w(samples, params, v_tilde, f)
+def test_OH_w():
+    actual = OH_w(samples, params, v_tilde, f)
     expected = reassemble_complex((ok_real.conjugate().transpose() @ v_tilde).real)
     assert tree_allclose(actual, expected)
 
@@ -170,7 +165,7 @@ def test_matvec_linear_transpose():
     # S^T = (O^H O)^T = O^T O* = (O^H O)* = S*
     # S^T w = S* w = (S w*)*
     expected = tree_conj(mat_vec(tree_conj(w), f, params, samples, 0.0))
-    # e, = jax.linear_transpose(lambda v_: reassemble_complex(S_real @ tree_toreal_flat(v_)), v)(v)
+    # (expected,) = jax.linear_transpose(lambda v_: reassemble_complex(S_real @ tree_toreal_flat(v_)), v)(v)
     assert tree_allclose(actual, expected)
 
 
--- a/Test/Optimizer/test_sr_onthefly.py
+++ b/Test/Optimizer/test_sr_onthefly.py
@@ -42,10 +42,64 @@ S = dok.conjugate().transpose() @ dok / n_samp
 
 real_ind = flatten(jax.tree_map(jax.numpy.isrealobj, params))
 def setzero_imag_part_of_real_params(x):
-    # workaround for imag not differentiable
+    # workaround for
+    # NotImplementedError: Transpose rule (for reverse-mode differentiation) for 'imag' not implemented
     return jax.ops.index_add(x, real_ind, -1j*(-1j*x[real_ind]).real)
 
 
+
+
+#flatten params to a vector of real params
+
+
+def toreal(x):
+    if jnp.iscomplexobj(x):
+        # workaround for
+        # NotImplementedError: Transpose rule (for reverse-mode differentiation) for 'imag' not implemented
+        return jnp.array([x.real, (-1j*x).real]) # need to use sth which jax thinks its a leaf
+    else:
+        return x
+
+def tree_toreal(x):
+    return jax.tree_map(toreal, x)
+
+def tree_toreal_flat(x):
+    return flatten(tree_toreal(x))
+
+params_real_flat = tree_toreal_flat(params)
+
+# invert the trafo using linear_transpose (AD)
+def reassemble_complex(x, fun=tree_toreal_flat, target=params):
+    # target: some tree with the shape and types we want
+    _lt = jax.linear_transpose(fun, target)
+    # jax gradient is actually the conjugated one, so we need to fix it:
+    return jax.tree_map(jax.lax.conj, _lt(x)[0])
+
+
+def f_real_flat(p, samples):
+    return f(reassemble_complex(p) , samples)
+
+grad_real_flat = tree_toreal_flat(grad)
+v_real_flat = tree_toreal_flat(v)
+
+def f_real_flat_scalar(params, x):
+    return f_real_flat(params, jnp.expand_dims(x, 0))[0]
+
+def _rea(x):
+    re, im = x
+    return re +1j*im
+
+
+@partial(jax.vmap, in_axes=(None,0))
+def grads_real(params, x):
+    r = jax.grad(lambda pars, v: f_real_flat_scalar(pars, v).real)(params_real_flat, x)
+    i = jax.grad(lambda pars, v: f_real_flat_scalar(pars, v).imag)(params_real_flat, x)
+    return _rea((r,i))
+
+ok_real = grads_real(params_real_flat, samples)
+okmean_real = ok_real.mean(axis=0)
+dok_real = ok_real - okmean_real
+S_real = dok_real.conjugate().transpose() @ dok_real / n_samp
 
 
 def test_f_flat():
@@ -108,5 +162,10 @@ def test_cg():
     def mv(v):
         return setzero_imag_part_of_real_params(S @ v + diag_shift * v)
     e = setzero_imag_part_of_real_params(cg(mv, grad_flat, x0=v_flat, tol=sparse_tol, maxiter=sparse_maxiter)[0])
+    def mv_real(v):
+        #_compose_result_real also takes real here
+        return (S_real @ v + diag_shift * v).real
+    e2 = flatten(reassemble_complex(cg(mv_real, grad_real_flat, x0=v_real_flat, tol=sparse_tol, maxiter=sparse_maxiter)[0]))
     assert jnp.allclose(a, e)
     assert jnp.allclose(b, e)
+    assert jnp.allclose(e, e2)
--- a/Test/Optimizer/test_sr_onthefly.py
+++ b/Test/Optimizer/test_sr_onthefly.py
@@ -136,7 +136,7 @@ def test_vjp():
 
 
 def test_obar():
-    a = Obar(samples, params, f, samples.shape[0])
+    a = Obar(samples, params, f)
     e = reassemble_complex(okmean_real.real)
     assert tree_allclose(tree_conj(a), e)
 
@@ -156,14 +156,14 @@ def test_odagov():
 
 
 def test_odagdeltaov():
-    a = odagdeltaov(samples, params, v, f, n_samp)
+    a = odagdeltaov(samples, params, v, f)
     e = reassemble_complex(S_real @ v_real_flat)
     assert tree_allclose(a, e)
 
 
 def test_matvec():
     diag_shift = 0.01
-    a = mat_vec(v, f, params, samples, diag_shift, n_samp)
+    a = mat_vec(v, f, params, samples, diag_shift)
     e = reassemble_complex(S_real @ v_real_flat + diag_shift * v_real_flat)
     assert tree_allclose(a, e)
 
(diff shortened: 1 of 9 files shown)


gcarleo:
This is a high priority PR, especially if it provides performance advance + storage advance (to be tested though)

Examples/Jax/jax_sr_example.py
```diff
@@ -1,38 +1,58 @@
...
 
+
 # Hilbert space of spins on the graph
 hi = nk.hilbert.Spin(s=0.5, graph=g)
 
 ha = nk.operator.Ising(h=1.0, hilbert=hi)
 
 alpha = 1
 ma = nk.machine.JaxRbm(hi, alpha, dtype=complex)
+ma2 = nk.machine.JaxRbm(hi, alpha, dtype=complex)
```
PhilipVinc, line 18: I don't get the point of this example.
Are you trying to show that you get exactly the same result?

Otherwise I think it's better to create another example file.

netket/optimizer/jax/_sr_onthefly.py
```diff
@@ -0,0 +1,92 @@
...
+        lambda x, target: (x if jnp.iscomplexobj(target) else x.real).astype(
+            target.dtype
+        ),
+        res,
+        v,
+    )
+    return res
+
+
+def mat_vec(v, forward_fn, params, samples, diag_shift, n_samp):
```
PhilipVinc, line 79: n_samp is a redundant parameter? And it's static when jitting so even better.

```suggestion
def mat_vec(v, forward_fn, params, samples, diag_shift):
    n_samp = samples.shape[0]
```

netket/optimizer/jax/_sr_onthefly.py
```diff
@@ -0,0 +1,92 @@
+import jax
+import jax.numpy as jnp
+from functools import partial
+
+
+# onthefly SR
+# since we cant store O if #samples x #params is too large
+
+
+def O_jvp(x, theta, v, vlogwf):
```
PhilipVinc, line 10: can you rename `vlogwf` to be consistent with what we use in the rest of netket? I'd say `forward_fn` should be good, or something else if you prefer.

PhilipVinc: also below

PhilipVinc:
I'd like to think a bit about the interface. 
The nested if blocks are ugly as hell.

Also, having `onthefly` as a property of `vmc` and not of SR is wrong, in my opinion.

Essentially, I think that an optimiser should somehow advertise what he needs to work (gradient, grad+jacobian, grad+(forward_fun/samples))... and then the driver should feed it what he asks.

Maybe we could do this with a class hierarchy? Would it still be hackable enough?

gcarleo:
In my opinion also the interface is not appropriate at the moment, and as I was also commenting ```on_the_fly``` should be, in case, a property of the sr optimizer... However, I think that if on the fly is actually always faster than the normal case, I would be in favor of completely removing the old one, since that also, is ugly as hell :)

femtobit:
@gcarleo There are cases where, as far as I understand, access to the full S matrix (or its SVD/eigendecomposition including the eigenvectors) is necessary. For example for the regularization proposed here: https://arxiv.org/abs/1912.08828.

So we should definitely keep the option of computing S fully instead of solving the SR equation on-the-fly for cases where it may be needed (or at least would be much simpler than rephrasing everything in terms of vjp/jvp).

gcarleo:
I think you can still rewrite the full S matrix (if needed) using only vip
and jvp. But my point is really that we should try to benchmark this
approach and see if it's as fast as or even faster than the one already
implemented

netket/optimizer/jax/_sr_onthefly.py
inailuig: I actually kept this so that if we want to parallelize it it with MPI at some point,
then we can divide by the n_samp across all MPI processes and just sum (via allreduce) at the end.
(in analogy to the matrix version)

What do you think?

netket/optimizer/jax/_sr_onthefly.py
inailuig: d3f6780a0eea376620c45af90ccd6d0590a4c851 done.

netket/optimizer/jax/_sr_onthefly.py
inailuig: Actually I think all that is missing for MPI is to add the linear transpose rule (ad.primitive_transposes) for mpi_allreduce_p in mpi4jax, (or use jax._src.scipy.sparse.linalg._cg_solve directly, which does not require it; hardcoding symmetric=True in jax._src.scipy.sparse.linalg.cg is also a possible workaround just for testing it)

PhilipVinc: The number of MPI nodes is a known and runtime constant, so you don't need it?

As for the rule... yes, IIRC I [have defined](https://github.com/PhilipVinc/mpi4jax/blob/875e60b1737014ecd6885f4c64564705b5ce3c81/mpi4jax/collective_ops/allreduce.py#L157) the backward jvp pass but not the forward vip pass on allreduce. 
 It should be trivial to add

Examples/Jax/jax_sr_example.py
inailuig: Yes, I abused this example a bit for testing, I think once we figured out the interface we can restore the old example and add a new one for the on-the-fly version

netket/optimizer/jax/stochastic_reconfiguration.py
```diff
@@ -102,11 +100,12 @@ def _shape_for_sr(grads, jac):
...
+_flatten_grad_and_oks = jax.jit(_shape_for_sr)
+
+
 @jit
-def _flatten_grad_and_oks(grad, oks):
-    grad, oks = _shape_for_sr(grad, oks)
-    oks -= jnp.mean(oks, axis=0)
-    return grad, oks
+def _subtract_mean_from_oks(oks, n_samp):
+    return oks - _sum_inplace(jnp.sum(oks, axis=0) / n_samp)
```
inailuig, line 109: Before my change the mean of ok was calculated on each rank separately only from the samples of that rank, was that on purpose in order to avoid communication? (its only one scalar per node anyway) ?

netket/optimizer/jax/stochastic_reconfiguration.py
```diff
@@ -62,9 +58,33 @@ def _jax_cg_solve(
...
+    x0,
+    forward_fn,
+    params,
+    samples,
+    grad,
+    diag_shift,
+    n_samp,
+    sparse_tol,
+    sparse_maxiter,
+):
```
inailuig, line 72: @PhilipVinc I could just add `n_samp` to `static_argnums` here (possibly also `diag_shift`, `sparse_tol`, `sparse_maxiter`)

PhilipVinc: n_samp could be a static_argnum because it de-facto is. 
But, as I said in my comment above, it is a redundant parameter.  You can compute it from the shape of `samples`, which is a `compile time constant` and the number of MPI nodes, which is a runtime constant. 
Why are you passing it? Unless I'm missing something? 

`sparse_tol` and `sparse_maxiter` would be better not be compile time constants to avoid unnecessary recompilation. 
However, since we are jitting this block and therefore the cg inside I'm unsure if jax will correctly-pick up changes in sparse_tol. I suspect not.

Can you check that if you change the tolerance, something you can do without creating a new SR object and that the api supports, the jitted function will correctly pick up the new tolerance even without it being a static_argnum?

If not, it must be static.

netket/optimizer/jax/stochastic_reconfiguration.py
PhilipVinc: This was... indeed a bug. And probably on me since I had rewritten that code at some point. 
Thanks for spotting it. In any case it was executed only once so... not really an overhead.

However, related to my comment above, can you please not use the argument `n_samp` unless there's a good reason, instead just compute it when you need it?

netket/optimizer/jax/stochastic_reconfiguration.py
```diff
@@ -232,6 +254,44 @@ def compute_update(self, oks, grad, out=None):
...
+        O_i(v_j) = ∂/∂x_i log Ψ(v_j) and f is a generalized force (the loss
+        gradient).
+
+        Args:
+            samples: An array of samples
+            grad: A pytree of the forces f
+            out: A pytree of the parameter updates that will be ignored
+        """
+
+        n_samp = samples.shape[0] * n_nodes
```
PhilipVinc, line 273: my point essentially is that this should go inside the jit-block.
That's because it's easier for the compiler to constant-propagate things that he knows are identical.
And that cleans up the function signatures so the code is more readable.

PhilipVinc:
@gcarleo I am starting to think that to solve the _interface problem_ in this context, something along the lines of the `AbstractVariationalState` you had proposed in #437 would go a long way...

Essentially, in that case what would happen is that the optimiser (SR) would ask
```python
for _ in range(vmc_steps):
   variational_state.reset()
   energy = variational_state.estimate(Hamiltonian)
   dp = compute_gradient(variational_state, hamiltonian, optimiser) 
   variational_state.update(dp)
```

```python
def compute_gradient(variational_state, operator, optimiser):
     optimiser.compute_gradient(variational_state, operator)
```

```python
# maybe @singledispatch on type of operator so you can differentiate Lindblad/Hailtonian/timeevolution...
def SR.compute_gradient(self, variational_state, operator):
    S = variational_state.quantum_geometric_tensor(full_matrix=self.full_matrix)
    grad =  variational_state.gradient()
   return dp = S^-1 * grad 
```

```python
# maybe @singledispatch on type of operator so you can differentiate Lindblad/Hailtonian/timeevolution...
def SR_onthefly.compute_gradient(self, variational_state, operator):
   ....
```

netket/stats/_sum_inplace.py
```diff
@@ -78,7 +78,8 @@ def sum_inplace_jax(x):
             if _n_nodes == 1:
                 return x
             else:
-                res, _ = mpi4jax.Allreduce(x, op=_MPI.SUM, comm=_MPI_comm)
+                token = jax.lax.create_token(0)
+                res, _ = mpi4jax.Allreduce(x, op=_MPI.SUM, comm=_MPI_comm, token=token)
```
inailuig, line 82: @PhilipVinc This is the only way I could get it working, although i think that is because the token might be eliminated by dead code elimination?

Otherwise I either get  NotImplementedError: Transpose rule (for reverse-mode differentiation) for 'create_token' not implemented  if I dont pass a token or let it depend on x here, or TypeError: No constant handler for type: <class 'jax.interpreters.xla.Token'>  if I create the token outside of the function and just capture it here.

Examples/Jax/jax_sr_example.py
inailuig: I removed the commit when I rebased it, so the example is again the way it was before.

netket/optimizer/jax/stochastic_reconfiguration.py
inailuig: done. 61b0beeb81e7ff651e3b08cb75fb476ac263b5cb

netket/optimizer/jax/stochastic_reconfiguration.py
inailuig: 61b0beeb81e7ff651e3b08cb75fb476ac263b5cb

netket/stats/_sum_inplace.py
PhilipVinc: This should be ok. 
In any case we don't need tokens for our usecase because there is only 1 mpi call and they are sequential.

However, I'll have to find a solution for mpi4jax in the future.

Test/Optimizer/test_sr_onthefly.py
```diff
@@ -0,0 +1,192 @@
+import pytest
+from jax.config import config
```
PhilipVinc, line 2: You have failing tests in the runner with no jax installed. 

Jax it's not an hard dependency on netket, and related functionality is only enabled when jax is detected. IIRC There is a bool flag in `netket.utils.jax_available` that you should use for that.

Likewise, tests should import jax only if this flag is true.

You can do something along the lines of [this](https://docs.pytest.org/en/latest/skipping.html), see the code below (untested). 

```
import netket as nk
import pytest


if not nk.utils.jax_available:
    pytest.skip("skipping jax-only SR-onthefly tests", allow_module_level=True)

import jax
etc...
```

PhilipVinc:
@inailuig I have tagged a new release of mpi4jax (v0.2.7) . 
You should bump the minimum required version on the `setup.py`.

However, since this is not the only way to install mpi4jax, could you also add a check to `netket/utils.py`, for `mpi4jax.__version__ ` to check that it's at or above 0.2.7, and if not raise an ImportError saying something along the lines we are only compatible with v0.2.7 or above, please update it (run pip install -U ....)

netket/optimizer/jax/stochastic_reconfiguration.py
```diff
@@ -226,6 +245,41 @@ def compute_update(self, oks, grad, out=None):
...
+        Args:
+            samples: An array of samples
+            grad: A pytree of the forces f
+            out: A pytree of the parameter updates that will be ignored
+        """
+
+        if self._x0 is None:
+            self._x0 = jax.tree_map(jnp.zeros_like, grad)  # x0 = jnp.zeros_like(grad)
+
+        if self._use_iterative:
```
gcarleo, line 267: do we check that this works only if we are using the iterative solver?

inailuig: We check it in the init function: https://github.com/netket/netket/blob/b28b2ab264ace962f2561d9336e2272215373556/netket/optimizer/jax/stochastic_reconfiguration.py#L144

netket/optimizer/jax/_sr_onthefly.py
```diff
@@ -0,0 +1,105 @@
...
+        return res, vjp_fun
+    else:
+        return res
+
+
+# calculate \langle O \rangle
+def Obar(samples, theta, forward_fn, **kwargs):
+    # TODO better way to get dtype
+    dtype = forward_fn(theta, samples[:1])[0].dtype
+    v = jax.lax.broadcast(
```
gcarleo, line 31: can you please add a comment on what broadcast is doing here?

PhilipVinc: Yep. In general if you could better document what exactly is the flow of the code among the several small functions you defined, it would be great.

It's a bit hard to follow.

netket/optimizer/jax/stochastic_reconfiguration.py
gcarleo: great thanks

netket/optimizer/jax/_sr_onthefly.py
inailuig: Do you have suggestion on how I could solve the TODO with the dtype here?

PhilipVinc:
@inailuig Could you also do a benchmark comparison on Ising with N=20, and different values of alpha... On the Fly vs Not-On the fly for 100 optimisation steps?

netket/optimizer/jax/_sr_onthefly.py
PhilipVinc: There is not really a way to avoid that. 
However, please check that the forward pass is being optimised away (I believe it should be, but I'm not 100% sure).
You can check on the generated IR by using `jax.make_jaxpr(Obar)(samples, theta...)`.

If it does not cut out the forward pass, then we will need to pass it as a static arg.

netket/stats/_sum_inplace.py
```diff
@@ -78,7 +78,8 @@ def sum_inplace_jax(x):
             if _n_nodes == 1:
                 return x
             else:
-                res, _ = mpi4jax.Allreduce(x, op=_MPI.SUM, comm=_MPI_comm)
+                token = jax.lax.create_token(0)
```
PhilipVinc, line 81:
```suggestion
                # Note: We must supply a token because we can't transpose `create_token`.
                # The token can't depend on x for the same reason
                # This token depends on a constant and will be eliminated by DCE
                token = jax.lax.create_token(0)
```

PhilipVinc:
@inailuig thanks for all the hard work up to this point.
I think there now only remain two issues to be addrsesed:
 - The core algorithm in SR_onthfly and the methods used internally are a bit hard to follow. Can you add some comments to make the intent clearer (and the code more manageable in the future?)
 -  I think we had agreed to switch ON by default on_the_fly now. (@gcarleo please confirm)

Lastly, upon reflecting more about it, 
 - I think that the `onthefly` switch should be a toggle in the `SR` object. This is because
   - Only jax machines can use it. So when you create a JaxSR object we can default to it, otherwise ignore the flag
   - If your jax machine has non-homogeneous weights (not same type) we can default to it and error in the other case.
   - If you try to pass this kwarg to a non Jax SR object, we error.

Then, the driver will just check if the SR is on the fly or not and call the right method with the right arguments.

This also makes it easier to add support for onthefly to `steadystate` and `qsr` which is something we should do.

netket/utils.py
```diff
@@ -47,6 +47,11 @@ def jit_if_singleproc(f, *args, **kwargs):
 else:
     mpi4jax_available = False
 
+if mpi4jax_available and not mpi4jax.__version__ >= "0.2.7":
```
PhilipVinc, line 50: You sure about that? 
https://stackoverflow.com/questions/11887762/how-do-i-compare-version-numbers-in-python

PhilipVinc:
Great work. 
I think that if you fix the formatting this is now good to go.

PhilipVinc (approved):

PR #521, opened 2020-12-08
Use __matmul__ (@) for the operator product

femtobit:
Python has (since [PEP 465](https://www.python.org/dev/peps/pep-0465/)) a matrix product operator `__matmul__` (`@`) distinct from `__mul__` (`*`) which is commonly used for element-wise multiplication.

Since this is a widely adopted convention in Python (in particular numpy), it makes sense for us to follow it as well, which is proposed in this PR. Then, expressions such as
```python
id = -1j * sx @ sy @ sz
```
have the same semantics for NetKet operators as for numpy arrays.

Note that `numpy.ndarray` uses `@` in this way, while the `numpy.matrix` uses `*` but is deprecated.

This PR also replaces the occasional `return NotImplementedError` with actually raising the exception.

### DIFF
--- a/Test/Operator/test_local_operator.py
+++ b/Test/Operator/test_local_operator.py
@@ -24,20 +24,17 @@ sp = [[0, 1], [0, 0]]
 g = nk.graph.Graph(edges=[[i, i + 1] for i in range(8)])
 hi = nk.hilbert.CustomHilbert(local_states=[-1, 1], graph=g)
 
-sx_hat = nk.operator.LocalOperator(hi, [sx] * 3, [[0], [1], [4]])
-sy_hat = nk.operator.LocalOperator(hi, [sy] * 4, [[1], [2], [3], [4]])
-szsz_hat = nk.operator.LocalOperator(hi, sz, [0]) * nk.operator.LocalOperator(
-    hi, sz, [1]
-)
-szsz_hat += nk.operator.LocalOperator(hi, sz, [4]) * nk.operator.LocalOperator(
-    hi, sz, [5]
-)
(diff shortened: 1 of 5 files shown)


gcarleo:
Thanks @femtobit. In principle this is not a bad idea but I personally don't like very much the idea of using @ to declare a product... it seems to me much more intuitive to use * also because for these operators the element-wise product is not really something we provide

femtobit:
I can understand the sentiment. In Julia, for example, I would not suggest this, since there `*` for product/composition of operators is more common (but Julia also has `.*` for element-wise multiplication). However, I would expect at least people used to working with numpy etc. to find `@` more intuitive by now, given that it is standardized (though some older parts of the ecosystem take some time to switch) and we do have a purely Python framework now.

The fact that numpy (and thus jax.numpy etc.) and other libraries use `__matmul__` means that one has to use different syntax for NetKet compared to the libraries we depend on, which I find needlessly confusing. I would at least suggest keeping `__matmul__` as an option.

gcarleo:
Yes I am def in favour of keeping matmul as an option but I wouldn’t remove the possibility of using * either!

femtobit:
@gcarleo Alright, one can now use `op1 * op2` (operator product or multiplication, if one factor is a number) and `op1 @ op2` (only operator product).

gcarleo (approved):
great thanks!

PhilipVinc:
i am not particularly fond of having two (* and @) ways of doing the multiplication, but I am not particularly against it either.

on an unrelated note, since you are on it, can you please add right division of an operator by a scalar? It is missing and it bugs me that you cannot use it

PR #573, opened 2021-03-22
Add permutation-invariant DenseSymm layer and RBMSymm

femtobit:
Reopening of #555.

This PR now adds a `DenseSymm` layer that can be used as follows:
https://github.com/netket/netket/blob/a37fa4a931882b03b76dbae18991bd66cf727469/netket/models/rbm.py#L330-L338

`DenseSymm` is implemented, as per @PhilipVinc's suggestion, by (once) computing a `symmerizer` tensor that is then applied as
https://github.com/netket/netket/blob/a37fa4a931882b03b76dbae18991bd66cf727469/netket/nn/linear.py#L282
which should do the same thing as the old C++ code:
https://github.com/netket/netket/blob/9b2a22b23dc307e396027ef99fd73b519824c102/Sources/Machine/rbm_spin_symm.cc#L224-L229

Performance looks good on my machine: Running the Heisenberg 1D example, 
```
# With RBMSymm(..., alpha=4)
100%|█| 300/300 [00:25<00:00, 11.72it/s, Energy=-35.611 ± 0.011 [σ²=0.125, R̂=0.9990]]]
# With non-symmetrized RBM(..., alpha=4)
100%|█| 300/300 [00:17<00:00, 17.37it/s, Energy=-35.217 ± 0.055 [σ²=3.099, R̂=1.0020]]]
```
Still a bit slower than `RBM`, but not even by a factor of 2 and there is probably room for optimization. Also, the compilation time is now pretty good.

Some open questions:
- Should `DenseSymm` have a user-friendly constructor (similar to `RBMSymm` right now)? Right now, one has to pass a lambda returning a `DeviceArray` of symmetries, otherwise weird errors happen.
- We can probably remove the `expand` implementation of `RBMSymm`. Should we also remove the `lax.scan` one? It is slower than the symmetrizer one but still a somewhat educational example, IMHO, given that it pretty directly implements the formula from the original Science paper. (And maybe it will be beneficial for very large systems?)

Open TODOs, which I'll get to soon:
- [x] Remove code duplication for the `RBM`, now that we have `DenseSymm`.
- [x] Write proper docs.
Not necessarily in this PR:
- [x] Add proper tests that run `VMC` on our models (some jax-related errors do not occur when initializing the machine but during VMC, since more code is jit-ed then).


gcarleo:
@femtobit I am all in for user friendly! Indeed at least the symmetries should be accepted also as regular numpy arrays or (even) just lists (can you maybe do something like ```jnp.as_array(symmetries)```?

Examples/Heisenberg1d/heisenberg1d.py
```diff
@@ -0,0 +1,58 @@
...
+sr = nk.optim.SR(0.1)
+
+# Variational monte carlo driver
+gs = nk.Vmc(ha, op, sa, ma, n_samples=1000, n_discard=100, sr=sr)
+
+# Print parameter structure
+print(f"parameter shape: {jax.tree_map(np.shape, gs.state.variables['params'])}")
+
+# Compile the model
+gs.run(n_iter=1)
```
PhilipVinc, line 54:
```suggestion
```

We don't really need that anymore I think. I reset the timer in run.

femtobit:
@gcarleo Now `DenseSymm` accepts arrays, `AbstractGraph`s, lists of lists, and similar. Same for `RBMSymm`.

The need for the external `create_...` functions is a bit ugly, but I don't see a better way at the moment (unless there is a good way to override `__init__` while properly inheriting from `nn.Module`, @PhilipVinc).

netket/models/rbm.py
```diff
@@ -127,3 +127,72 @@ def __call__(self, x):
...
+        activation: The nonlinear activation function
+        alpha: feature density. Number of features equal to alpha * input.shape[-1]
+        use_bias: if True uses a bias in the dense layer (hidden layer bias)
+        use_visible_bias: if True adds a bias to the input
+        kernel_init: initializer function for the weight matrix.
+        bias_init: initializer function for the bias.
+        visible_bias_init: initializer function for the visible_bias.
+    """
+
+    permutations: Callable
```
PhilipVinc, line 147: What should be the signature?

netket/models/rbm.py
```diff
@@ -127,3 +127,72 @@ def __call__(self, x):
...
+    use_bias: bool = True
+    use_visible_bias: bool = True
+
+    kernel_init: Callable[[PRNGKey, Shape, Dtype], Array] = normal(stddev=0.1)
+    bias_init: Callable[[PRNGKey, Shape, Dtype], Array] = normal(stddev=0.1)
+    visible_bias_init: Callable[[PRNGKey, Shape, Dtype], Array] = normal(stddev=0.1)
+
+    @nn.compact
+    def __call__(self, x_in):
+        x = nknn.DenseSymm(
```
PhilipVinc, line 160: is this calling the class or the class constructor?

netket/nn/linear.py
```diff
@@ -207,6 +208,99 @@ def __call__(self, inputs: Array) -> Array:
         return y
 
 
+class DenseSymm(Module):
+    permutations: Any
+    alpha: Union[float, int]
```
PhilipVinc, line 213: I'd switch the two around, and check if my callable signature is ok
```suggestion
    alpha: Union[float, int]
    permutations: Callable[[], Array]
```

PhilipVinc: also, flax uses features by default (which is not alpha).
In this case our Alpha is features, no?

netket/nn/linear.py
```diff
@@ -207,6 +208,99 @@ def __call__(self, inputs: Array) -> Array:
...
+
+        self.symmetrizer = jnp.array(
+            np.fromfunction(
+                symmetrizer_ijkl,
+                shape=(self.n_sites, self.n_hidden, self.n_sites, self.features),
+                dtype=int,
+            )
+        )
+
+    def matmul(self, inputs: Array, kernel: Array) -> Array:
```
PhilipVinc, line 248: maybe mark it as internal?
```suggestion
    def _matmul(self, inputs: Array, kernel: Array) -> Array:
```

netket/nn/linear.py
```diff
@@ -207,6 +208,99 @@ def __call__(self, inputs: Array) -> Array:
...
+            cond_k = k == jnp.asarray(perms)[j % self.n_symm, i]
+            cond_l = l == jsymm
+            return jnp.asarray(jnp.logical_and(cond_k, cond_l), dtype=int)
+
+        self.symmetrizer = jnp.array(
+            np.fromfunction(
+                symmetrizer_ijkl,
+                shape=(self.n_sites, self.n_hidden, self.n_sites, self.features),
+                dtype=int,
+            )
```
PhilipVinc, line 245: CPUs don't do float*int (some very recent gpus do in low precision but let's avoid that for now).
```python
>>> def mm(x,y):
...  return x*y
...
>>> import jax
>>> x=jax.numpy.array([1,2,3])
>>> y=jax.numpy.array([1.0,2.0,3.0])
>>> jax.make_jaxpr(mm)(x,y)
{ lambda  ; a b.
  let c = convert_element_type[ new_dtype=float64
                                weak_type=False ] a
      d = mul c b
  in (d,) }
```

So i'd do
  ```suggestion
        self.symmetrizer = jnp.array(
            np.fromfunction(
                symmetrizer_ijkl,
                shape=(self.n_sites, self.n_hidden, self.n_sites, self.features),
                dtype=int,
            ), dtype=dtype,
```

**Comment by PhilipVinc** [comment]
>@gcarleo Now DenseSymm accepts arrays, AbstractGraphs, lists of lists, and similar. Same for RBMSymm.

So DenseSymm is no longer a nn.Module but is a function right? How does it play with the fact that you build modules with `("blabla", alpha=1, symmetries)`?

**Comment by PhilipVinc** [comment]
> unless there is a good way to override __init__ while properly inheriting from nn.Module, @PhilipVinc
Normally for frozen data classes you'd override `__post_init__(self)` and modify attributes with `object.__setattr__(self, 'field', new_value)`. 
But I think Flax is doing weird stuff to the `__post_init__` and I could not get it was erroring when calling the parent post init method

File: netket/models/rbm.py (hunk 3 above)
Comment (femtobit) [reply netket/models/rbm.py:160]: The constructor.

File: netket/models/rbm.py (hunk 2 above)
Comment (femtobit) [reply netket/models/rbm.py:147]: `Callable[[], Array]` - I can add that, though with the constructors this is pretty internal anyways.

**Comment by femtobit** [comment]
> So DenseSymm is no longer a nn.Module but is a function right? How does it play with the fact that you build modules with `("blabla", alpha=1, symmetries)`?

It works with keyword arguments `nk.nn.DenseSymm(name="Dense", permutations=g, alpha=2)`. Maybe I should force passing `permutations` as kwarg? Then the signature should be equivalent to `nn.Module` otherwise.

File: netket/nn/linear.py (hunk 4 above)
Comment (femtobit) [reply netket/nn/linear.py:213]: Hm, I am still a bit confused by the `alpha`, `features` terminology (at least in the case where `N_sym != N_sites`). Since `alpha` is used only to compute `self.features`, it should be possible to just pass features - we just need to make sure the code works e.g. for non-integer `alpha`s (not sure if it actually does right now).

File: netket/nn/linear.py [hunk 7]
```diff
@@ -207,6 +208,99 @@ def __call__(self, inputs: Array) -> Array:
...
+        Returns:
+          The transformed input.
+        """
+        dtype = jnp.promote_types(inputs.dtype, self.dtype)
+        inputs = jnp.asarray(inputs, dtype)
+
+        kernel = self.param(
+            "kernel", self.kernel_init, (inputs.shape[-1], self.features), self.dtype
+        )
+        kernel = jnp.einsum("ijkl,kl->ij", self.symmetrizer, kernel)
```
PhilipVinc, line 271: it might be worth benchmarking if doing jnp.vectorize-> matmul->reshape is more efficient than einsum.
It shouldn't be, but in the past I have seen that sometimes it is.

netket/nn/linear.py
PhilipVinc: if` N_sym == N_sites ` we all agree that `alpha=features` in this case.

In the other case... alpha doesn't make too much sense, but features does.
At least in my mind.

You see it because the `int` call down there fore features is flooring the result of the multiplication, which I think might lead to subtle bugs and I'd avoid, even if that was our old behaviour.

PhilipVinc: cc @gcarleo

PhilipVinc:
My point is that that 'utility' constructor is kind of useless if someone tries to build a model a'la rbm, as you have to put that code at the outer-most level. 
You have this issue because you have to define `create_RBMSymm` too.

Moreover, right now, we are recompiling the module every time we are passing in a different symmetry, because it's a different closure, which leads to a different hash... ecc.. ecc..

I'm wondering if there is a nicer way to address this.
Overriding __post_init__ would be the best way.

femtobit:
> Moreover, right now, we are recompiling the module every time we are passing in a different symmetry, because it's a different closure, which leads to a different hash... ecc.. ecc..

That I can live with, I don't think changing the symmetries is a very frequent operation.

>  Overriding `__post_init__` would be the best way.

I can take a look. Defining `permutations: Array` as an [init-only](https://docs.python.org/3/library/dataclasses.html#init-only-variables) variable so that it is passed to `__post_init__` should be the way, but I have to see whether that works well with `flax`.

femtobit:
> My point is that that 'utility' constructor is kind of useless if someone tries to build a model a'la rbm, as you have to put that code at the outer-most level.

If you define the model yourself, then you can just pass the lambda or hard-code the array instead of making it a property, right? The main reason for the extra constructors is that we cannot generically add a non-hashable property.

PhilipVinc:
>I can take a look. Defining permutations: Array as an init-only variable so that it is passed to __post_init__ should be the way, assuming we can get this to work with flax.

I looked a bit and I don't think it's feasible.
But it would be worth asking over at flax about it.

netket/nn/linear.py
femtobit: I've compared `einsum` to this:
```python
kernel = jnp.matmul(self.symmetrizer, kernel.reshape(-1)).reshape(inputs.shape[-1], -1)
```
There does not seem to be much of a difference, at least for the Heisenberg1d example.
The `matmul` version might be a bit faster - but there is some variance so I'd need to look at it more carefully to make sure.

netket/nn/linear.py
femtobit: The `DenseSymm` layer uses `features` now, with `n_hidden = features * n_symm`. `RBMSymm` still has `alpha` for compatibility with `RBM`.

PhilipVinc:
(also to do, add RbmSymm and DenseSymm to `docs/docs/api.rst`)

Examples/Heisenberg1d/heisenberg1d.py
```diff
@@ -0,0 +1,55 @@
...
+
+# Ising spin hamiltonian
+ha = nk.operator.Heisenberg(hilbert=hi, graph=g)
+
+# RBM Spin Machine
+ma = nk.models.RBMSymm(
+    permutations=g.periodic_translations(),
+    alpha=4,
+    use_visible_bias=False,
+    use_bias=True,
```
gcarleo, line 36: maybe we can say use_hidden_bias to avoid confusion with the old API?

gcarleo: or ideally use_bias activates both... but OK...

Examples/Heisenberg1d/heisenberg1d.py
```diff
@@ -0,0 +1,55 @@
...
+
+# Optimizer
+op = nk.optim.Sgd(learning_rate=0.01)
+sr = nk.optim.SR(0.1)
+
+# Variational monte carlo driver
+gs = nk.VMC(ha, op, sa, ma, n_samples=1000, n_discard=100, sr=sr)
+
+# Print parameter structure
+print(f"parameter shape: {jax.tree_map(np.shape, gs.state.variables['params'])}")
```
gcarleo, line 51: is this a debug line?

Examples/Heisenberg1d/heisenberg1d.py
femtobit: We can use `use_hidden_bias` - in that case we should update it on the `RBM` as well.

Examples/Heisenberg1d/heisenberg1d.py
femtobit: I find it nice to have in the example because it shows the reduced number of parameters compared to the standard RBM.

gcarleo: agreed, but what about using ```n_par``` from the variational state?

femtobit: Alright, changed.

Examples/Heisenberg1d/heisenberg1d.py
femtobit: This PR should now update `use_bias` to `use_hidden_bias` on all RBMs.

PhilipVinc: The reason for using use_bias was that dense layers use that, so I was trying to be consistent with the rest of the ecosystem.

gcarleo: yes true... but then use_visible_bias becomes problematic... :) I think it's fine to use RBM specific nomenclature

PhilipVinc:
Just a last minor comment:
I think it would be better to export from netket.nn both `DenseSymm` and `construct_DenseSymm` with the different names.
The docstring of DenseSymm shuould point to `construct_DenseSymm` with the reason why.

femtobit:
@PhilipVinc I see you point. OTOH, I still consider this to be a workaround. Ideally, we could do something like
```python
class Example(nn.Module):
    some_array: dataclasses.InitVar[Array]

    def setup(self, some_array):
        self.some_array = preprocess(some_array)

    @nn.compact
    def __call__(self, x_in):
        return x_in @ self.some_array

```
in the same way it is currently possible to write 
```python
@dataclasses.dataclass
class DataClass:
    some_array: dataclasses.InitVar[Array]

    def __post_init__(self, some_array):
        self.some_array = preprocess(some_array)
```
for a pure dataclass. But that would need support from `flax`.

In any case, the `create_...` functions are in this sense an implementation detail to me - and I still have the hope we can get rid of them later, but I would not let that block us here at the moment. If people use `create_DenseSymm` explicitly (and since it's the more convenient method, I would expect that), that'd make it harder to replace later on.

PhilipVinc:
>@PhilipVinc I see your point. OTOH, I still consider this to be a workaround. Ideally, we could do something like

Yes I agree, my point is simply that, from the point of API stability, is less 'committing' to commit to having a `DenseSymm` that does not work with arbitrary inputs, and a `construct_***` that does, than a `DenseSymm` that works with everything.

While we figure that everything works, I prefer a slightly uglier API, and commit to something only at the end.

PhilipVinc:
Whaat do you think?

femtobit:
@PhilipVinc Fair enough. It is also true that if we make it more convenient to create `DenseSymm` directly in the future, defining
```python
@deprecated
def create_DenseSymm(...):
    return DenseSymm(...)
```
is easy enough, if the workaround has been present for long enough.

femtobit:
@gcarleo What do you say?

PhilipVinc:
THat is exactly what I was advocating for

--
Filippo Vicentini
Collaborateur Scientific
EPFL, Lausanne
Google Scholar
+41 21 693 58 38
Il 23 mar 2021, 12:53 +0100, Damian Hofmann ***@***.***>, ha scritto:

PhilipVinc:
Also, If you feel like it, for the nn.Modules the doc strings are rendered better if you put them in the style I used in the last PR I opened.

Otherwise I’ll fix them right after we merge.

If you just address the constructor then we can merge.

--
Filippo Vicentini
Collaborateur Scientific
EPFL, Lausanne
Google Scholar
+41 21 693 58 38
Il 23 mar 2021, 12:53 +0100, Damian Hofmann ***@***.***>, ha scritto:

PhilipVinc:
btw, my point was to do that a only for the DenseSymm, not for RBMSymm, which is somewhat always an end-user interface.

PhilipVinc (approved):

Examples/Heisenberg1d/heisenberg1d.py
```diff
@@ -0,0 +1,55 @@
...
+# Lattice translation operations
+
+# Hilbert space of spins on the graph
+hi = nk.hilbert.Spin(s=1 / 2, N=g.n_nodes, total_sz=0)
+
+# Ising spin hamiltonian
+ha = nk.operator.Heisenberg(hilbert=hi, graph=g)
+
+# RBM Spin Machine
+ma = nk.models.create_RBMSymm(
```
gcarleo, line 32: @femtobit I think I overlooked this, I thought we were still keeping ```RBMSymm`` as a simple constructor ?

femtobit: True, I was a bit confused by @PhilipVinc's request, though now I understand he only advocated to export both `DenseSymm` and `create_DenseSymm` for the layer and not for the RBM.

I can correct this - though treating layer and RBM differently still looks somewhat weird to me - while the layer is a bit more "internal", it is still part of the public API and definitely reusable outside of NetKet code.

femtobit: In the long run, we should probably find a solution along the lines of https://github.com/netket/netket/pull/573#issuecomment-804819668 (probably by discussing this with flax developers), which I could try in a few days.

PhilipVinc: I agree 100% with your point that this is not consistent.

But my idea is that this is temporary until we find a decent workaround.
And it achieves the objective to keep a simple API for users that won't try to do more advanced things like defining their models.
In any case, if someone wants to define a model, I expect them to read at least a few bits of documentation (otherwise how can they know how to do that?), so they should see the warnings about `Create_DenseSymm`.

PR #608, opened 2021-04-09
cleanup SR

inailuig:
This is in preparation for completing the TODO regarding blocking the computations.


gcarleo (approved):
thanks @inailuig cleaning up is always a great idea!

PR #620, opened 2021-04-13
Group Equivariant Convolutional Networks

chrisrothUT:
Hi everybody,

I now have a working implementation of G-CNNs that works with a generic SymmGroup. GCNN works exactly like RBM_Symm. Here is a 4 layer GCNN on the Kagome lattice:

```
graph = nk.graph.Lattice(basis_vectors=[[1.,0.],[-1./2.,np.sqrt(3)/2]],extent=[2,2],atoms_coord=[[0.,0.],[-1./4.,np.sqrt(3)/4.],[1./2.,0.]])

ma = nk.models.GCNN(
    symmetries = graph,
    layers = 4,
    features = 4,
    dtype=complex
)
```

And we're off! The N layer model involves a layer DenseSymm, followed by N-1 DenseEquivariant layers which convolve over the entire set of poses based on their relative symmetry orientations. Like RBMSymm, this will produce identical outputs for each automorphism of the lattice, and I've included a similar test.

Questions
- What would be a good test for the layer DenseEquivariant?
- Can we figure out a way to only call reshape() once in DenseEquivariant (instead of twice)
- What are good defaults for initialization? 

Future
- Implement restricted convolutions. This requires information about the symmetry group, not just a randomly ordered list of automorphisms. 
- After testing on a variety of lattices, we tend to get caught in local minima on frustrated lattices. So we should think about adding an annealing method (i.e. optimizing phases first) to VMC (obviously beyond the scope of this PR)


netket/graph/lattice.py
```diff
@@ -22,6 +22,8 @@
```
PhilipVinc, line 23:
```suggestion
cutoff_tol = 1e-5
"""Tolerance for the maximum distance cutoff when computing the sparse distance matrix.
This is necessary because of floating-point errors when computing the distance in non-trivial 
lattices.
""" 

```

PhilipVinc: Add this here and use it below as needed.
This way it is more clear why we have it and we always use the same number (easier to change in the future if necessary).

netket/nn/linear.py
```diff
@@ -347,6 +347,115 @@ def create_DenseSymm(
...
+        the full Dense kernel of shape (n_sites, features * n_symm).
+        """
+
+        result = jnp.take(kernel, self.group_algebra, 0)
+        result = result.reshape(
+            self.n_symm, self.n_symm, self.in_features, self.out_features
+        )
+        result = result.transpose(2, 0, 3, 1).reshape(
+            self.n_symm * self.in_features, -1
+        )
```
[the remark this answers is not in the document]
chrisrothUT: I was under the impression that reshapes copy memory. At least that is the case for Pytorch where .view() was the analog of reshape that doesn't touch memory. I trust your read on the situation.

I can't think of a good way to avoid the transposition. The mapping is transformed_kernel[a*n_symm + i, b*n_symm + j] = kernel[group_algebra[i,j], a, b]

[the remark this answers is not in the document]
chrisrothUT: right now group_algebra is a flattened and turned into a tuple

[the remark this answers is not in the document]
PhilipVinc: I can't speak for torch as I never used it, but that is the case* in jax. (*: Technically it's not guaranteed as it's treated like an optimization, but the Jax compiler is very good at optimizing away memory reuses)

netket/nn/linear.py
```diff
@@ -347,6 +347,115 @@ def create_DenseSymm(
...
+            y += bias
+
+        return y
+
+
+def create_DenseEquivariant(
+    symmetries: Union[AbstractGraph, Array],
+    *args,
+    **kwargs,
+):
```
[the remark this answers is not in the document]
chrisrothUT: Yeah, it definitely feels a little clunky. The big issue IMO is that the Flax module only wants static attributes, probably something to do with the jit compilation. They're may be a workaround I don't know about

netket/graph/symmetry.py
```diff
@@ -80,6 +108,16 @@ def remove_duplicates(self, *, return_inverse=False):
...
 
+    def inverse(self):
+
+        return _inverse(self.to_array())
+
+    def group_algebra(self):
+
+        inverse = self.inverse()
+
+        return _group_algebra(self.to_array(), inverse)
```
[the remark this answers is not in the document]
chrisrothUT: fixed

netket/graph/symmetry.py
```diff
@@ -21,6 +21,34 @@
...
+    for i, perm1 in enumerate(automorphisms):
+        for j, perm2 in enumerate(automorphisms):
+            perm_sq = perm1[perm2]
+            if np.all(perm_sq == np.arange(len(perm_sq))):
+                inverse[i] = j
+
+    return automorphisms[inverse]
+
+
+def group_algebra(automorphisms, inverse):
```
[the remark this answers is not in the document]
chrisrothUT: should I add annotation in both places?

netket/graph/lattice.py
chrisrothUT: Yeah this makes sense

netket/graph/symmetry.py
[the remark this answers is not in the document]
PhilipVinc: Yes.
They make the code clearer to whoever will read it in the future.

netket/nn/linear.py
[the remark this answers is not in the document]
PhilipVinc: See what we settled with in #621.
You should get rid of `create_DenseEquivariant` and users should pass in a `SymmetryGroup` object or an `HashableArray` if they want to do something custom.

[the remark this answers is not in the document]
chrisrothUT: Looks good. There's still a bit of a problem on my end. group_algebra() is slow and if I call it in setup.py, it will call it at every iteration

netket/graph/symmetry.py
```diff
@@ -80,6 +80,49 @@ def remove_duplicates(self, *, return_inverse=False):
...
+        inverse = self.inverse()
+        group_algebra = np.zeros([n_symm, n_symm], dtype=int)
+
+        for i, inv in enumerate(inverse):
+            for j, perm in enumerate(automorphisms):
+                for k, filter in enumerate(automorphisms):
+                    if np.all(perm[inv] == filter):
+                        group_algebra[i, j] = k
+
+        return tuple(group_algebra.ravel())
```
PhilipVinc, line 124: Why is this converted to a tuple? Is this in order to work with Flax?

netket/models/equivariant.py
```diff
@@ -0,0 +1,141 @@
...
+    precision: Any = None
+    """numerical precision of the computation see `jax.lax.Precision`for details."""
+    kernel_init: NNInitFunc = variance_scaling(1.0, "fan_in", "normal")
+    """Initializer for the Dense layer matrix."""
+    bias_init: NNInitFunc = zeros
+    """Initializer for the hidden bias."""
+
+    def setup(self):
+
+        self.n_symm, _ = self.symmetries().shape
```
PhilipVinc, line 64: See the recent PR by @femtobit and changes to DenseSymm. Here you can just call 'symm.shape'

netket/models/equivariant.py
```diff
@@ -0,0 +1,141 @@
...
+def create_GCNN(
+    symmetries: Union[AbstractGraph, Array],
+    *args,
+    **kwargs,
+):
+    """
+    Constructor for GCNN
+    """
+
+    if isinstance(symmetries, AbstractGraph):
```
PhilipVinc, line 122: We opted for not taking in a graph when symmetries are required. It's either a matrix or symmgroup. So remove this if branch

netket/graph/symmetry.py
chrisrothUT: Yeah, I've fixed it and made it a HashableArray

PhilipVinc: Great. I just would like @femtobit opinion here about whoever it makes sense to return an hashable array here or if we should keep the build_GCN method and return a normal array here...

I think that if we make HashableArray behave satisfy the numpy API we might get away with just returning a HashableArray. 

This means adding dtype,shape,__len__,ndim,getitem I think...

Then you could use it as a numpy array and all is good as long as you don't modify it.

netket/models/equivariant.py
chrisrothUT: Yes, I read the PR and I've implemented the changes.  Now the constructors are gone and GCNN excepts a SymmGroup or a HashableArray. If it's an array, you also need to specify group_algebra

PhilipVinc: Yeah sorry, all is fine.
I guess i did not refresh the page... my bad

chrisrothUT: No problem, just didn't want to waste your time looking through my code when I was fixing things already :)

netket/graph/symmetry.py
femtobit: > I think that if we make HashableArray behave satisfy the numpy API we might get away with just returning a HashableArray.

I wanted to avoid that. Making it behave like a numpy array _most of the time_ is doable, but there might be additional bugs or at least inconsistencies when it doesn't. (Just as `DeviceArray` and `ndarray` mostly behave the same but have subtle differences.)

The only time you need a `HashableArray` is when you use it as a model attribute. I feel we should just wrap it there and use normal objects otherwise.

femtobit: Maybe it also makes sense for group algebra to be its own class? I haven't had time to look at this PR in more detail so I don't yet know if that makes any sense though. I'll take a closer look in the next days.

netket/models/equivariant.py
femtobit: Should be `np.asarray(self.symmetries).shape` as of the current master.

netket/graph/symmetry.py
chrisrothUT: I moved group_algebra to be a function in SymmGroup. I don't see why you would need to call it on something that isn't a SymmGroup object

chrisrothUT:
Okay, made a few changes. GCNN takes symmetries=SymmGroup or arrays need to be specified for symmetries and group_algebra. I added a test for GCNN_creation as well

Test/Models/test_nn.py
```diff
@@ -22,26 +22,31 @@
...
 
 
-def _setup_symm(symmetries, N):
-    hi = nk.hilbert.Spin(1 / 2, N)
+def _setup_symm(symmetries, N, lattice="chain"):
+
+    if lattice == "chain":
+        g = nk.graph.Chain(N)
+    else:
+        g = nk.graph.Square(N)
```
femtobit, line 30: You could just pass a function, like
```suggestion
def _setup_symm(symmetries, N, lattice=nk.graph.Chain):
     g = lattice(N)
```
and change the calling sites accordingly. That saves the additional `if` blocks.

Test/Models/test_rbm.py
```diff
@@ -39,7 +40,37 @@ def test_RBMSymm(use_hidden_bias, use_visible_bias, symmetries):
...
+        nk.optim.Sgd(0.1),
+        nk.sampler.MetropolisLocal(hi),
+        ma,
+    )
+    vmc.advance(1)
+
+
+@pytest.mark.parametrize("use_bias", [True, False])
+@pytest.mark.parametrize("symmetries", ["trans", "autom"])
+@pytest.mark.parametrize("lattice", ["chain", "square"])
```
femtobit, line 59:
```suggestion
@pytest.mark.parametrize("lattice", [nk.graph.Chain, nk.graph.Square])
```
see above

Test/Models/test_nn.py
chrisrothUT: this works for me

PhilipVinc:
You should also add the new classes/model to the documentation by putting their name in the correct subsection of docs/docs/api.rst

netket/nn/linear.py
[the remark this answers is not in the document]
chrisrothUT: I tested tensordot with and without transpose, as well as jax.lax.dot and they all take about 3 microseconds per iteration on a Colab GPU. The intra run variance for N=1000 was larger than the individual differences between runs  

```
n_sym = 100
n_hid = 16
batch_size = 100

def transpose(inp,kernel):
    kernel = kernel.transpose(2,0,3,1)
    kernel = kernel.reshape(n_hid*n_sym,n_hid*n_sym)

    out = jax.lax.dot_general(inp,kernel,(((inp.ndim - 1,), (0,)), ((), ())))

    return out

def lax_dot(inp,kernel):
    kernel = kernel.transpose(2,0,3,1)
    kernel = kernel.reshape(n_hid*n_sym,n_hid*n_sym)

    out = jax.lax.dot(inp,kernel)

    return out

def no_transpose(inp,kernel):
    inp = inp.reshape(batch_size,n_hid,n_sym) 

    out = jax.lax.dot_general(inp,kernel,(((1,2), (2,0)), ((), ())))
    out = out.reshape(batch_size,n_sym*n_hid)

    return out

jit_transpose = jax.jit(partial(transpose,kernel=kernel_unshaped))
jit_no_transpose = jax.jit(partial(no_transpose,kernel=kernel_unshaped))
jit_dot = jax.jit(partial(lax_dot,kernel=kernel_unshaped))
```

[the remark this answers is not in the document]
PhilipVinc: I think the XLA compiler is compiling the same code for both. So that is good.

netket/graph/symmetry.py
```diff
@@ -80,6 +81,49 @@ def remove_duplicates(self, *, return_inverse=False):
         else:
             return group
 
+    def inverse(self):
+
```
PhilipVinc, line 85:
```suggestion
```

netket/graph/symmetry.py
```diff
@@ -80,6 +81,49 @@ def remove_duplicates(self, *, return_inverse=False):
...
+        for i, perm1 in enumerate(automorphisms):
+            for j, perm2 in enumerate(automorphisms):
+                perm_sq = perm1[perm2]
+                if np.all(perm_sq == np.arange(len(perm_sq))):
+                    inverse[i] = j
+
+        return automorphisms[inverse]
+
+    def group_algebra(self):
+
```
PhilipVinc, line 105:
```suggestion
```

chrisrothUT:
Okay, I added a test for DenseEquivariant, and updated the docs. I think this should be ready for review

chrisrothUT:
Okay there's still this annoying issue. SymmGroup.group_algebra() is an expensive function, where compute time scales like n_symm^3. If you pass SymmGroup to GCNN it computes group_algebra() in setup, and for whatever reason calls the function several times while its compiling. There's a few options for remedying this issue:

1. Speed up group_algebra(). Right now I use loops and there's probably some ways to take advantage of linear search algorithms
2. Give the user a warning. "This symmetry group has a lot of elements, for quicker compile time pass group_algebra as an argument directly" 
3. Call group_algebra in the init method of SymmGroup, but this might waste time for folks who don't want access to this tensor

PhilipVinc:
how expensive is that function? I thought it was fairly cheap...
how long does it take on reasonably sized lattices?

In any case, as it's simply three nested loop, I'd slap a `@numba.njit` decorator on it to make us happier.
Also, since the SymmGroup is non mutable/static, you can just compile it once and then serve it cached, like we do for the hash

```python
@numba.njit
def group_algebra(automorphisms, inverse):
    return whatever

class SymmGroup:
   def __init__():
        ....
        self.__group_algebra = None

    def group_algebra(self):
         if self.__group_algebra is None:
            self.__group_algebra = HashableArray(group_algebra(self.automorphisms, self.inverse())
         return self.__group_algebra
```

You could also do the same for `inverse` and `to_array()` if you think they take considerable time.

chrisrothUT:
For a 6x6 honeycomb with n_symm = 432 it takes 5 mins. So for a 10x10 lattice it probably takes a couple of hours. If we can find a way to make it take n_symm^2 time it shouldn't be a problem at all. 

This is a nice solution to the problem of calling it multiple times.

chrisrothUT:
Ohh shoot, that can't work because SymmGroup is a frozen dataclass

PhilipVinc:
Its legit, just use __setattr__

gcarleo:
sorry I am a bit lost in the conversation, maybe can you tell me where ```SymmGroup.group_algebra()```is implemented and why it takes so long on small lattices? (if finding symmetries for a 10x10 lattice takes a few hours this is a very serious bottleneck we should solve, most likely going back to the automorphisms or finding a faster implementation!)

netket/graph/symmetry.py
```diff
@@ -80,6 +81,47 @@ def remove_duplicates(self, *, return_inverse=False):
...
+        This array is flattened and converted to a tuple before returning
+
+        """
+
+        automorphisms = self.to_array()
+        n_symm = len(automorphisms)
+        inverse = self.inverse().to_array()
+        group_algebra = np.zeros([n_symm, n_symm], dtype=int)
+
+        for i, inv in enumerate(inverse):
```
gcarleo, line 117: oh I see, it's this triple loop + an if running an all the vector!!... uhm isn't there a way to rewrite this is a faster way?

PhilipVinc:
Also @chrisrothUT , while this PR is in a good state (assuming we can address the N^3 scaling), would you consider also adding a commented tutorial (jupyter notebook) walking people through the model and the parameters, how it works and what it does, and maybe comparing it to other networks? A bit like https://www.netket.org/tutorials/Heisenberg1d.html

netket/graph/symmetry.py
chrisrothUT: I'm sure there's a solution with hashing that is a lot faster. I'm trying to identify the indices of duplicates between a list of length of N and a list of length N^2 (which I can make with arrays). I think I can make a dict with the N^2 list and loop over the N list?

chrisrothUT:
@PhilipVinc I already wrote something for my blog so I can make that into a tutorial!

netket/graph/symmetry.py
chrisrothUT: Yeah, I think I have to hash over the length N list but this should still be fast enough

chrisrothUT: Okay, problem has been solved, it's a lot faster than graph.automorphisms(), and scales approximately like 17microseconds*n^2. So even for a 10x10 triangular that is like 25 seconds

netket/models/equivariant.py
```diff
@@ -0,0 +1,130 @@
...
+from netket.utils.types import PRNGKeyT, Shape, DType, Array, NNInitFunc
+
+
+from netket import nn as nknn
+from netket.nn.initializers import lecun_complex, zeros, variance_scaling
+
+
+class GCNN(nn.Module):
+    """Implements a group convolutional neural network with symmetry
+    averaging in the last layer as described in Roth et al. 2021."""
```
PhilipVinc, line 36: Can you also write something more here?

Put a new line and then write a bit more, so the website documentation is actually helpful.
you can also write latex equations.

netket/graph/symmetry.py
```diff
@@ -80,6 +81,47 @@ def remove_duplicates(self, *, return_inverse=False):
         else:
             return group
 
+    def inverse(self):
+        """
+        Returns reordered symmgroup where the each element is the inverse of
+        the original symmetry element
+
```
PhilipVinc, line 88: can you explain this better?

netket/graph/symmetry.py
PhilipVinc: Still, could you check if putting it in a numba jit block makes it even a bit faster?

netket/models/equivariant.py
```diff
@@ -0,0 +1,130 @@
...
+    """A group of symmetry operations (or array of permutation indices) over which the layer should be invariant.
+    Numpy/Jax arrays must be wrapped into an :class:`netket.utils.HashableArray`. 
+    """
+    layers: int
+    """Number of layers (not including sum layer over output)"""
+    features: Union[Tuple, int]
+    """Number of features in each layer starting from the input. If
+    a single number is given, all layers will have the same number
+    of features"""
+    group_algebra: HashableArray = None
```
PhilipVinc, line 48:
```suggestion
    group_algebra: Optional[HashableArray] = None
```

netket/models/equivariant.py
chrisrothUT: Okay that is good to know I'll try to make these more helpful

netket/models/equivariant.py
```diff
@@ -0,0 +1,130 @@
...
+    """A flattened array representing the mapping between relative and absolute poses.
+    Does not to be specified if symmetries is a SymmGroup
```
PhilipVinc, line 50:
```suggestion
    Does not need to be specified if symmetries is a SymmGroup
```

chrisrothUT:
@PhilipVinc Considering group algebra is significantly faster than automorphisms(), do you still think it's worth it to make the changes we discussed

netket/nn/linear.py
```diff
@@ -314,6 +318,95 @@ def __call__(self, inputs: Array) -> Array:
         return y
 
 
+class DenseEquivariant(Module):
+    """An equivariant layer that takes a feature map with embeddings corresponding to various
+    poses of a symmetry group, shape = [n_symm*in_features] and outputs a feature map of identical
```
PhilipVinc, line 323: maybe i am ignorant, but what is a pose of a symmetry group?

netket/nn/linear.py
```diff
@@ -314,6 +318,95 @@ def __call__(self, inputs: Array) -> Array:
...
+    shape [n_symm*out_features]. This transformation respects the group structure such that
```
PhilipVinc, line 324: identical_shape but `[n_symm*in_features]` and `[n_symm*out_features]` are clearly different?
What do you mean?

netket/nn/linear.py
```diff
@@ -314,6 +318,95 @@ def __call__(self, inputs: Array) -> Array:
...
+    g L({f}) = L(g{f}) where g is an element of the symmetry group and {f} are the feature embeddings.
+    """
+
+    group_algebra: HashableArray
+    """Flattened array generated  SymmGroup.group_algebra() that specifies the mapping
```
PhilipVinc, line 329: typo?

netket/nn/linear.py
```diff
@@ -314,6 +318,95 @@ def __call__(self, inputs: Array) -> Array:
         return y
 
 
+class DenseEquivariant(Module):
+    """An equivariant layer that takes a feature map with embeddings corresponding to various
+    poses of a symmetry group, shape = [n_symm*in_features] and outputs a feature map of identical
+    shape [n_symm*out_features]. This transformation respects the group structure such that
+    g L({f}) = L(g{f}) where g is an element of the symmetry group and {f} are the feature embeddings.
```
PhilipVinc, line 325: Maybe you could also give an explicit example of what happens to shapes. 
Given an input shape (N_batches, N_spins, blablabla) you get ...

netket/nn/linear.py
```diff
@@ -314,6 +318,95 @@ def __call__(self, inputs: Array) -> Array:
...
+    """
+
+    group_algebra: HashableArray
+    """Flattened array generated  SymmGroup.group_algebra() that specifies the mapping
+    between absolute and relative poses """
+    in_features: int
+    """The number of symmetry-reduced input features. The full input size
```
PhilipVinc, line 332: I'm not sure that this is correct. The input/output size will also depend on batches... I think giving an explicit example with what the shapes should be would be clearer.

netket/nn/linear.py
chrisrothUT: Pose is an ML term, is the word orientation more clear? It's quite tricky to explain what is going on in a paragraph, that's why I plan to write more stuff

netket/nn/linear.py
chrisrothUT: Good catch

PhilipVinc:
If its really not the bottleneck, ok. Let's add it to the todo and try to speed  up automorphisms @femtobit with numba in the future.

netket/models/equivariant.py
PhilipVinc: A good example of a complete documentation and using latex is in LocalLiouvillian.

In general everything you write before a newline is considered the short-summary used in short descriptions in the website, and then after the newline you can write everything else, generates the docpage.

netket/nn/linear.py
PhilipVinc: I can only guess what you mean by orientation...

As a short description I would specify carefully what should be the inputs and outputs and with maybe easy/to/understand words what it means. 
Then newline, 
then try to explain concepts you need to say what this does. 
Links to websites/papers are welcome 
you write a link with the syntax 
```python
` text <https:://link.to.something>`_
```

chrisrothUT:
Okay I think I've covered everything, I'm taking the afternoon off but I will do the tutorial tomorrow!

PhilipVinc (approved):

PhilipVinc:
This looks good to me.
I'll wait for a review by @femtobit too.

The tutorial (to be added in tutorials/ folder and also in the list of tutorials in the `docs/turoails.rst`) would be great, but I think we can merge it later when it's ready.

femtobit:
Yes, sorry about letting you wait but I'd like to look at this PR and the group algebra a bit more closely and didn't get around to it yet. I hope I can take a look soon (definitely in the next days).

chrisrothUT:
No problem at all, take your time. I plan to finish the tutorial today, so maybe we can add that in before the merge

netket/graph/lattice.py
```diff
@@ -20,8 +20,15 @@
...
 
+cutoff_tol = 1e-5
+"""Tolerance for the maximum distance cutoff when computing the sparse distance matrix.
+This is necessary because of floating-point errors when computing the distance in non-trivial 
+lattices.
+"""
+
 
 def get_edges(atoms_positions, cutoff):
+    cutoff = cutoff + cutoff_tol
```
femtobit, line 31: This innocent-looking change changes the time it takes for `Test/Graph/test_graph.py::test_automorphisms` by _a lot_:
```
# before:
2.13s call     Test/Graph/test_graph.py::test_automorphisms
# after:
226.23s call     Test/Graph/test_graph.py::test_automorphisms
```

This is probably because the lattice graphs previously did not recognize all edges. In any case, we should probably fix this test.

By the way, is there a reason `test_automorphism` is only run on the lattices? Also, do we really need this test in the current form? At the moment, it only seems to check that two different `networkx` methods return the same results (as all our graphs use networkx anyway for this).

PhilipVinc: I'm all in for writing better tests.
I don't know who wrote them originally.

netket/graph/symmetry.py
```diff
@@ -80,6 +81,62 @@ def remove_duplicates(self, *, return_inverse=False):
...
+        automorphisms = np.array(automorphisms)
+        for i, perm1 in enumerate(automorphisms):
+            for j, perm2 in enumerate(automorphisms):
+                perm_sq = perm1[perm2]
+                if np.all(perm_sq == np.arange(len(perm_sq))):
+                    inverse[i] = j
+
+        return SymmGroup([self.elems[i] for i in inverse], self.graph)
+
+    def group_algebra(self):
```
femtobit, line 104: I'm not sure "group algebra" is a clear name for this function. My group theory knowledge is somewhat rusty, but (aside from using `g^-1`) doesn't it essentially compute a representation of the group's Cayley table (i.e., it's multiplication table)?

Or is there an established meaning of "group algebra" that coincides with what this function does? Otherwise I think we should try and find a different name.

netket/graph/symmetry.py
```diff
@@ -80,6 +81,62 @@ def remove_duplicates(self, *, return_inverse=False):
         else:
             return group
 
+    def inverse(self):
+        """
+        Returns reordered SymmGroup where the each element is the inverse of
+        the original symmetry element. If g = self[element] and h = self.inverse()[element]
+        gh = I for all elements.
```
femtobit, line 88: Maybe like this?
```suggestion
        the original symmetry element. If :code:`g = self[element]` and :code:`h = self.inverse()[element]`,
        then :code:`gh = product(g, h)` will act as the identity on the sites of the graph, i.e., :code:`np.all(gh(sites) == sites)`.
```

netket/graph/lattice.py
PhilipVinc: There is overlap with what is being discussed [here](https://github.com/netket/netket/pull/633#discussion_r616672597)

netket/graph/symmetry.py
chrisrothUT: It's a Cayley table but one side is the involution of the group. I'm open to renaming it though.

netket/graph/lattice.py
femtobit: I think for now we could honestly just remove that test. At least to some extent the `NetworkX.autmorphisms` method at least is covered in the symmetry group tests as well, so we would not lose significant coverage.

netket/graph/symmetry.py
chrisrothUT: How about flattenen_Cayley(). And something like returns a flattened Cayley table, s.t. for u = flattenen_Cayley()[elem1*n_symm + elem2],  g = self.inverse()[elem1],  and h = self[elem2], u = gh

chrisrothUT: actually it would be u = self[flattenen_Cayley()[elem1*n_symm + elem2]], maybe this is too cumbersome

chrisrothUT: *flattened

femtobit: Now that you mention it: Is there a reason the flattened table is returned? It sounds more generally useful for me to return the `ndim=2` table. Flattening it when needed is easier than the reverse operation.

netket/graph/symmetry.py
```diff
@@ -80,6 +81,62 @@ def remove_duplicates(self, *, return_inverse=False):
...
+            (index, hash_auto[hash(element.tobytes())])
+            for index, element in enumerate(inv_auto)
+            if hash(element.tobytes()) in hash_auto
+        ]
+
+        inds = np.asarray(inds)
+
+        group_algebra[inds[:, 0] // n_symm, inds[:, 0] % n_symm] = inds[:, 1]
+
+        return HashableArray(group_algebra.ravel())
```
femtobit, line 138: I think this should return an array. The only place I'd use `HashableArray` is as an attribute of a frozen dataclass (which includes our models). Otherwise it's not worth the hassle.

netket/graph/symmetry.py
femtobit: Now that you mention it: Is there a reason the flattened table is returned? It sounds more generally useful for me to return the `ndim=2` table. Flattening it when needed is easier than the reverse operation.

chrisrothUT: Okay I've gone ahead and changed the name to flattened_Cayley, and rewrote the description with that in mind

femtobit: Cayley table is I think a bit too much group theory jargon (especially given that it's not exactly the table of `g*h` due to the inverse) for the name. How about just `product_table`?

netket/graph/symmetry.py
femtobit: I find it sufficiently clear now.

netket/graph/symmetry.py
chrisrothUT: product table sounds fine. I guess I could just return an array of n_dim = 2 and flatten/make it Hashable inside of setup()?

netket/graph/symmetry.py
```diff
@@ -80,6 +81,61 @@ def remove_duplicates(self, *, return_inverse=False):
...
+        n_symm = len(automorphisms)
+        inverse = np.zeros([n_symm], dtype=int)
+        automorphisms = np.array(automorphisms)
+        for i, perm1 in enumerate(automorphisms):
+            for j, perm2 in enumerate(automorphisms):
+                perm_sq = perm1[perm2]
+                if np.all(perm_sq == np.arange(len(perm_sq))):
+                    inverse[i] = j
+
+        return SymmGroup([self.elems[i] for i in inverse], self.graph)
```
femtobit, line 102: I wonder which is better: Returning a `SymmGroup` here or just the indices of the inverse elements (i.e., `inverse` directly). The indices could be a bit more generally useful and are harder to reconstruct from `.elems` later.

chrisrothUT: I've vacilated on this. I'm happy to return the indices

femtobit: Yes, let's return the indices.

netket/graph/symmetry.py
chrisrothUT: Okay I'll fix this

netket/graph/symmetry.py
chrisrothUT: Actually I'm just going to have DenseEquivariant take flattened_product_table and complain if it gets something else. GCNN will do the same if product_table is specified, and if it just gets SymmGroup

femtobit: > product table sounds fine. I guess I could just return an array of n_dim = 2 and flatten/make it Hashable inside of setup()?

How about this: The `DenseEquivariant` layer accepts `Union[HashableArray, SymmGroup]` and in setup you either use the array directly or, if the attribute is a `SymmGroup` call something like `group.product_table().ravel()` and work with that?

Then everything is consistent with `DenseSymm`.

chrisrothUT: Okay great, I will do that

netket/graph/lattice.py
PhilipVinc: can you open a PR for this?

femtobit (approved):
This looks pretty good to me now, thanks @chrisrothUT. As far as I am concerned this can be merged.

An example would be nice but since the plan is to add a tutorial, that will be addressed then anyways.

netket/graph/symmetry.py
PhilipVinc: It looks to me this was not reverted to indices?

PhilipVinc: nevermind, i was on an old diff

Test/Models/test_nn.py
```diff
@@ -55,6 +60,41 @@ def test_DenseSymm(symmetries, use_bias):
...
+    sym_op = np.where(inv_pt == chosen_op, 1.0, 0.0)
+
+    v = random.normal(random.PRNGKey(0), [3, n_symm])
+    v_trans = dot(v, sym_op)
+
+    out = ma.apply(pars, v)
+    out_trans = ma.apply(pars, v_trans)
+
+    # output should be involution
+    assert jnp.allclose(dot(out, sym_op.transpose(0, 1)), out_trans)
```
attila-i-szabo, line 95:
```suggestion
    # output should be equivariant
    assert jnp.allclose(dot(out, sym_op), out_trans)
```
see https://github.com/netket/netket/issues/698#issuecomment-834948357 for details

PR #633, opened 2021-04-19
Fix lattice

VolodyaCO:
Couldn't reopen #613. I just put the tolerance of kdtree cutoff inside the get_edges method. And also provided a bit more custom arguments for drawing the graph, while making sure that lattices are 1d or 2d in order to be drawn.


Test/Sampler/test_sampler.py
```diff
@@ -26,7 +26,9 @@
 # Hilbert space of spins from given graph
 hi = nk.hilbert.Spin(s=0.5, N=g.n_nodes)
 ha = nk.operator.Ising(hilbert=hi, graph=g, h=1.0)
-move_op = sum([nk.operator.spin.sigmax(hi, i) for i in range(hi.size)])
+move_op = sum(
+    [nk.operator.LocalOperator(hi, [[0, 1], [1, 0]], i) for i in range(hi.size)]
+)
```
PhilipVinc, line 31: please drop those lines

netket/graph/lattice.py
```diff
@@ -221,6 +224,73 @@ def atoms_coord(self):
...
+    def draw(
+        self,
+        ax=None,
+        figsize: Optional[Tuple[Union[int, float]]] = None,
+        node_color: Optional[str] = "#1f78b4",
+        node_size: Optional[int] = 300,
+        edge_color: Optional[str] = "k",
+        curvature: Optional[float] = 0.2,
+        font_size: Optional[int] = 12,
+        font_color: Optional[str] = "k",
```
PhilipVinc, line 236: `Optional[T]=Union[T,None]`, so I guess those aren't optional?

netket/graph/lattice.py
```diff
@@ -19,11 +19,12 @@
 import itertools
 import networkx as _nx
 import warnings
+from typing import Tuple, Union, Optional
 
 
-def get_edges(atoms_positions, cutoff):
+def get_edges(atoms_positions, cutoff, tol=1e-5):
```
PhilipVinc, line 25: We already discussed this in the other PR.
Can you use this code in here?
https://github.com/netket/netket/blob/17035c8b8b89c401d49acd0b743d8383bef0d0db/netket/graph/lattice.py#L23

PhilipVinc (commented):
Also add a test for the plot function on a bunch of lattices. Simply check that it does not raise errors.

VolodyaCO:
> Also add a test for the plot function on a bunch of lattices. Simply check that it does not raise errors.

@PhilipVinc wouldn't this require to install matplotlib on the test environment?

Test/Sampler/test_sampler.py
VolodyaCO: Maybe I did not rebase my branch with the new master locally because I'm seeing some tests with old code in this PR... I will try to fix this

PhilipVinc:
yes, but that is fine.
Just add it as a test dependency to setup.py.
And anyway I think we're installing it already for some random dependency.

netket/graph/lattice.py
```diff
@@ -19,9 +19,17 @@
 import itertools
 import networkx as _nx
 import warnings
+from typing import Tuple, Union, Optional
+
+cutoff_tol = 1e-5
```
gcarleo, line 24: I think this should be a parameter that is passed as an option at creation time of ```Lattice``` (with a default value of 1.0e-5) in other words, we should leave the possibility to change it to the user

netket/graph/lattice.py
```diff
@@ -19,9 +19,17 @@
...
+"""Tolerance for the maximum distance cutoff when computing the sparse distance matrix.
+This is necessary because of floating-point errors when computing the distance in non-trivial 
+lattices.
+"""
 
 
 def get_edges(atoms_positions, cutoff):
+    cutoff = cutoff + cutoff_tol
     kdtree = cKDTree(atoms_positions)
     dist_matrix = kdtree.sparse_distance_matrix(kdtree, cutoff)
```
gcarleo, line 34: it's really unfortunate that ```sparse_distance_matrix``` does not have a ```tol``` parameter...  I am pretty sure that they have similar issue internally if the cutoff is set up to be too close to the tolerance!

gcarleo (changes requested):

netket/graph/lattice.py
PhilipVinc: I'd keep this as is? 
This is really only a floating-point precision error, I don't see it creating issues in general...

gcarleo: ```numpy.isclose(a, b, rtol=1e-05, atol=1e-08, equal_nan=False)```

I would something like this, passing expliciitly rtol as a parameter. Some lattices might require smaller rtol than our (rather arbitrary chosen) 1.0e-5... the main issue is that in ```sparse_distance_matrix``` we are effectively fixing the atol=1.0e-5, which I think can be rather problematic....

PhilipVinc:
BTW, @VolodyaCO when you have some time after this PR, could you contribute some docstrings to lattice.py? For example i don't understand the difference between `get_edges` and `get_current_edges`. What do they do and what's the difference?

netket/graph/lattice.py
PhilipVinc: ok i agree

PhilipVinc: Let's put it as `distance_atol`, a member in the class and pass it around.

PhilipVinc: I think this is an absolute tolerance rather than a relative?

gcarleo: yes this is an absolute tolerance, unfortunately we can only use an absolute one since ```cKDTree``` does not allow us to pass relative tolerances

VolodyaCO:
Sure, I can do that @PhilipVinc

netket/graph/lattice.py
VolodyaCO: Ok, so I will define `distance_atol` as an argument of the init method for the `Lattice` with a 1e-5 default value, and pass it to the function that uses the `cKDTree`

PhilipVinc:
What's happening?
Can you rebase on master?

VolodyaCO:
Looks like I made a mess. I get some strange diffs when trying to solve merge conflicts for lattice.py when rebasing... The latest changes that I pushed to the forked repo are not shown in the diffs when trying to solve the merge conflicts.

image

netket/graph/lattice.py
VolodyaCO: I opened an [issue](https://github.com/scipy/scipy/issues/13924). Maybe in the future they'll take a look at this.

netket/graph/lattice.py
```diff
@@ -27,8 +28,8 @@
 """
 
 
-def get_edges(atoms_positions, cutoff):
-    cutoff = cutoff + cutoff_tol
+def get_edges(atoms_positions, cutoff, distance_atol):
```
PhilipVinc, line 31:
```suggestion
def get_edges(atoms_positions, cutoff, distance_atol = cutoff_tol):
```

netket/graph/lattice.py
```diff
@@ -83,10 +84,10 @@ def create_points(basis_vectors, extent, atom_coords, pbc):
     return atoms, cellANDlabel_to_site
 
 
-def get_true_edges(basis_vectors, atoms, cellANDlabel_to_site, extent):
+def get_true_edges(basis_vectors, atoms, cellANDlabel_to_site, extent, distance_atol):
```
PhilipVinc, line 87:
```suggestion
def get_true_edges(basis_vectors, atoms, cellANDlabel_to_site, extent, distance_atol = cutoff_tol):
```

VolodyaCO:
@gcarleo @PhilipVinc I think this one is ready

PhilipVinc (approved):

gcarleo (approved):

PR #647, opened 2021-04-22
Added tutorials

chrisrothUT:
This is the PR for tutorials on the G-CNN. I'm not quite done yet but I have some questions.

- Should I do two separate tutorials, one for bipartite lattice (where you just need amplitudes) and one for spin liquids (where you need to pre-optimize phases)? Or just group it into one tutorial?
- Are there any good tools in NetKet for computing things like a structure factor on non-square lattices?
- Any ideas of what to show I could show besides the GS energy?


PhilipVinc:
Maybe @gcarleo could add to the organization https://www.reviewnb.com/#pricing , which is free for us since we are open source and would allow us to comment and review your notebook PR more easily

gcarleo:
done, not sure it's already working

PhilipVinc:
https://app.reviewnb.com/netket/netket/pull/647/

It seems yes to me
Il 22 apr 2021, 18:11 +0200, Giuseppe Carleo ***@***.***>, ha scritto:

PhilipVinc:
@chrisrothUT can you confirm me that you can open the link above and see the interface?

chrisrothUT:
yes, I can see the interface

Tutorials/G-CNN Honeycomb.ipynb
```diff
@@ -0,0 +1,374 @@
+{
```
PhilipVinc, line 1: <p>Remarks:</p><ul><li>I studied group theory so I know what abelian groups are... But I know many who do not. Maybe it's worth mentioning, or just putting a link to wikipedia? </li><li>Does the formula have a small type? Or maybe it's just this thing that renders it badly... it should be $\sum_{\langle I,j \rangle}...$</li></ul>
<br/>

 _Reply via <a href='https://app.reviewnb.com/netket/netket/pull/647/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

PhilipVinc, line 1: <p>this is very well written! Still, if you felt like playing with the groups by hand in code to show how the filters g^{-1}h actually behave and how they affect the inputs for example, I think it would make it even more accessible.</p><p> </p><p>On a high level, I'd try to build by hand the action of the GCNN on one or two inputs and show that the output is equivariant.</p><p> </p><p>Of course, only if you feel like doing it!</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/netket/netket/pull/647/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

PhilipVinc, line 1: <p>I'd say more like...</p><p> </p><p>We sample using Metropolis-Hastings, using the Exchange Rule to propose new states. The exchange rule swaps the spin of two neighbouring sites, therefore it keeps the magnetisation fixed (in this case, 0).</p><p> </p><p>uzing -&gt; using</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/netket/netket/pull/647/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='12'/>

PhilipVinc, line 1: <p>mini batches is a very ML kind of jargon that I'm not sure applies here</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/netket/netket/pull/647/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

PhilipVinc, line 1: <p>you don't need to specify <code>out=None</code>  explicitly.</p><p>However, I'd actually use a logger (json or runtime) and plot the data coming out from them, so the tutorial has a nice plot?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/netket/netket/pull/647/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='17'/>

chrisrothUT: <p>Yes, I'll add a plot of the energy</p>

chrisrothUT: <p>Thanks, I definitely think it would be informative to show that the output is invariant so I'll add that in.</p><p> </p><p>I took your advice and looked at the weights and there's basically no discernible structure. In fact their values are basically random normal. However, this did help me realize that G-CNN isn't initializing the weights as intended, so it was a worthwhile endeavor.</p>

chrisrothUT: <p>yeah, iterations is a better word</p>

chrisrothUT: <p>Yeah, that is a good rewriting of that sentence</p>

chrisrothUT: <p>I thought maybe W^{ij}_{identity} would be larger (in magnitude) on average but even that wasn't the case</p>

chrisrothUT:
I also added a few changes to GCNN here.

I fixed the bug with the initialization, made the default non-linearity SELU, and downscaled the output by sqrt(feature_dim*n_symm). Now the output should have mean 0 and variance 1 at initialization independent of the size of the NN which is a desirable feature.

Tutorials/G-CNN Honeycomb.ipynb
gcarleo, line 1: <p>maybe it'd be worth adding that these energies are for different sizes hence the difference? by the way, maybe we can run the full tutorial for the 3x3 directly, or is it too bad for symmetries? </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/netket/netket/pull/647/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='22'/>

chrisrothUT: Yeah, I may as well run a 3x3 to check against ED and something larger (say 6x6). The 3x3 lattice does have an extra factor of 2 of automorphisms, but as long as the ground state is symmetric w.r.t. that extra symmetry op, the model should still work. If anyone knows where the factor of two comes from let me know.

chrisrothUT:
Okay so there's still this issue that product_table() gets called a lot (maybe 20-30 times) during startup. @femtobit do you care if I make inverse and product table an attribute of SymmGroup? This way they get called once when the SymmGroup is initialized. They take an order of magnitude less time than automorphisms() so I think the cost is negligible

PhilipVinc:
You could always them make lazily cached

```python
class lazyobject:
   @property
   def thingy(self):
       if self.thingy is None:
            setattr(self, "thingy", compute_thungy(self))
       return self.thingy
```

chrisrothUT:
The problem with this is that if I call setattr when I'm not in an init loop i get dataclasses.FrozenInstanceError

PhilipVinc:
I think the correct way to do this is `object.__setattr__(self, "fieldname", value)`

chrisrothUT:
Hmm, I tried it that way and I got the same error.

PhilipVinc:
let me try. Can you give me a script to run?

chrisrothUT:
Should I just push the non-working code?

PhilipVinc:
Fine by me.

PhilipVinc:
A MWE is easier for me to work with

PhilipVinc:
(MNonWorkingE I guess)

chrisrothUT:
Huh, its not breaking when I'm trying to design a minimum working example

chrisrothUT:
Nvm I found the bug. Unsurprisingly I'm the dumb one

netket/graph/symmetry.py
```diff
@@ -101,7 +105,7 @@ def inverse(self):
 
         return inverse
 
-    def product_table(self):
```
PhilipVinc, line 104: Missing documentation on the argument (inverse) of this function

netket/graph/symmetry.py
```diff
@@ -136,6 +141,20 @@ def product_table(self):
 
         return product_table
 
+    def inverse(self):
+        if np.all(self._inverse) == None:
```
PhilipVinc, line 145: why np.all ? shouldn't inverse just be either None or what it must be?

you can check with `if self._inverse is None:`

netket/graph/symmetry.py
```diff
@@ -136,6 +141,20 @@ def product_table(self):
...
+            object.__setattr__(self, "_inverse", self.__inverse())
+
+        return self._inverse
+
+    def product_table(self):
+        if np.all(self._product_table) == None:
```
PhilipVinc, line 151: as above

netket/models/equivariant.py
```diff
@@ -139,6 +138,6 @@ def __call__(self, x_in):
 
         if not self.output_activation == None:
             x = self.output_activation(x)
-        x = jnp.sum(x, axis=-1)
+        x = jnp.sum(x, axis=-1) / np.sqrt(x.shape[-1])
```
PhilipVinc, line 141: why `sqrt`? can you put a comment explaining the choice of normalisation?

Tutorials/G-CNN Honeycomb.ipynb
```diff
@@ -0,0 +1,614 @@
...
+   "id": "74190a36",
+   "metadata": {},
+   "source": [
+    "# Using a group convolutional neural network to learn the ground-state of a symmetric spin model\n",
+    "\n",
+    "The goal of this tutorial is to learn about group convolutional neural networks (G-CNNs), a useful tool for simulating lattices with high symmetry. The G-CNN is a generalization to the convolutional neural network (CNN) to non-abelian symmetry groups (groups that contain at least one pair of non-commuting elements). G-CNNs are a natural fit for lattices that have both point group and translational symmetries, as rotations, reflections and translations don't commute with one-another.\n",
+    "\n",
+    "In this tutorial we will learn the ground state of the antiferromagnetic Heisenberg model on  the honeycomb lattice. The Heisenberg Hamiltonian is defined as follows:\n",
+    "\n",
+    "$$ H = \\sum_{i,j \\in \\langle \\rangle} \\vec{\\sigma}_{i} \\cdot \\vec{\\sigma}_{j},$$\n",
```
PhilipVinc, line 14:
```suggestion
    "$$ H = \\sum_{\\langle i,j\\rangle} \\vec{\\sigma}_{i} \\cdot \\vec{\\sigma}_{j},$$\n",
```

Tutorials/G-CNN Honeycomb.ipynb
```diff
@@ -0,0 +1,614 @@
...
+    "\n",
+    "where $\\vec{\\sigma}_{i}$ are Pauli matrices and $<>$ denotes nearest neighbor interactons. \n",
```
PhilipVinc, line 16:
```suggestion
    "where $\\vec{\\sigma}_{i}$ are Pauli matrices and $<>$ denotes nearest neighbor interactions. \n",
```

Tutorials/G-CNN Honeycomb.ipynb
```diff
@@ -0,0 +1,614 @@
+{
```
PhilipVinc, line 1: <p>This should be cleaned up. It's erroring (or you did not rerun the code)</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/netket/netket/pull/647/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='22'/>

PhilipVinc, line 1: <p>can we have axes labels?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/netket/netket/pull/647/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='31'/>

netket/models/equivariant.py
chrisrothUT: I'll add a comment.

This is basically just having variance scaling in the final layer. This makes sure the GCNN always returns outputs with unit normal distribution at initialization

netket/graph/symmetry.py
chrisrothUT: That is a much better solution, thanks!

Tutorials/G-CNN Honeycomb.ipynb
chrisrothUT: hmm i thought I had ran everything, I'll see what's going on

chrisrothUT: <p>of course</p>

chrisrothUT: <p>Ahh i see, yeah I think this is from running things in the wrong order. I will run the notebook on the GPU and upload that version</p>

netket/graph/symmetry.py
```diff
@@ -101,18 +105,21 @@ def inverse(self):
 
         return inverse
 
-    def product_table(self):
+    def __product_table(self, inverse):
```
femtobit, line 108: Does `inverse` really have to be passed explicitly here? The caching also works if, as before, `self.inverse()` is called in the body of this function. Or am I missing something?

chrisrothUT: It definitely does not. I can change it back

netket/graph/symmetry.py
```diff
@@ -105,20 +105,16 @@ def __inverse(self):
...
         solving the equation u = gh
-
-        Arguments:
-            inverse: The indices of the group involution as returned by self.__inverse()
         """
 
         automorphisms = self.to_array()
-        inverse = automorphisms[inverse].squeeze()
-
+        inverse = automorphisms[self.__inverse()].squeeze()
```
femtobit, line 117: This should be `self.inverse()`, though, so that `__inverse` is really only called when no cached result is available.
```suggestion
        inverse = automorphisms[self.inverse()].squeeze()
```

chrisrothUT: ohh yeah my mistake

PhilipVinc (approved):
This looks good for me now.
I'll let @femtobit give the final word on this.

PhilipVinc:
Actually not!

Please check the output of the website... it seems like the formatting is off. 

To check it you can either download it from the GitHub runs (click on actions->Website->Artefacts and download) or run it yourself by doing 

```
cd netket
cd docs
make html
open _build/html/index.html
```

Schermata 2021-04-26 alle 15 50 15

docs/tutorials.rst
```diff
@@ -33,4 +33,11 @@
     :doc:`Heisenberg model with different variational states <tutorials/Heisenberg1d>`
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+    ---
+
```
PhilipVinc, line 37:
```suggestion
```

docs/tutorials.rst
```diff
@@ -33,4 +33,11 @@
...
     Learning the ground-state of a spin model with different Neural Networks available in NetKet.
+
```
PhilipVinc, line 39:
```suggestion

    ---
    
```

PhilipVinc: I think those two things should fix it

chrisrothUT:
I managed to fix the broken link but there's still a strange formatting bug where the tutorials don't display in a 2x2 grid.

chrisrothUT:
Looks like it is an issue related to the browser as things display correctly in Chrome (but not Safari)

Tutorials/G-CNN_Honeycomb.ipynb
```diff
@@ -0,0 +1,812 @@
+{
```
femtobit, line 1: <p>"Marhall-Perels sign rule" -&gt; "Marshall-Peierls sign rule"</p><p> </p><p>Just out of curiosity: Does the GCNN have difficulties learning the signs if we set <code>sign_rule=False</code>?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/netket/netket/pull/647/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

femtobit, line 1: <p>Maybe say something like "on my CPU ..." or add the approximate CPU/GPU models you've used for these timings as naturally performance will differ a bit between machines (e.g., on my CPU this is more like 8 s/it), though the orders of magnitude should probably remain for most setups.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/netket/netket/pull/647/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

femtobit, line 1: <p><code>_sties</code>   -&gt; <code>_sites</code> <span class="ql-cursor">﻿</span></p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/netket/netket/pull/647/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='24'/>

femtobit, line 1: <blockquote>Based on the standard deviation, it looks like our wavefunction has on the order of 1% error. Not bad for 20 minutes of training!</blockquote><p> </p><p>To make a statement like this you should probably use the <code>error_of_mean</code> statistic, which is not the same as <code>sqrt(variance).</code> For example, the error of the mean goes to zero when the MC sample sizes becomes infinite, the variance will not (except in an exact eigenstate) since it has a physically meaningful value.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/netket/netket/pull/647/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='35'/>

femtobit: <p>Okay, sorry, I guess <code>error_of_mean</code> is also not  really what we want here.</p><p> </p><p>Still, while using the variance as a measure of "distance to an eigenstate" is appropriate, I still think "1% of error" a bit misleading. We do not actually know, a priori, if we are close to the ground state or a local minimum. Not sure what the best way to phrase this is, though.</p>

chrisrothUT: <p>I don’t the G-CNN would have trouble learning a simple sign rule like this (I haven’t tried yet) but it would definitely struggle with more difficult lattices if you optimized the amplitudes+phases together from the start (this is a pretty universal problem).</p><p> </p><p>Usually it’s good practice to optimize the phases first (I.e. return 1j*Im(output)) from the wave function) and then train the amplitudes. With that practice, I’ve gotten G-CNN to learn the signs of Pyrochlore, Kagome, etc., with a little work of picking a good learning rate</p><p> </p>

PhilipVinc: Maybe write this down? it's interesting discussion

chrisrothUT: <p>This a good point, I’ll try to make it more clear</p>

chrisrothUT: <p>Yeah, you’re right that this is probably a statement I can’t make. The correct way to do this is to see how the energy scales with variance. I could actually try that, it would be neat to look at anyway. It might make this section require a couple hours on a GPU but people can always hop in or out when they like</p>

femtobit: Yes, the variance scaling sounds like a good way to check this. If that's too much effort I think it's also fine to look at the variance, just with a bit more cautious language (pointing out that it indicates closeness to an eigenstate).

chrisrothUT: Yeah I added a comment. It’d be nice to have a separate tutorial. It’s just hard to find a problem that is difficult enough to need phase optimization and easy enough for a tutorial

PhilipVinc: Thanks! I know how hard is to write good tutorials, but I think some of those issues are worth just 'throwing' them there for the interested reader. If people care, they'll notice that someone thought about it, else, they'll just ignore it.

Tutorials/G-CNN_Honeycomb.ipynb
```diff
@@ -0,0 +1,1324 @@
...
+        "for en in data[\"Energy\"][\"Mean\"]:\n",
+        "    energy.append(en)\n",
+        "\n",
+        "vmc_energy_18sites = np.mean(np.asarray(energy)[-20:])/18\n",
+        "\n",
+        "ED_energy_18sites = E_gs[0]/18\n",
+        "\n",
+        "print(vmc_energy_18sites)\n",
+        "print(ED_energy_18sites)\n",
+        "print((ED_energy_18sites- vmc_energy_18sites)/ED_energy_18sties)"
```
femtobit, line 525:
```suggestion
        "print((ED_energy_18sites- vmc_energy_18sites)/ED_energy_18sites)"
```

femtobit (approved):
Great, thank you! I think this is now ready to be merged.

PhilipVinc:
I might just have crashed github UI

Tutorials/G-CNN_Honeycomb.ipynb
```diff
@@ -0,0 +1,1324 @@
...
+        {
+          "output_type": "stream",
+          "text": [
+            "\n",
+            "\n",
+            "-156.84739927679757\n",
+            "11.455579683144682\n"
+          ],
+          "name": "stdout"
+        },
```
PhilipVinc, line 1189:
```suggestion
```

Tutorials/G-CNN_Honeycomb.ipynb
chrisrothUT: Agreed!

PR #695, opened 2021-05-07
QGTJacobianPyTree: Make the logic more modular and adapt the tests

inailuig:


Test/Optimizer/test_sr_logic.py
```diff
@@ -308,36 +312,110 @@ def mvt(v, f, params, samples, centered, w):
...
+        apply_fun, e.params, e.samples, model_state, mode, rescale_shift
+    )
+    actual = reassemble(mv(v, centered_oks, diag_shift))
     expected = reassemble_complex(
         e.S_real @ e.v_real_flat + diag_shift * e.v_real_flat, target=e.target
     )
     assert tree_allclose(actual, expected)
 
 
+@pytest.mark.parametrize("holomorphic", [True])
```
PhilipVinc, line 397: that's not much of a parametrisation

setup.py
```diff
@@ -23,6 +23,7 @@
     "flax>=0.3.0, <0.4",
     "orjson~=3.4",
     "optax>=0.0.2, <=0.0.7",
+    "compose>=1.1.1",
```
PhilipVinc, line 26: what is compose?

PhilipVinc: is `compose(a,b)` roughly equivalent to `lambda *args, **kwargs: a(b(*args, **kwargs))` ?

Anyhow, if I am not mistaken it is  not on `conda-forge` which is a bit unfortunate

PhilipVinc: I mean I love this kind of thing. I also think it makes the code cleaner.
I am not opposed to it, but another dependency (which we will have to also bring to conda-forge) I don't know. 
Before merging this I'd like the recipe to be available on conda-forge at least.

Is there any alternative a bit more high profile? Or should we rather we vendor it in `netket.utils` and its tests than add it as a dependency...

netket/optimizer/qgt/qgt_jacobian_pytree_logic.py
```diff
@@ -30,116 +31,133 @@
...
     Returns:
-        The Jacobian matrix ∂/∂pₖ ln Ψ(σⱼ) as a PyTree"""
-    y, vjp_fun = jax.vjp(forward_fn, params, samples)
+        The Jacobian matrix ∂/∂pₖ ln Ψ(σⱼ) as a PyTree
+    """
+    y, vjp_fun = jax.vjp(single_sample(forward_fn), params, samples)
     gr, _ = vjp_fun(np.array(1.0, dtype=jnp.result_type(y)))
     gi, _ = vjp_fun(np.array(-1.0j, dtype=jnp.result_type(y)))
-    return gr, gi
+    return jax.tree_multimap(jax.lax.complex, gr, gi)
```
attila-i-szabo, line 88: This wastes memory: you allocate enough memory to hold `gr_` and `gi_` (which will be quite big in production code!) when you
1. calculate them
2. turn them into a complex number
3. take the same complex numbers apart
4. stack them

I rewrote this part of the logic specifically to get rid of the two steps in the middle.

setup.py
attila-i-szabo: > is `compose(a,b)` roughly equivalent to `lambda *args, **kwargs: a(b(*args, **kwargs))` ?

It's not used for anything else at least... we could just add this definition somewhere and not generate more dependencies

Test/Optimizer/test_sr_logic.py
inailuig: It's only supposed to test RR, RC and holomorphic CC, so it's enough to use the holomorphic example function

PhilipVinc:
Let’s do that
Il 7 mag 2021, 19:02 +0200, Clemens Giuliani ***@***.***>, ha scritto:

netket/optimizer/qgt/qgt_jacobian_pytree_logic.py
inailuig: In my mind when dreaming of the perfect jit compiler this should be optimized away (especially 2 and 3)...

I wanted to implement the real-imag split such that it only reinterprets the data at some point
https://github.com/netket/netket/blob/b8782770a5b1bb772c5ba6d4c0b607e814a51b50/netket/jax/utils.py#L240
which would avoid any copy in 2 and also 3 if done properly

and I also wanted to keep the flexibility of being able to change the implementation of `jacobian_cplx` later on where the most logical thing is that it returns the complex jacobian

However I guess we can do what you suggested for now with a few extra arguments here and there

PhilipVinc: >In my mind when dreaming of the perfect jit compiler this should be optimized away (especially 2 and 3)...

I think that jits right now are not very smart about complex numbers.
However I like your idea! Maybe leave it in a comment?

Alas, from what I understood the best layout for complex numbers is SOA (a la Matlab) and not AOS like we/jax/C/everyone does, because you'd get this kind of things for free...

netket/optimizer/qgt/qgt_jacobian_pytree.py
```diff
@@ -39,7 +39,7 @@ def QGTJacobianPyTree(
         )
```
PhilipVinc, line 39: I don't know who wrote this, but this won't work: the eval shape gets apply_fun which is most likely an HahablePartial and it's not compatible with jax.eval_shape..

attila-i-szabo: I wrote it, I copied it out from somewhere in @inailuig's original autodetection logic without really understanding it (jax magick...)
Can you think of something that would do what I intended to do?

attila-i-szabo: also, does it need something along these lines?
```python
    # pre-apply the model state
    def forward_fn(W, σ):
        return apply_fun({"params": W, **model_state}, σ)
```

PhilipVinc: eval shape basically evaluates the shape and dtype of the result (somehow):
```
def eval_shape(fun, *args) :
     res = fun(*args)
     return res.dtype, res.shape
```

It's just that, annoyingly, it does not support out of the box arbitrary callable but just functions.
There are many workarounds to do it. I

attila-i-szabo: Do you mean something like this?
```python
def f(W,σ):
    return vstate._apply_fun({"params": W, **model_state}, σ)
complex_output = jax.eval_shape(vstate._apply_fun, vstate.parameters, vstate.samples).dtype
```

attila-i-szabo: > also, does it need something along these lines?
> 
> ```python
>     # pre-apply the model state
>     apply_fun({"params": W, **model_state}, σ)
> ```

I wasn't referring to this issue so much as the facility to provide `model_state` (it's done in many places, so I kept copying, but I'm still not sure what it achieves or how you can use it...)

attila-i-szabo: Update: this is exactly the error I caught in https://github.com/netket/netket/issues/701#issuecomment-835834164 so I tried these fixes. Turns out `eval_shape` is perfectly happy with `vstate._apply_fun` as it is, but it does need these extra arguments. See a working fix below.

attila-i-szabo:
https://github.com/netket/netket/blob/53e223b54209941b8a86e12e37e5d0a23b82742d/netket/optimizer/qgt/qgt_jacobian_pytree.py#L38
needs to be changed to 
```python
jax.eval_shape(vstate._apply_fun, {"params": vstate.parameters,**vstate.model_state}, vstate.samples)
```

PR #702, opened 2021-05-10
Symmetry Operations on Lattice.py

chrisrothUT:
This implements translations, rotations and reflections on graphs generated by Lattice.py. 

Let’s define a square grid to show how we use this

`graph = nk.graph.Lattice(basis_vectors = [[1,0],[0,1]], extent=[5,5])`

We can generate a SymmGroup with translations by the basis vectors as follows

`graph.basis_translations()`

In order to generate rotations we need to specify the period. This will generate the C4 rotational symmetry group

`graph.planar_rotations(period=4,axes=(0,1))`

If we specify a period that doesn’t map the lattice to itself we get a value error. Finally we can generate the reflections about a plane

`graph.reflections(axis=0)`

This should work for arbitrary dimensions (I only tested a cube). @femtobit @PhilipVinc @attila-i-szabo can you help me bug hunt?

 `

### DIFF
--- a/netket/graph/lattice.py
+++ b/netket/graph/lattice.py
@@ -371,7 +371,7 @@ class Lattice(NetworkX):
         for hash_coord in ref_coords:
             hash_coord = _np.matmul(hash_coord, self._inv_dims) % 1
             hash_coord = hash_coord - hash_coord // (1 - cutoff_tol)
-            hash_coord = hash((1e5 * _np.around(hash_coord, 5)).astype(int).tobytes())
+            hash_coord = hash((_np.power(10,tol_digits) * _np.around(hash_coord, tol_digits)).astype(int).tobytes())
             if hash_coord in self._hash_positions:
                 perm.append(self._hash_positions[hash_coord])
             else:
--- a/netket/graph/lattice.py
+++ b/netket/graph/lattice.py
@@ -228,7 +228,7 @@ class Lattice(NetworkX):
                 for atom_coord in atoms_coord
             ]
         )
-        if atoms_coord_fractional.min() < 0 or atoms_coord_fractional.max() >= 1:
+        if atoms_coord_fractional.min() < -cutoff_tol or atoms_coord_fractional.max() > 1 + cutoff_tol:
             # Maybe there is another way to state this. I want to avoid that there exists the possibility that two atoms from different cells are at the same position:
             raise ValueError(
                 "atoms must reside inside their corresponding unit cell, which includes only the 0-faces in fractional coordinates."
@@ -290,8 +290,9 @@ class Lattice(NetworkX):
         graph.add_edges_from(edges)
 
         self._inv_dims = _np.linalg.inv(self._lattice_dims)
-        frac_positions = _np.matmul(self._coords, self._inv_dims)
-        int_positions = (1e5 * _np.around(frac_positions, 5)).astype(int)
+        frac_positions = _np.matmul(self._coords,self._inv_dims) % 1
+        frac_positions = frac_positions - frac_positions // 1
+        int_positions = (1e5 * _np.around(frac_positions,5)).astype(int)
         self._hash_positions = {
             hash(element.tobytes()): index
             for index, element in enumerate(int_positions)
@@ -324,7 +325,9 @@ class Lattice(NetworkX):
             perm = []
             for coord in self._coords:
                 hash_coord = coord.copy() + vec
-                hash_coord = _np.matmul(hash_coord, self._inv_dims) % 1
+                hash_coord = _np.matmul(hash_coord,self._inv_dims) % 1 
+                # make sure 1 and 0 are treated the same
+                hash_coord = hash_coord - hash_coord // 1
                 hash_coord = (1e5 * _np.around(hash_coord, 5)).astype(int)
                 hash_coord = hash(hash_coord.tobytes())
                 perm.append(self._hash_positions[hash_coord])
@@ -345,8 +348,9 @@ class Lattice(NetworkX):
         rot_coords[:, axes] = _np.matmul(rot_coords[:, axes], rot_mat)
 
         for hash_coord in rot_coords:
-            hash_coord = _np.matmul(hash_coord, self._inv_dims) % 1
-            hash_coord = (1e5 * _np.around(hash_coord, 5)).astype(int)
+            hash_coord = _np.matmul(hash_coord,self._inv_dims) % 1 
+            hash_coord = hash_coord - hash_coord // 1
+            hash_coord = (1e5*_np.around(hash_coord,5)).astype(int)
             hash_coord = hash(hash_coord.tobytes())
             if hash_coord in self._hash_positions:
                 perm.append(self._hash_positions[hash_coord])
@@ -363,8 +367,11 @@ class Lattice(NetworkX):
         ref_coords[:, axis] = -1 * ref_coords[:, axis]
 
         for hash_coord in ref_coords:
-            hash_coord = _np.matmul(hash_coord, self._inv_dims) % 1
-            hash_coord = hash((1e5 * _np.around(hash_coord, 5)).astype(int).tobytes())
+            hash_coord = _np.matmul(hash_coord,self._inv_dims)%1
+            hash_coord = hash_coord - hash_coord // 1
+            hash_coord = hash(
+                        (1e5 * _np.around(hash_coord, 5)).astype(int).tobytes()
+                    )
             if hash_coord in self._hash_positions:
                 perm.append(self._hash_positions[hash_coord])
             else:
--- a/netket/graph/lattice.py
+++ b/netket/graph/lattice.py
@@ -347,7 +347,7 @@ class Lattice(NetworkX):
 
         for hash_coord in rot_coords:
             hash_coord = _np.matmul(hash_coord, self._inv_dims) % 1
-            hash_coord = (_np.around(frac_positions * 10**tol_digits)).astype(int) % (10**tol_digits)
+            hash_coord = _np.around(frac_positions * 10**tol_digits).astype(int) % (10**tol_digits)
             hash_coord = hash(hash_coord.tobytes())
             if hash_coord in self._hash_positions:
                 perm.append(self._hash_positions[hash_coord])
--- a/netket/graph/lattice.py
+++ b/netket/graph/lattice.py
@@ -27,7 +27,8 @@ from dataclasses import dataclass
 from netket.utils.semigroup import Identity, Element
 from .symmetry import SymmGroup
 
-cutoff_tol = 1e-5
+tol_digits = 5
+cutoff_tol = _np.power(10,-tol_digits)
 """Tolerance for the maximum distance cutoff when computing the sparse distance matrix.
 This is necessary because of floating-point errors when computing the distance in non-trivial 
 lattices.
--- a/netket/graph/lattice.py
+++ b/netket/graph/lattice.py
@@ -294,8 +294,7 @@ class Lattice(NetworkX):
 
         self._inv_dims = _np.linalg.inv(self._lattice_dims)
         frac_positions = _np.matmul(self._coords, self._inv_dims) % 1
-        frac_positions = frac_positions - frac_positions // 1
-        int_positions = _np.around(frac_positions * 10**tol_digits).astype(int)
+        int_positions = _np.around(frac_positions * 10**tol_digits).astype(int)  %  (10**tol_digits)
         self._hash_positions = {
             hash(element.tobytes()): index
             for index, element in enumerate(int_positions)
--- a/netket/graph/lattice.py
+++ b/netket/graph/lattice.py
@@ -351,10 +351,7 @@ class Lattice(NetworkX):
 
         for hash_coord in rot_coords:
             hash_coord = _np.matmul(hash_coord, self._inv_dims) % 1
-            hash_coord = hash_coord - hash_coord // (1 - cutoff_tol)
-            hash_coord = (
-                _np.power(10, tol_digits) * _np.around(hash_coord, tol_digits)
-            ).astype(int)
+            hash_coord = (_np.around(frac_positions * 10**tol_digits)).astype(int) % (10**tol_digits)
             hash_coord = hash(hash_coord.tobytes())
             if hash_coord in self._hash_positions:
                 perm.append(self._hash_positions[hash_coord])
--- a/netket/graph/lattice.py
+++ b/netket/graph/lattice.py
@@ -28,7 +28,7 @@ from netket.utils.semigroup import Identity, Element
 from .symmetry import SymmGroup
 
 tol_digits = 5
-cutoff_tol = _np.power(10,-tol_digits)
+cutoff_tol = _np.power(10.0, -tol_digits)
 """Tolerance for the maximum distance cutoff when computing the sparse distance matrix.
 This is necessary because of floating-point errors when computing the distance in non-trivial 
 lattices.
@@ -295,7 +295,9 @@ class Lattice(NetworkX):
         self._inv_dims = _np.linalg.inv(self._lattice_dims)
         frac_positions = _np.matmul(self._coords, self._inv_dims) % 1
         frac_positions = frac_positions - frac_positions // 1
-        int_positions = (_np.power(10,tol_digits) * _np.around(frac_positions, 5)).astype(int)
+        int_positions = (
+            _np.power(10, tol_digits) * _np.around(frac_positions, 5)
+        ).astype(int)
         self._hash_positions = {
             hash(element.tobytes()): index
             for index, element in enumerate(int_positions)
@@ -331,7 +333,9 @@ class Lattice(NetworkX):
                 hash_coord = _np.matmul(hash_coord, self._inv_dims) % 1
                 # make sure 1 and 0 are treated the same
                 hash_coord = hash_coord - hash_coord // (1 - cutoff_tol)
-                hash_coord = (_np.power(10,tol_digits) * _np.around(hash_coord, tol_digits)).astype(int)
+                hash_coord = (
+                    _np.power(10, tol_digits) * _np.around(hash_coord, tol_digits)
+                ).astype(int)
                 hash_coord = hash(hash_coord.tobytes())
                 perm.append(self._hash_positions[hash_coord])
 
@@ -353,7 +357,9 @@ class Lattice(NetworkX):
         for hash_coord in rot_coords:
             hash_coord = _np.matmul(hash_coord, self._inv_dims) % 1
             hash_coord = hash_coord - hash_coord // (1 - cutoff_tol)
-            hash_coord = (_np.power(10,tol_digits) * _np.around(hash_coord, tol_digits)).astype(int)
+            hash_coord = (
+                _np.power(10, tol_digits) * _np.around(hash_coord, tol_digits)
+            ).astype(int)
             hash_coord = hash(hash_coord.tobytes())
             if hash_coord in self._hash_positions:
                 perm.append(self._hash_positions[hash_coord])
@@ -372,7 +378,11 @@ class Lattice(NetworkX):
         for hash_coord in ref_coords:
             hash_coord = _np.matmul(hash_coord, self._inv_dims) % 1
             hash_coord = hash_coord - hash_coord // (1 - cutoff_tol)
-            hash_coord = hash((_np.power(10,tol_digits) * _np.around(hash_coord, tol_digits)).astype(int).tobytes())
+            hash_coord = hash(
+                (_np.power(10, tol_digits) * _np.around(hash_coord, tol_digits))
+                .astype(int)
+                .tobytes()
+            )
             if hash_coord in self._hash_positions:
                 perm.append(self._hash_positions[hash_coord])
             else:
--- a/netket/graph/lattice.py
+++ b/netket/graph/lattice.py
@@ -345,11 +345,8 @@ class Lattice(NetworkX):
     def rotation_perm(self, period, axes=[0, 1]):
         perm = []
         axes = list(axes)
-        rot_mat = _np.zeros([2, 2])
-        rot_mat[0, 0] = _np.cos(2 * pi / period)
-        rot_mat[1, 0] = -_np.sin(2 * pi / period)
-        rot_mat[0, 1] = _np.sin(2 * pi / period)
-        rot_mat[1, 1] = _np.cos(2 * pi / period)
+        angle = 2*pi/period
+        rot_mat = _np.array([[_np.cos(angle), -_np.sin(angle)], [_np.sin(angle), _np.cos(angle)]])
 
         rot_coords = self._coords.copy()
         rot_coords[:, axes] = _np.matmul(rot_coords[:, axes], rot_mat)
--- a/netket/graph/lattice.py
+++ b/netket/graph/lattice.py
@@ -329,10 +329,7 @@ class Lattice(NetworkX):
                 hash_coord = coord.copy() + vec
                 hash_coord = _np.matmul(hash_coord, self._inv_dims) % 1
                 # make sure 1 and 0 are treated the same
-                hash_coord = hash_coord - hash_coord // (1 - cutoff_tol)
-                hash_coord = (
-                    _np.power(10, tol_digits) * _np.around(hash_coord, tol_digits)
-                ).astype(int)
+                hash_coord = _np.around(hash_coord * 10**tol_digits).astype(int)  %  (10**tol_digits)
                 hash_coord = hash(hash_coord.tobytes())
                 perm.append(self._hash_positions[hash_coord])
 
--- a/netket/graph/lattice.py
+++ b/netket/graph/lattice.py
@@ -26,6 +26,7 @@ from dataclasses import dataclass
 
 from netket.utils.semigroup import Identity, Element
 from .symmetry import SymmGroup
+from netket.utils import HashableArray
 
 tol_digits = 5
 cutoff_tol = _np.power(10.0, -tol_digits)
@@ -294,10 +295,11 @@ class Lattice(NetworkX):
 
         self._inv_dims = _np.linalg.inv(self._lattice_dims)
         frac_positions = _np.matmul(self._coords, self._inv_dims) % 1
-        int_positions = _np.around(frac_positions * 10**tol_digits).astype(int)  %  (10**tol_digits)
+        int_positions = _np.around(frac_positions * 10 ** tol_digits).astype(int) % (
+            10 ** tol_digits
+        )
         self._hash_positions = {
-            hash(element.tobytes()): index
-            for index, element in enumerate(int_positions)
+            HashableArray(element): index for index, element in enumerate(int_positions)
         }
 
         super().__init__(graph)
@@ -329,8 +331,10 @@ class Lattice(NetworkX):
                 hash_coord = coord.copy() + vec
                 hash_coord = _np.matmul(hash_coord, self._inv_dims) % 1
                 # make sure 1 and 0 are treated the same
-                hash_coord = _np.around(hash_coord * 10**tol_digits).astype(int)  %  (10**tol_digits)
-                hash_coord = hash(hash_coord.tobytes())
+                hash_coord = _np.around(hash_coord * 10 ** tol_digits).astype(int) % (
+                    10 ** tol_digits
+                )
+                hash_coord = HashableArray(hash_coord)
                 perm.append(self._hash_positions[hash_coord])
 
             perms.append(tuple(perm))
@@ -339,16 +343,20 @@ class Lattice(NetworkX):
     def rotation_perm(self, period, axes=[0, 1]):
         perm = []
         axes = list(axes)
-        angle = 2*pi/period
-        rot_mat = _np.array([[_np.cos(angle), -_np.sin(angle)], [_np.sin(angle), _np.cos(angle)]])
+        angle = 2 * pi / period
+        rot_mat = _np.array(
+            [[_np.cos(angle), -_np.sin(angle)], [_np.sin(angle), _np.cos(angle)]]
+        )
 
         rot_coords = self._coords.copy()
         rot_coords[:, axes] = _np.matmul(rot_coords[:, axes], rot_mat)
 
         for hash_coord in rot_coords:
             hash_coord = _np.matmul(hash_coord, self._inv_dims) % 1
-            hash_coord = _np.around(hash_coord * 10**tol_digits).astype(int) % (10**tol_digits)
-            hash_coord = hash(hash_coord.tobytes())
+            hash_coord = _np.around(hash_coord * 10 ** tol_digits).astype(int) % (
+                10 ** tol_digits
+            )
+            hash_coord = HashableArray(hash_coord)
             if hash_coord in self._hash_positions:
                 perm.append(self._hash_positions[hash_coord])
             else:
@@ -365,8 +373,10 @@ class Lattice(NetworkX):
 
         for hash_coord in ref_coords:
             hash_coord = _np.matmul(hash_coord, self._inv_dims) % 1
-            hash_coord = _np.around(hash_coord * 10**tol_digits).astype(int)  %  (10**tol_digits)
-            hash_coord = hash(hash_coord.tobytes())
+            hash_coord = _np.around(hash_coord * 10 ** tol_digits).astype(int) % (
+                10 ** tol_digits
+            )
+            hash_coord = HashableArray(hash_coord)
             if hash_coord in self._hash_positions:
                 perm.append(self._hash_positions[hash_coord])
             else:
--- a/netket/graph/lattice.py
+++ b/netket/graph/lattice.py
@@ -283,7 +283,6 @@ class Lattice(NetworkX):
 
         self._coords = dicts_to_array(self._atoms, "r_coord")
         self._lattice_dims = _np.expand_dims(self.extent, 1) * self.basis_vectors
-        self._hash_positions = None
 
         # Order node names
         nodes = sorted(graph.nodes())
@@ -330,7 +329,7 @@ class Lattice(NetworkX):
                 hash_coord = coord.copy() + vec
                 hash_coord = _np.matmul(hash_coord, self._inv_dims) % 1
                 # make sure 1 and 0 are treated the same
-                hash_coord = hash_coord - hash_coord // 1
+                hash_coord = hash_coord - hash_coord // (1 - cutoff_tol)
                 hash_coord = (1e5 * _np.around(hash_coord, 5)).astype(int)
                 hash_coord = hash(hash_coord.tobytes())
                 perm.append(self._hash_positions[hash_coord])
@@ -352,7 +351,7 @@ class Lattice(NetworkX):
 
         for hash_coord in rot_coords:
             hash_coord = _np.matmul(hash_coord, self._inv_dims) % 1
-            hash_coord = hash_coord - hash_coord // 1
+            hash_coord = hash_coord - hash_coord // (1 - cutoff_tol)
             hash_coord = (1e5 * _np.around(hash_coord, 5)).astype(int)
             hash_coord = hash(hash_coord.tobytes())
             if hash_coord in self._hash_positions:
@@ -371,7 +370,7 @@ class Lattice(NetworkX):
 
         for hash_coord in ref_coords:
             hash_coord = _np.matmul(hash_coord, self._inv_dims) % 1
-            hash_coord = hash_coord - hash_coord // 1
+            hash_coord = hash_coord - hash_coord // (1 - cutoff_tol)
             hash_coord = hash((1e5 * _np.around(hash_coord, 5)).astype(int).tobytes())
             if hash_coord in self._hash_positions:
                 perm.append(self._hash_positions[hash_coord])
--- a/netket/graph/lattice.py
+++ b/netket/graph/lattice.py
@@ -20,6 +20,8 @@ import itertools
 import networkx as _nx
 import warnings
 from typing import Tuple, Union, Optional
+from math import pi
+from itertools import product
 
 cutoff_tol = 1e-5
 """Tolerance for the maximum distance cutoff when computing the sparse distance matrix.
@@ -226,6 +228,9 @@ class Lattice(NetworkX):
         new_nodes = {old_node: new_node for new_node, old_node in enumerate(old_nodes)}
         graph = _nx.relabel_nodes(graph, new_nodes)
 
+        self._atoms_coord = dicts_to_array(self._atoms,"r_coord")
+        self._lattice_dims = _np.expand_dims(self.extent,1)*self.basis_vectors
+        
         # Order node names
         nodes = sorted(graph.nodes())
         edges = list(graph.edges())
@@ -245,6 +250,51 @@ class Lattice(NetworkX):
         Coordinates of atoms in the unit cell.
         """
         return self._atoms_coord
+
+    def translation_perm(self):
+      perms = []
+      for vec in self.basis_vectors:
+        perm = []
+        for coord in self._atoms_coord:
+          new_coord = coord.copy() + vec
+          searching = 1
+          while searching:
+            for i, old_coord in enumerate(self._atoms_coord):
+              shift_lattice = product(range(-1,2),repeat=len(self._lattice_dims))
+              for shift in shift_lattice:
+                move_coord = old_coord + _np.sum(_np.expand_dims(_np.asarray(shift),1)*self._lattice_dims,0)
+                if _np.all(_np.isclose(move_coord,new_coord)):
+                  perm.append(i)
+                  searching = 0 
+
+        perms.append(perm)
+      return perms 
+
+    def rotation_perm(self,period,axes=[0,1]):
+      perm = []
+      rot_mat = _np.zeros([2,2])
+      rot_mat[0,0] = _np.cos(2*pi/period)
+      rot_mat[1,0] = -_np.sin(2*pi/period)
+      rot_mat[0,1] = _np.sin(2*pi/period)
+      rot_mat[1,1] = _np.cos(2*pi/period)
+
+      for coord in self._atoms_coord:
+        new_coord = coord.copy()
+        new_coord[axes] = _np.matmul(rot_mat,new_coord[axes])
+
+        searching = 1
+        while searching:
+          for i,old_coord in enumerate(self._atoms_coord):
+            shift_lattice = product(range(-1,2), repeat = len(self._lattice_dims))
+            for shift in shift_lattice:
+              move_coord = old_coord + _np.sum(_np.expand_dims(_np.asarray(shift),1)*self._lattice_dims,0)
+              if _np.all(_np.isclose(new_coord,move_coord)):
+                perm.append(i)
+                searching = 0
+          if searching:
+            raise ValueError("Rotation with the specified period and axes does not map to itself")
+
+      return perm
 
     def draw(
         self,
--- a/netket/graph/lattice.py
+++ b/netket/graph/lattice.py
@@ -365,7 +365,7 @@ class Lattice(NetworkX):
 
         for hash_coord in ref_coords:
             hash_coord = _np.matmul(hash_coord, self._inv_dims) % 1
-            hash_coord = _np.around(frac_positions * 10**tol_digits)).astype(int)  %  (10**tol_digits)
+            hash_coord = _np.around(frac_positions * 10**tol_digits).astype(int)  %  (10**tol_digits)
             hash_coord = hash(hash_coord.tobytes())
             if hash_coord in self._hash_positions:
                 perm.append(self._hash_positions[hash_coord])
--- a/netket/graph/lattice.py
+++ b/netket/graph/lattice.py
@@ -295,7 +295,7 @@ class Lattice(NetworkX):
         self._inv_dims = _np.linalg.inv(self._lattice_dims)
         frac_positions = _np.matmul(self._coords, self._inv_dims) % 1
         frac_positions = frac_positions - frac_positions // 1
-        int_positions = (1e5 * _np.around(frac_positions, 5)).astype(int)
+        int_positions = (_np.power(10,tol_digits) * _np.around(frac_positions, 5)).astype(int)
         self._hash_positions = {
             hash(element.tobytes()): index
             for index, element in enumerate(int_positions)
--- a/netket/graph/lattice.py
+++ b/netket/graph/lattice.py
@@ -290,8 +290,8 @@ class Lattice(NetworkX):
         graph.add_edges_from(edges)
 
         self._inv_dims = _np.linalg.inv(self._lattice_dims)
-        frac_positions = _np.matmul(self._coords,self._inv_dims)
-        int_positions = (1e5 * _np.around(frac_positions,5)).astype(int)
+        frac_positions = _np.matmul(self._coords, self._inv_dims)
+        int_positions = (1e5 * _np.around(frac_positions, 5)).astype(int)
         self._hash_positions = {
             hash(element.tobytes()): index
             for index, element in enumerate(int_positions)
@@ -324,7 +324,7 @@ class Lattice(NetworkX):
             perm = []
             for coord in self._coords:
                 hash_coord = coord.copy() + vec
-                hash_coord = _np.matmul(hash_coord,self._inv_dims) % 1
+                hash_coord = _np.matmul(hash_coord, self._inv_dims) % 1
                 hash_coord = (1e5 * _np.around(hash_coord, 5)).astype(int)
                 hash_coord = hash(hash_coord.tobytes())
                 perm.append(self._hash_positions[hash_coord])
@@ -345,8 +345,8 @@ class Lattice(NetworkX):
         rot_coords[:, axes] = _np.matmul(rot_coords[:, axes], rot_mat)
 
         for hash_coord in rot_coords:
-            hash_coord = _np.matmul(hash_coord,self._inv_dims) % 1 
-            hash_coord = (1e5*_np.around(hash_coord,5)).astype(int)
+            hash_coord = _np.matmul(hash_coord, self._inv_dims) % 1
+            hash_coord = (1e5 * _np.around(hash_coord, 5)).astype(int)
             hash_coord = hash(hash_coord.tobytes())
             if hash_coord in self._hash_positions:
                 perm.append(self._hash_positions[hash_coord])
@@ -363,10 +363,8 @@ class Lattice(NetworkX):
         ref_coords[:, axis] = -1 * ref_coords[:, axis]
 
         for hash_coord in ref_coords:
-            hash_coord = _np.matmul(hash_coord,self._inv_dims)%1
-            hash_coord = hash(
-                        (1e5 * _np.around(hash_coord, 5)).astype(int).tobytes()
-                    )
+            hash_coord = _np.matmul(hash_coord, self._inv_dims) % 1
+            hash_coord = hash((1e5 * _np.around(hash_coord, 5)).astype(int).tobytes())
             if hash_coord in self._hash_positions:
                 perm.append(self._hash_positions[hash_coord])
             else:
--- a/netket/graph/lattice.py
+++ b/netket/graph/lattice.py
@@ -295,9 +295,7 @@ class Lattice(NetworkX):
         self._inv_dims = _np.linalg.inv(self._lattice_dims)
         frac_positions = _np.matmul(self._coords, self._inv_dims) % 1
         frac_positions = frac_positions - frac_positions // 1
-        int_positions = (
-            _np.power(10, tol_digits) * _np.around(frac_positions, 5)
-        ).astype(int)
+        int_positions = _np.around(frac_positions * 10**tol_digits).astype(int)
         self._hash_positions = {
             hash(element.tobytes()): index
             for index, element in enumerate(int_positions)
--- a/netket/graph/lattice.py
+++ b/netket/graph/lattice.py
@@ -278,7 +278,7 @@ class Lattice(NetworkX):
         new_nodes = {old_node: new_node for new_node, old_node in enumerate(old_nodes)}
         graph = _nx.relabel_nodes(graph, new_nodes)
 
-        self._atoms_coord = dicts_to_array(self._atoms, "r_coord")
+        self._coords = dicts_to_array(self._atoms, "r_coord")
         self._lattice_dims = _np.expand_dims(self.extent, 1) * self.basis_vectors
 
         # Order node names
@@ -301,15 +301,23 @@ class Lattice(NetworkX):
         """
         return self._atoms_coord
 
+    @property
+    def coords(self):
+      """
+      Returns list of coordinates of lattice points 
+      """
+
+      return self._coords
+
     def translation_perm(self):
         perms = []
         for vec in self.basis_vectors:
             perm = []
-            for coord in self._atoms_coord:
+            for coord in self._coords:
                 new_coord = coord.copy() + vec
                 searching = 1
                 while searching:
-                    for i, old_coord in enumerate(self._atoms_coord):
+                    for i, old_coord in enumerate(self._coords):
                         shift_lattice = product(
                             range(-1, 2), repeat=len(self._lattice_dims)
                         )
@@ -335,24 +343,28 @@ class Lattice(NetworkX):
         rot_mat[0, 1] = _np.sin(2 * pi / period)
         rot_mat[1, 1] = _np.cos(2 * pi / period)
 
-        for coord in self._atoms_coord:
-            new_coord = coord.copy()
-            new_coord[axes] = _np.matmul(rot_mat, new_coord[axes])
+        rot_coords= _np.ndarray.astype(1e5*_np.around(_np.matmul(_np.asarray(self.coords)[:,axes],rot_mat),5),int)
 
+        rot_hash = {
+            hash(element.tobytes()): index
+            for index, element in enumerate(rot_coords)
+        }
+
+        for i, old_coord in enumerate(self._coords):
             searching = 1
             while searching:
-                for i, old_coord in enumerate(self._atoms_coord):
-                    shift_lattice = product(
-                        range(-1, 2), repeat=len(self._lattice_dims)
+                shift_lattice = product(
+                    range(-1, 2), repeat=len(self._lattice_dims)
+                )
+                for shift in shift_lattice:
+                    move_coord = old_coord + _np.sum(
+                        _np.expand_dims(_np.asarray(shift), 1) * self._lattice_dims,
+                        0,
                     )
-                    for shift in shift_lattice:
-                        move_coord = old_coord + _np.sum(
-                            _np.expand_dims(_np.asarray(shift), 1) * self._lattice_dims,
-                            0,
-                        )
-                        if _np.all(_np.isclose(new_coord, move_coord)):
-                            perm.append(i)
-                            searching = 0
+                    move_coord = hash(_np.ndarray.astype(1e5*_np.around(move_coord,5),int).tobytes())
+                    if move_coord in rot_hash:
+                        perm.append(rot_hash[move_coord])
+                        searching = 0
                 if searching:
                     raise ValueError(
                         "Rotation with the specified period and axes does not map lattice to itself"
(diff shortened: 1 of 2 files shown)


gcarleo:
nice job @chrisrothUT , this is going to be very useful thanks

netket/graph/lattice.py
```diff
@@ -246,6 +301,129 @@ def atoms_coord(self):
         """
         return self._atoms_coord
 
+    def translation_perm(self):
+        perms = []
+        for vec in self.basis_vectors:
```
gcarleo, line 306: please recall that for loops in python are very expensive and should be avoided as much as possible!

netket/graph/lattice.py
```diff
@@ -362,24 +374,30 @@ def rotation_perm(self, period, axes=[0, 1]):
...
-                for i, old_coord in enumerate(self.atoms_coord):
-                    shift_lattice = product(
-                        range(-1, 2), repeat=len(self._lattice_dims)
+                shift_lattice = product(
+                    range(-1, 2), repeat=len(self._lattice_dims)
+                )
+                for shift in shift_lattice:
+                    move_coord = old_coord + _np.sum(
+                        _np.expand_dims(_np.asarray(shift), 1) * self._lattice_dims,
+                        0,
```
attila-i-szabo, line 395: You could eliminate this bit by hashing the fractional coordinates of the sites in the simulation box. You can calculate that by multiplying the position vector with the inverse of the matrix made up of the vectors `basis_vectors[i]*extent[i]` and taking the fractional part. 
This guarantees you find matching sites even if the box is oddly shaped (now you can't know if something just maps outside the 4^d copies you generate) and gets rid of the 4^d-long loop.

attila-i-szabo: Also I tend to think reciprocal lattice vectors and the like will come useful when implementing #658

attila-i-szabo: I really like the hashing solution btw!

attila-i-szabo: Also, since there is a lot of code duplication with the hashing, you might want to have a method that takes a bunch of vectors and outputs their hashes. You might even want to cache the hashes of the original positions!

chrisrothUT: Thanks, that’s an elegant solution

chrisrothUT:
Thanks for all of the help, the code should be fast now. 

How can we make this more user friendly?
- One tricky thing is that the user has to specify the lattice s.t. the origin is a center of rotation. Otherwise the rotation generator will fail. I fell into this trap while I was testing the Hexagonal/Kagome lattice
- Perhaps we should add a way for the user to output the full space group?

femtobit:
> I think this is fine, we just need to document it properly. Standard Wyckoff positions exist for a reason...

Yes, requiring the origin as center of roation seems reasonable to me. Maybe it would be good to have a methods that, given a `Lattice` and site `i`, returns a new `Lattice` with all sites translated in space so that `i` is at the origin? Then users could use whatever coordinates make it easy for them to set up the lattice and just translate it in the end.

PhilipVinc:
My problem with putting stuff inside of classes is that it makes extending things harder.
Because you need to edit the file with the class itself.

Instead if everything is out of a class (functional style) you can always define those objects + dispatch rules and everything works and prototyping is easier.
Then adding methods to a class for simpler construction is another story, that comes after (and i'm in favour of).

attila-i-szabo:
See https://github.com/netket/netket/issues/703#issuecomment-838428813 for a possible out-of-class implementation for point groups, which would also help clean things up.

I think there is a real difference between point and translation groups, though. 
Valid translations are determined by the lattice vectors, and their number by the size of the lattice. It wouldn't make much sense to have an external `translation_group(Graph)` object, because it would have to ask literally everything from the underlying graph, so we would need a large interface for giving this information, which would go unimplemented for stuff other than `Lattice` and `Grid` anyway. 
On the other hand, there is a small handful of possible valid point groups, and it can make sense to pass a subgroup of the true point group of the lattice, so being able to pass one externally is a good thing. Also, doing it externally requires a small interface (https://github.com/netket/netket/issues/703#issuecomment-838428813) that makes sense on its own.

chrisrothUT:
I added some tests and marked this ready for review because I think that the interface is already pretty useful. For example it is pretty easy to generate character tables with only one broken symmetry. These are commonly the low-lying excited states on Heiseneberg models. 

```
L = Lattice()
rots = L.planar_rotations(6) 
trans = L.basis_translations() 
ref = L.reflections() 
symmgroup =  trans @ ref @ rots
eigvals = np.exp(2*1j*pi/6*np.arange(6)) 
eigvals = eigvals.repeat(len(trans)*len(ref))
```

Obviously it would be nice if @attila-i-szabo and @femtobit could help us write some more generally useful software but I think this immediate step is worth it.
 

 



It sounds like @attila-i-szabo may

netket/graph/lattice.py
chrisrothUT: I somehow missed this. Yeah I will fix that

netket/graph/lattice.py
```diff
@@ -20,6 +20,12 @@
...
 import warnings
 from typing import Tuple, Union, Optional
+from math import pi
+from itertools import product
+from dataclasses import dataclass
+
+from netket.utils.semigroup import Identity, Element
+from .symmetry import SymmGroup
 
 cutoff_tol = 1e-5
```
chrisrothUT, line 30:
```suggestion
tol_digits = 5
cutoff_tol = np.power(10,-tol_digits)
```

netket/graph/lattice.py
```diff
@@ -226,13 +281,25 @@ def __init__(
...
         nodes = sorted(graph.nodes())
         edges = list(graph.edges())
         graph = _nx.MultiGraph()
         graph.add_nodes_from(nodes)
         graph.add_edges_from(edges)
 
+        self._inv_dims = _np.linalg.inv(self._lattice_dims)
+        frac_positions = _np.matmul(self._coords, self._inv_dims) % 1
+        frac_positions = frac_positions - frac_positions // 1
+        int_positions = (1e5 * _np.around(frac_positions, 5)).astype(int)
```
chrisrothUT, line 297:
```suggestion
        int_positions = (_np.power(10,tol_digits) * _np.around(frac_positions, 5)).astype(int)
```

netket/graph/lattice.py
```diff
@@ -246,6 +313,111 @@ def atoms_coord(self):
...
+    def translation_perm(self):
+        perms = []
+        for vec in self.basis_vectors:
+            perm = []
+            for coord in self._coords:
+                hash_coord = coord.copy() + vec
+                hash_coord = _np.matmul(hash_coord, self._inv_dims) % 1
+                # make sure 1 and 0 are treated the same
+                hash_coord = hash_coord - hash_coord // (1 - cutoff_tol)
+                hash_coord = (1e5 * _np.around(hash_coord, 5)).astype(int)
```
chrisrothUT, line 333:
```suggestion
                hash_coord = (_np.power(10,tol_digits) * _np.around(hash_coord, 5)).astype(int)
```

netket/graph/lattice.py
```diff
@@ -246,6 +313,111 @@ def atoms_coord(self):
...
+        rot_mat[0, 1] = _np.sin(2 * pi / period)
+        rot_mat[1, 1] = _np.cos(2 * pi / period)
+
+        rot_coords = self._coords.copy()
+        rot_coords[:, axes] = _np.matmul(rot_coords[:, axes], rot_mat)
+
+        for hash_coord in rot_coords:
+            hash_coord = _np.matmul(hash_coord, self._inv_dims) % 1
+            hash_coord = hash_coord - hash_coord // (1 - cutoff_tol)
+            hash_coord = (1e5 * _np.around(hash_coord, 5)).astype(int)
```
chrisrothUT, line 355:
```suggestion
            hash_coord = (_np.power(10,tol_digits) * _np.around(hash_coord, tol_digits)).astype(int)
```

netket/graph/lattice.py
```diff
@@ -246,6 +313,111 @@ def atoms_coord(self):
...
+
+    def reflection_perm(self, axis=0):
+        perm = []
+        ref_coords = self._coords.copy()
+        ref_coords[:, axis] = -1 * ref_coords[:, axis]
+
+        for hash_coord in ref_coords:
+            hash_coord = _np.matmul(hash_coord, self._inv_dims) % 1
+            hash_coord = hash_coord - hash_coord // (1 - cutoff_tol)
+            hash_coord = hash((1e5 * _np.around(hash_coord, 5)).astype(int).tobytes())
```
chrisrothUT, line 374:
```suggestion
            hash_coord = hash((_np.power(10,tol_digits) * _np.around(hash_coord, tol_digits)).astype(int).tobytes())
```

attila-i-szabo:
> I added some tests and marked this ready for review because I think that the interface is already pretty useful. 
> 
> Obviously it would be nice if @attila-i-szabo and @femtobit could help us write some more generally useful software but I think this immediate step is worth it.

Cool, thanks! I agree this is a good thing to have as it is, and would be happy to merge it on `master` after checking the code. I'll do that when I have a bit more time, and also get started on the "more generally useful software."

> For example it is pretty easy to generate character tables with only one broken symmetry. These are commonly the low-lying excited states on Heiseneberg models.
> ```
> L = Lattice()
> rots = L.planar_rotations(6) 
> trans = L.basis_translations() 
> ref = L.reflections() 
> symmgroup =  trans @ ref @ rots
> eigvals = np.exp(2*1j*pi/6*np.arange(6)) 
> eigvals = eigvals.repeat(len(trans)*len(ref))
> ```

This is almost right. Reflections, however, flip the rotation eigenvalue, so you cannot set both of them at the same time (this is reflected even in the [proper character table of D_6](https://en.wikipedia.org/wiki/List_of_character_tables_for_chemically_important_3D_point_groups#Dihedral_groups_(Dn))). You could write
```
> eigvals = np.concatenate((np.exp(2*1j*pi/6*np.arange(6)), [0]*6))
> eigvals = eigvals.repeat(len(trans))
```
to get something that doesn't contain components with different symmetry. 

I'm not a fan of this "vulgar representation theory", though (even if it's quite popular in practice), because you project out much of the Hilbert space that is guaranteed to be degenerate by symmetry. I wonder what we could do with the interface (when it comes to that) to lower the bar to using proper character tables.

chrisrothUT:
> This is almost right. Reflections, however, flip the rotation eigenvalue

I think the fact that I made this mistake is evidence we need something more robust :). We’re working on it (#703)

>  because you project out much of the Hilbert space that is guaranteed to be degenerate by symmetry

I want to understand this point better. In this example, isn’t the reflection symmetry just gone? 

> I wonder what we could do with the interface (when it comes to that) to lower the bar to using proper character tables.

This seems doable, we just need to check if the representation can be expressed as a sum of irreps. Well at least in 2D and 3D where we could store all of the character tables

attila-i-szabo:
> > This is almost right. Reflections, however, flip the rotation eigenvalue
> 
> I think the fact that I made this mistake is evidence we need something more robust :). We’re working on it (#703)

I know, I opened that issue :P 

> > because you project out much of the Hilbert space that is guaranteed to be degenerate by symmetry
> 
> I want to understand this point better. In this example, isn’t the reflection symmetry just gone?

Not quite, it still guarantees that e^{+iπ/3} and e^{-iπ/3} rotation eigenvalues lead to degenerate states. Using the first projection only keeps one of them, but keeping both isn't any harder.

> > I wonder what we could do with the interface (when it comes to that) to lower the bar to using proper character tables.
> 
> This seems doable, we just need to check if the representation can be expressed as a sum of irreps. 

What do you mean? Any set of degenerate states form **one** irrep, so you don't want to sum them. 

> Well at least in 2D and 3D where we could store all of the character tables

I guess, though [it would be a lot to set up](https://en.wikipedia.org/wiki/List_of_character_tables_for_chemically_important_3D_point_groups). Also, this doesn't solve the case of k!=0, where you need to pass an irrep of the corresponding little group, rather than the full point group. Dealing with that in full generality would be a library unto itself.
One idea I had was to allow specifying the irreps in a format closer to how character tables are compiled. Say, you can tell automatically if something is a (say) 60° rotation, and these often belong to the same conjugacy class. So, you could pass something like 'space_group_irrep({'Rot60': -1,...}) and it would be passed to every 60° rotation automatically. The caveat is that you can have similar-looking transformations in different conjugacy classes (e.g. in a pure 2D rotation group, +60 and -60 are distinct), then you have to figure out which one takes which character...

attila-i-szabo:
Anyway, this is much further down the line, and at the end of the day, users who want to get this right will read up on it. Probably the best I can do is writing a very practical summary of the rep theory notes in #700 and deploy it in examples.

chrisrothUT:
> Not quite, it still guarantees that e^{+iπ/3} and e^{-iπ/3} rotation eigenvalues lead to degenerate states. Using the first projection only keeps one of them, but keeping both isn't any harder.

This actually brings up an interesting point. In a sense the cosine representation is just choosing a different representation in the degenerate subspace $a \psi_{π/3} + b \psi_{-π/3}$. There might be a way to make a richer model using this information.

> Probably the best I can do is writing a very practical summary of the rep theory notes in #700 and deploy it in examples when we have the library for it

Agreed, this seems like a tractable goal

attila-i-szabo:
> > Not quite, it still guarantees that e^{+iπ/3} and e^{-iπ/3} rotation eigenvalues lead to degenerate states. Using the first projection only keeps one of them, but keeping both isn't any harder.
> 
> This actually brings up an interesting point. In a sense the cosine representation is just choosing a different representation in the degenerate subspace $a \psi_{π/3} + b \psi_{-π/3}$. There might be a way to make a richer model using this information.

I doubt it. Remember that we are building projection operators rather than wave functions here. Also, you can derive the "official" projection operators from basic representation theory, so you probably have little wiggle room. For instance, I suspect (but don't have time to verify it) that if you put sines instead of the cosines, that would be orthogonal to all irreps (I'm not sure this means it vanishes, but I don't want to explore...). The "vulgar rep theory" projection works though because it's a valid irrep of the rotational subgroup.

attila-i-szabo:
By the way, I've just noticed a really cute feature of the algorithm I proposed for building full-space-group irreps in #700. It goes via summing the irrep you supply over all the different ways it can be deformed by the point group. This I do mostly to account for the different wave vectors generated by rotating/reflecting the supplied one, but in the process it also symmetrises the representation it gets over different symmetries in the same conjugacy class.

**Example:** consider the representation with a fixed rotation eigenvalue @chrisrothUT suggested earlier. Then, conjugating each of these rotations over any of the reflections maps them onto rotating the opposite direction: when you add the corresponding characters together, you get a cosine which is precisely the *E*<sub>1</sub> irrep of *D*<sub>6</sub>.

This means that one gets a lot of leeway with supplying "vulgar irreps" as long as one is sure they are specific enough to map on a single symmetry sector. While I'm not sure I want to advertise this feature, it is beautiful

netket/graph/lattice.py
```diff
@@ -226,13 +282,27 @@ def __init__(
...
         graph = _nx.MultiGraph()
         graph.add_nodes_from(nodes)
         graph.add_edges_from(edges)
 
+        self._inv_dims = _np.linalg.inv(self._lattice_dims)
+        frac_positions = _np.matmul(self._coords, self._inv_dims) % 1
+        frac_positions = frac_positions - frac_positions // 1
+        int_positions = (
+            _np.power(10, tol_digits) * _np.around(frac_positions, 5)
+        ).astype(int)
```
attila-i-szabo, line 300: this assumes `tol_digits == 5`
```suggestion
        int_positions = _np.around(frac_positions * 10**tol_digits)
```

attila-i-szabo: also I just think the power operator is perfectly good for our purposes

netket/graph/lattice.py
```diff
@@ -246,6 +316,119 @@ def atoms_coord(self):
...
+        rot_mat[1, 0] = -_np.sin(2 * pi / period)
+        rot_mat[0, 1] = _np.sin(2 * pi / period)
+        rot_mat[1, 1] = _np.cos(2 * pi / period)
+
+        rot_coords = self._coords.copy()
+        rot_coords[:, axes] = _np.matmul(rot_coords[:, axes], rot_mat)
+
+        for hash_coord in rot_coords:
+            hash_coord = _np.matmul(hash_coord, self._inv_dims) % 1
+            hash_coord = hash_coord - hash_coord // (1 - cutoff_tol)
```
attila-i-szabo, line 359: I suppose you mean
```suggestion
            hash_coord = _np.remainder(hash_coord @ self._inv_dims, 1)
```
dividing with `(1-cutoff_tol)` is particularly odd

netket/graph/lattice.py
```diff
@@ -246,6 +316,119 @@ def atoms_coord(self):
...
+        return tuple(perms)
+
+    def rotation_perm(self, period, axes=[0, 1]):
+        perm = []
+        axes = list(axes)
+        rot_mat = _np.zeros([2, 2])
+        rot_mat[0, 0] = _np.cos(2 * pi / period)
+        rot_mat[1, 0] = -_np.sin(2 * pi / period)
+        rot_mat[0, 1] = _np.sin(2 * pi / period)
+        rot_mat[1, 1] = _np.cos(2 * pi / period)
```
attila-i-szabo, line 352:
```suggestion
        angle = 2*pi/period
        rot_mat = _np.array([[_np.cos(angle), -_np.sin(angle)], [_np.sin(angle), _np.cos(angle)]])
```

netket/graph/lattice.py
```diff
@@ -176,7 +229,10 @@ def __init__(
                 for atom_coord in atoms_coord
             ]
         )
-        if atoms_coord_fractional.min() < 0 or atoms_coord_fractional.max() >= 1:
+        if (
+            atoms_coord_fractional.min() < -cutoff_tol
+            or atoms_coord_fractional.max() > 1 + cutoff_tol
+        ):
```
attila-i-szabo, line 235: I guess you coudl do a hashing trick instead of this too if you wanted. Also, the way it is written now, it won't stop you from duplicating sites near the surface of the unit cell.

netket/graph/lattice.py
attila-i-szabo, line 359: what does this line do??

netket/graph/lattice.py
```diff
@@ -226,13 +282,27 @@ def __init__(
...
         # Order node names
         nodes = sorted(graph.nodes())
         edges = list(graph.edges())
         graph = _nx.MultiGraph()
         graph.add_nodes_from(nodes)
         graph.add_edges_from(edges)
 
+        self._inv_dims = _np.linalg.inv(self._lattice_dims)
+        frac_positions = _np.matmul(self._coords, self._inv_dims) % 1
+        frac_positions = frac_positions - frac_positions // 1
```
attila-i-szabo, line 297: `frac_positions // 1` is guaranteed to be 0 after the previous line...

netket/graph/lattice.py
chrisrothUT: Apparently float(1) % float(1) sometimes returns 1 and sometimes 0. So I need this extra step to make sure one and zero are treated the same way. Another approach would be to add cutoff_tol to the coordinate before taking the fractional part and hashing

chrisrothUT: But yeah, it would be nice to make this cleaner

netket/graph/lattice.py
chrisrothUT: Hmm, yeah that’s a good point. I’ll look into this tomorrow

netket/graph/lattice.py
chrisrothUT: See below, there’s this issue with 1.%1. Returning 1 or 0

netket/graph/lattice.py
attila-i-szabo: Ah I see, this is what happens when you don't have IEEE specs for something... 
The more fundamental problem is that we have this "branch cut" between 0 and 1 that should be treated identically. One way around it would be a "circular hash" that hashes cos(2πx) and sin(2πx) rounded to some precision. Because this is a recurring issue in this class, I'll write a general hashing function to do this when I go off to implement #703. I wouldn't worry about it for now, though, if it passes all the nontrivial tests that you provide, it is robust enough to be deployed.

attila-i-szabo: There is another thorny issue with the whole hashing thing, though. Imagine that a site is right on the border of two of the hashing bins that we define when rounding 10^5*x to an integer (it's not unreasonable in a large superlattice with a complicated enough unit cell). Then, macheps-level errors can move the site from one bin to the next and cause errors. I'll have a think if there is a hash-based fix for this

netket/graph/lattice.py
attila-i-szabo: See below, I think it is good enough for now, and there is more fine-tuning about the whole hashing idea that needs to be done, and will probably become a small module in itself. We can come back to plugging it in here once that's done. Anyway, it's also the user's responsibility not to supply stupid things!

netket/graph/lattice.py
attila-i-szabo:
```suggestion
            hash_coord = hash_coord - hash_coord // 1.0
```
I'd worry that dividing with `1-cutoff_tol` might lead to very slightly negative `hash_coord`s if it was just below 1

attila-i-szabo: > There is another thorny issue with the whole hashing thing, though. Imagine that a site is right on the border of two of the hashing bins that we define when rounding 10^5*x to an integer (it's not unreasonable in a large superlattice with a complicated enough unit cell). Then, macheps-level errors can move the site from one bin to the next and cause errors. I'll have a think if there is a hash-based fix for this

~One dumb but working solution would be to store two sets of hashes, one with twice as large bins as the other. Then, if something falls on the edge of a fine bin, it would be guaranteed to be inside (at a quarter-point) of a coarser one.~ This still fails if your lattice point happens to be at the edges of different-width bins in different coordinates.

Another dumb solution (much in the spirit of hashing) would be using 1/some-large-prime-number as the bin width, so getting the outcome I describe above becomes very unlikely. This just feels ugly though :(

@gcarleo, do you think having a transparent O(N^2) check would be too bad? All-to-all distances can be calculated in a vectorised fashion, so we would be saved from long, slow `for` loops, but memory might still be a concern for large lattices (although, we then go on to use GCNNs that convolve with bigger matrices than these...)

attila-i-szabo:
I think this is ready for merging, even though there are some caveats (see above). @chrisrothUT if you have quick fixes you want to add, feel free. Other than that run `black` so the formatting test passes and I will approve.

netket/graph/lattice.py
```diff
@@ -20,14 +20,67 @@
...
+
+    def __call__(self, sites):
+        sites = _np.take(sites, self.perm)
+
+        return sites
+
+    def __repr__(self):
+        return f"Ref"
+
+
```
femtobit, line 83: Do I see it correctly that these classes are just "rebranded" permutations? In that case, it might be better to use `netket.utils.semigroup.Permutation` (which already implements this).

Maybe it makes sense to subclass the `Permutation` class so that it is immediately visible what kinds of operations a specific element consists of. But I would prefer not duplicate the actual code that applies the permutation.

femtobit (changes requested):

netket/graph/lattice.py
attila-i-szabo: I think the trick is that `__call__` applies the same permutation a number of times, so you can generate all elements of the rotation group (say) from a single rotation. I'm not sure if `Permutation` is happy with this implementation

femtobit: Since a product of permutations is still a permutation, that amounts to specifying the final permutation indices once on creation of the element, right? (One could even construct the permutations and use `product()` repeatedly, which is implemented for the types `Permutation × Permutation -> Permutation`.

netket/graph/lattice.py
gcarleo: I would avoid Nˆ2 operations...

netket/graph/lattice.py
attila-i-szabo: That is true. I'm not sure though whether polishing this code is the best way of spending our efforts, given that implementing #703 will tear all of this up...

attila-i-szabo: Btw, I tried to understand the code of `SymmGroup::__product_table` and I'm a bit puzzled by the fact that you can apparently use hashes to compare very different objects, like `semigroup::Identity` and an identity permutation.

femtobit: The problem is not whether we have an `Identity` element or not.

First, hashes should never be used alone to check for equality of elements; distinct objects can have the same hash, only the converse implication (equal objects have equal hashes) is required.

Second, equivalent group elements (like `Identity` and the identity permutation or a `Translation` by zero etc.) may compare as unequal. This was a deliberate decision to reduce complexity in the implementation so far. Elements of `SymmGroup` (which is tied to a specific graph) should only be compared based on their action on the sites (i.e., whether `e1(nodes)` and `e2(nodes)` are equal). If `__product_table` does not do this, it should be fixed.

femtobit: @attila-i-szabo It seems to me `__product_table` works on `automorphisms = self.to_array()`, which is an array of permutations in all cases.

There might be a problem with the hashing code (even for the permutation indices) since, as mentioned, unequal elements may in principle have the same hash (though this is rare in practice which could make an error here unlikely to spot in a test).

chrisrothUT: > and I'm a bit puzzled by the fact that you can apparently use hashes to compare very different objects, like semigroup::Identity and an identity permutation

There’s no way this is happening/would work. The hashes only compare the array output of SymmGroup which is always a permutation (for Identity as well). 

>Identity sounds like a dangerous object to have!

I think the Identity element makes things a lot nicer, and I don’t get why it would be dangerous.

attila-i-szabo: > There might still be a problem with the hashing code (even for the permutation indices) since, as mentioned, unequal elements may in principle have the same hash (though this is rare in practice which could make an error here unlikely to spot in a test).

I think I can live with this possibility of error, it's highly unlikely any practical case would error out on this...

> @attila-i-szabo It seems to me `__product_table` works on `automorphisms = self.to_array()`, which is an array of permutations in all cases.

That's good, I haven't spotted that. 

However, I would change this going forward. You mentioned a while ago that `PointGroup` in #703 could be given the same product table, etc. routines as `SymmGroup` has. It turns out this would be quite useful – I found a very simple algorithm to generate character tables for the point groups from the times table, which would be simple enough to code that we could expose it to the users. But then, we would replicate a lot of logic between different specific groups, which we might want to move over to `SemiGroup` or some appropriate subclass. However, for this to work generally with a hash-based test, we can't have a jolly-joker `Identity` element...

netket/graph/lattice.py
chrisrothUT: I tried that fix and it also failed. What if we changed cutoff tol to 10^(-8). Would that make everyone more comfortable?

netket/graph/lattice.py
chrisrothUT: @femtobit I agree with that idea but remember permutation forces the label to be Permutation() which we don’t want. I could change permutation to take a user specified label perhaps?

femtobit: > However, for this to work generally with a hash-based test,

I am not sure what you mean. There is no possibility of working solely with a "hash-based" test in Python, one always has to equality-compare if the hashes match.

If you argue that all elements of a `Semigroup` should have the same type (is this the issue you have with the identity object?), that would be a very strong restriction: Even without the `Identity`, `Translation`s, `Rotation`s, and their mixed products (which have type `Composite`) all have different types.

netket/graph/lattice.py
attila-i-szabo: Let's get around floating point weirdness!
```suggestion
        frac_positions = frac_positions - frac_positions // 1
        int_positions = (_np.around(frac_positions * 10**tol_digits)).astype(int) % (10**tol_digits)
```

attila-i-szabo: this is integer arithmetics, so `%` is guaranteed to do what we want

netket/graph/lattice.py
```diff
@@ -246,6 +316,119 @@ def atoms_coord(self):
...
+        rot_coords = self._coords.copy()
+        rot_coords[:, axes] = _np.matmul(rot_coords[:, axes], rot_mat)
+
+        for hash_coord in rot_coords:
+            hash_coord = _np.matmul(hash_coord, self._inv_dims) % 1
+            hash_coord = hash_coord - hash_coord // (1 - cutoff_tol)
+            hash_coord = (
+                _np.power(10, tol_digits) * _np.around(hash_coord, tol_digits)
+            ).astype(int)
```
attila-i-szabo, line 357: and here too
```suggestion
            hash_coord = (_np.around(frac_positions * 10**tol_digits)).astype(int) % (10**tol_digits)
```

chrisrothUT: Nice fix, this should work!

netket/graph/lattice.py
femtobit: > I agree with that idea but remember permutation forces the label to be Permutation() which we don’t want. I could change permutation to take a user specified label perhaps?

Maybe subclassing `Permutation` is an option? Even something like
```
class Translation(Permutation):
    pass
```
might be enough for the labeling. (Though this will "decay" to a `Permutatrion` as soon as you take a product or similar unless you provide a custom `product` method as well.)

attila-i-szabo: > > However, for this to work generally with a hash-based test,
> 
> I am not sure what you mean. There is no possibility of working solely with a "hash-based" test in Python (unless of course the objects you are working with adhere to a stronger contract than the usual Python convention, but that's not easy to achieve in the general case -- also, a one-to-one mapping between objects and hash values is no longer "hashing" in the sense of the word [well, there is "perfect hashing", but this line of thought is getting to theoretical maybe for what amounts to a question of naming], its essentially "encoding" then), one always has to equality-compare if the hashes match.

Sure, but I am pragmatic about this – it's highly unlikely any solely hash-based test would return a wrong result. We can of course run a test on the hashtables we build to make sure they have as many entries as we expect, but I wouldn't slow code down with actual equality testing.

> If you argue that all elements of a `Semigroup` should have the same type (is this the issue you have with the identity object?), that would be a very strong restriction: Even without the `Identity`, `Translation`s, `Rotation`s, and their mixed products (which have type `Composite`) all have different types.

OK, I'll have to think about this a bit more. I'm not terribly worried about `Translation`s and `Rotation`s being a different type, because we only ever multiply them as permutations of a graph. Maybe we can have a `Group` class that is required to produce a canonical representation of its elements that can be compared directly (with or without a hash in between)

netket/graph/lattice.py
chrisrothUT: I can’t commit this one for some strange reason

netket/graph/lattice.py
chrisrothUT: Why would it break if we overwrite the __repr__ function?

netket/graph/lattice.py
attila-i-szabo: it seems github decided that this file changed so much with the previous commit that it can't slot this change in... I'll do it

netket/graph/lattice.py
```diff
@@ -226,13 +282,25 @@ def __init__(
...
         nodes = sorted(graph.nodes())
         edges = list(graph.edges())
         graph = _nx.MultiGraph()
         graph.add_nodes_from(nodes)
         graph.add_edges_from(edges)
 
+        self._inv_dims = _np.linalg.inv(self._lattice_dims)
+        frac_positions = _np.matmul(self._coords, self._inv_dims) % 1
+        frac_positions = frac_positions - frac_positions // 1
+        int_positions = _np.around(frac_positions * 10**tol_digits).astype(int)
```
attila-i-szabo, line 298:
```suggestion
        int_positions = _np.around(frac_positions * 10**tol_digits).astype(int)  %  (10**tol_digits)
```

netket/graph/lattice.py
```diff
@@ -246,6 +314,113 @@ def atoms_coord(self):
...
+        for vec in self.basis_vectors:
+            perm = []
+            for coord in self._coords:
+                hash_coord = coord.copy() + vec
+                hash_coord = _np.matmul(hash_coord, self._inv_dims) % 1
+                # make sure 1 and 0 are treated the same
+                hash_coord = hash_coord - hash_coord // (1 - cutoff_tol)
+                hash_coord = (
+                    _np.power(10, tol_digits) * _np.around(hash_coord, tol_digits)
+                ).astype(int)
```
attila-i-szabo, line 336:
```suggestion
                hash_coord = _np.around(hash_coord * 10**tol_digits).astype(int)  %  (10**tol_digits)
```

netket/graph/lattice.py
```diff
@@ -246,6 +314,113 @@ def atoms_coord(self):
...
+        axes = list(axes)
+        angle = 2*pi/period
+        rot_mat = _np.array([[_np.cos(angle), -_np.sin(angle)], [_np.sin(angle), _np.cos(angle)]])
+
+        rot_coords = self._coords.copy()
+        rot_coords[:, axes] = _np.matmul(rot_coords[:, axes], rot_mat)
+
+        for hash_coord in rot_coords:
+            hash_coord = _np.matmul(hash_coord, self._inv_dims) % 1
+            hash_coord = (_np.around(frac_positions * 10**tol_digits)).astype(int) % (10**tol_digits)
```
attila-i-szabo, line 354:
```suggestion
            hash_coord = _np.around(frac_positions * 10**tol_digits).astype(int) % (10**tol_digits)
```

netket/graph/lattice.py
```diff
@@ -246,6 +314,113 @@ def atoms_coord(self):
...
+        ref_coords[:, axis] = -1 * ref_coords[:, axis]
+
+        for hash_coord in ref_coords:
+            hash_coord = _np.matmul(hash_coord, self._inv_dims) % 1
+            hash_coord = hash_coord - hash_coord // (1 - cutoff_tol)
+            hash_coord = hash(
+                (_np.power(10, tol_digits) * _np.around(hash_coord, tol_digits))
+                .astype(int)
+                .tobytes()
+            )
```
attila-i-szabo, line 377:
```suggestion
            hash_coord = _np.around(frac_positions * 10**tol_digits)).astype(int)  %  (10**tol_digits)
            hash_coord = hash(hash_coord.tobytes())
```

attila-i-szabo (commented):
A bunch of changes to calculating hash coordinates

netket/graph/lattice.py
femtobit: @chrisrothUT It wouldn't break, just that `product(SubclassOfPermutation, SubclassOfPermutation)` would be of the (plain) type `Permutation`. OTOH, the current code I think would produce `product(lattice.Translation, lattice.Translation) -> Composite`, which prints as `T(...) @ T(...)`.

femtobit: > Sure, but I am pragmatic about this – it's highly unlikely any solely hash-based test would return a wrong result.

@attila-i-szabo I'd be a bit careful with that in Python, for example `hash(-1) == hash(-2) == -2` and if we have elements containing tuples, it's not that hard to imagine a case where, say, `(1, -1, 3)` and `(1, -2, 3)` could be hashed (which of course also collide).

But, anyways, this discussion is mostly academic. Comparing elements by hash and only comparing the elements if the hashes happen to match is easy to do and not even a performance problem, really. (And I do argue we should do that if we care about the correctness of the result.)

attila-i-szabo:
There are some mismatched partheses... fixing time!

netket/graph/lattice.py
```diff
@@ -246,6 +313,106 @@ def atoms_coord(self):
...
+        return tuple(perm)
+
+    def reflection_perm(self, axis=0):
+        perm = []
+        ref_coords = self._coords.copy()
+        ref_coords[:, axis] = -1 * ref_coords[:, axis]
+
+        for hash_coord in ref_coords:
+            hash_coord = _np.matmul(hash_coord, self._inv_dims) % 1
+            hash_coord = _np.around(frac_positions * 10**tol_digits)).astype(int)  %  (10**tol_digits)
```
attila-i-szabo, line 368:
```suggestion
            hash_coord = _np.around(frac_positions * 10**tol_digits).astype(int)  %  (10**tol_digits)
```

netket/graph/lattice.py
attila-i-szabo: > @attila-i-szabo I'd be a bit careful with that in Python, for example `hash(-1) == hash(-2) == -2` and if we have elements containing tuples, it's not that hard to imagine a case where, say, `(1, -1, 3)` and `(1, -2, 3)` (which of course also collide) represent distinct group elements.

@femtobit Ouch, I would've expected better from python. The good thing is that we usually compare integer arrays and the other 8 bytes help stir things up a lot.
 
> But, anyways, this discussion is mostly academic. Comparing elements by hash and only comparing the elements if the hashes happen to match is easy to do and not even a performance problem, really. (And I do argue we should do that if we care about the correctness of the result.)

The problem is not so much direct equality testing (I wouldn't do that via a hash), but matching large amounts of stuff quickly (so as to avoid N^2 algorithms that @gcarleo doesn't like, nor do I). For that, hash collisions would be *very* annoying

femtobit: > The problem is not so much direct equality testing (I wouldn't do that via a hash), but matching large amounts of stuff quickly

If that is about the `hash_coordinates`: If the coordinates themselves can be both hashed and equality-compare (you are rounding the before you compare them, from what I am seeing, right?), why use a `dict` of the form `{hash: position}`? Python `dict`s are using the hash of the key for lookup and equality compare if there is a collision.

Thus, can't we just have
```
self._hash_positions = {
    HashableArray(elements): index
    for index, element in enumerate(int_positions)
}        
```
(or use `elements.tobytes()` as key, if you want, both should work) instead of using `hash(elements.tobytes())` as key?
That should have essentially the same lookup efficiency (`dict` hashes the key anyways) and is more likely to be correct in edge cases.

attila-i-szabo: > If that is about the `hash_coordinates`: 

It's a more general pattern: we also use hash-based matching in `__product_table`, which might proliferate further.

> If the coordinates themselves can be both hashed and equality-compared properly (you are rounding them before you compare them, from what I am seeing, right?), why use a `dict` of the form `{hash: index}`? Python `dict`s are already using the hash of the key for lookup and properly handle hash collisions if they occur.
> 
> Thus, can't we just have
> 
> ```
> self._hash_positions = {
>     HashableArray(elements): index
>     for index, element in enumerate(int_positions)
> }        
> ```
> 
> (or use `elements.tobytes()` as key, if you want, both should work) instead of using `hash(elements.tobytes())` as key?
> That should have essentially the same lookup efficiency (`dict` hashes the key anyways) and is more likely to be correct in edge cases.

This is great! I wasn't sure about the efficiency of using a `dict` with complicated keys, but if it works well, then this is way cleaner than handling hashes ourselves!

attila-i-szabo: I guess we can carry on this discussion more generally when it gets to implementing #703, for now, @chrisrothUT should we switch to @femtobit's suggestion?

fabienalet:
''Envoyé depuis mon smartphone Samsung Galaxy.
-------- Message d'origine --------De : Attila Szabó ***@***.***> Date : 13/05/2021  16:16  (GMT+01:00) À : netket/netket ***@***.***> Cc : Fabien Alet ***@***.***>, Mention ***@***.***> Objet : Re: [netket/netket] Symmetry Operations on Lattice.py (#702) 

@attila-i-szabo commented on this pull request.







In netket/graph/lattice.py:

> +                perm.append(self._hash_positions[hash_coord])

+            else:

+                    "Rotation with the specified period and axes does not map lattice to itself"

+                )

+

+        return tuple(perm)

+

+    def reflection_perm(self, axis=0):

+        perm = []

+        ref_coords = self._coords.copy()

+        ref_coords[:, axis] = -1 * ref_coords[:, axis]

+

+        for hash_coord in ref_coords:

+            hash_coord = _np.matmul(hash_coord, self._inv_dims) % 1

+            hash_coord = _np.around(frac_positions * 10**tol_digits)).astype(int)  %  (10**tol_digits)





⬇️ Suggested change

-            hash_coord = _np.around(frac_positions * 10**tol_digits)).astype(int)  %  (10**tol_digits)

+            hash_coord = _np.around(frac_positions * 10**tol_digits).astype(int)  %  (10**tol_digits)







—You are receiving this because you were mentioned.Reply to this email directly, view it on GitHub, or unsubscribe.

[

{

***@***.***": "http://schema.org",

***@***.***": "EmailMessage",

"potentialAction": {

***@***.***": "ViewAction",

"target": "https://github.com/netket/netket/pull/702#pullrequestreview-658948250",

"url": "https://github.com/netket/netket/pull/702#pullrequestreview-658948250",

"name": "View Pull Request"

},

"description": "View this Pull Request on GitHub",

"publisher": {

***@***.***": "Organization",

"name": "GitHub",

"url": "https://github.com"

}

}

]

netket/graph/lattice.py
```diff
@@ -246,6 +313,106 @@ def atoms_coord(self):
...
+        axes = list(axes)
+        angle = 2*pi/period
+        rot_mat = _np.array([[_np.cos(angle), -_np.sin(angle)], [_np.sin(angle), _np.cos(angle)]])
+
+        rot_coords = self._coords.copy()
+        rot_coords[:, axes] = _np.matmul(rot_coords[:, axes], rot_mat)
+
+        for hash_coord in rot_coords:
+            hash_coord = _np.matmul(hash_coord, self._inv_dims) % 1
+            hash_coord = _np.around(frac_positions * 10**tol_digits).astype(int) % (10**tol_digits)
```
attila-i-szabo, line 350:
```suggestion
            hash_coord = _np.around(hash_coord * 10**tol_digits).astype(int) % (10**tol_digits)
```

netket/graph/lattice.py
```diff
@@ -246,6 +313,106 @@ def atoms_coord(self):
...
+        return tuple(perm)
+
+    def reflection_perm(self, axis=0):
+        perm = []
+        ref_coords = self._coords.copy()
+        ref_coords[:, axis] = -1 * ref_coords[:, axis]
+
+        for hash_coord in ref_coords:
+            hash_coord = _np.matmul(hash_coord, self._inv_dims) % 1
+            hash_coord = _np.around(frac_positions * 10**tol_digits).astype(int)  %  (10**tol_digits)
```
attila-i-szabo, line 368:
```suggestion
            hash_coord = _np.around(hash_coord * 10**tol_digits).astype(int)  %  (10**tol_digits)
```

fabienalet:
'   jjjEnvoyé depuis mon smartphone Samsung Galaxy.
-------- Message d'origine --------De : Attila Szabó ***@***.***> Date : 13/05/2021  16:16  (GMT+01:00) À : netket/netket ***@***.***> Cc : Fabien Alet ***@***.***>, Mention ***@***.***> Objet : Re: [netket/netket] Symmetry Operations on Lattice.py (#702) 

@attila-i-szabo commented on this pull request.







In netket/graph/lattice.py:

> +                perm.append(self._hash_positions[hash_coord])

+            else:

+                    "Rotation with the specified period and axes does not map lattice to itself"

+                )

+

+       '    return tuple(perm)

+

+    def reflection_perm(self, axis=0):

+        perm = []

+        ref_coords = self._coords.copy()

+        ref_coords[:, axis] = -1 * ref_coords[:, axis]

+

+        for hash_coord in ref_coords:

+            hash_coord = _np.matmul(hash_coord, self._inv_dims) % 1

+            hash_coord = _np.around(frac_positions * 10**tol_digits)).astype(int)  %  (10**tol_digits)





⬇️ Suggested change

-            hash_coord = _np.around(frac_positions * 10**tol_digits)).astype(int)  %  (10**tol_digits)

+            hash_coord = _np.around(frac_positions * 10**tol_digits).astype(int)  %  (10**tol_digits)







—You are receiving this because you were mentioned.Reply to this email directly, view it on GitHub, or unsubscribe.

[

{

***@***.***": "http://schema.org",

***@***.***": "EmailMessage",

"potentialAction": {

***@***.***": "ViewAction",

"target": "https://github.com/netket/netket/pull/702#pullrequestreview-658948250",

"url": "https://github.com/netket/netket/pull/702#pullrequestreview-658948250",

"name": "View Pull Request"

},

"description": "View this Pull Request on GitHub",

"publisher": {

***@***.***": "Organization",

"name": "GitHub",

"url": "https://github.com"

}

}

]

netket/graph/lattice.py
chrisrothUT: Yeah that works. I guess we can leave the dataclasses as is and address that on the next PR?

attila-i-szabo: I think so – they work, they add something compared to simple `Permutation`s and we'll probably eliminate them soon, so I wouldn't worry. I hope the tests with the new `float→int` logic will pass, so we can merge this after you change the hashing logic

chrisrothUT: Yeah the new int logic worked and it’s definitely a better solution. I think “always use ints when comparing hashes” might be a new rule of mine.

attila-i-szabo: That is generally a good rule for anything that hinges on equality testing or reproducible results in general, integer arithmetics is much more reliable and predictable. (Though this might just show I've been exposed too much to theoretical computer scientists who think floating-point numbers are the work of the devil...)

attila-i-szabo:
Okay, I think we're done here. If @femtobit signs off on the superfluous permutation objects, I'm happy to merge this.

femtobit (approved):
I'm also fine with merging the symmetry operations in the current state and revisit them in connection with #703. And I am happy the coordinate-index mapping is now collision-safe as a side-effect of our hashing discussion.

PhilipVinc:
I think I answered the wrong thread..

I am sorry I did not have the time to review this yet but there’s a bunch of things I’d like to raise.

I’ll try to do that tomorrow
Il 13 mag 2021, 17:49 +0200, Damian Hofmann ***@***.***>, ha scritto:

attila-i-szabo:
Oops, we've just merged it :/ I think the way we ended up viewing it is that it is a stopgap to provide some useful functionality for GCNNs, and a neat API would be designed in connection with #703.

chrisrothUT:
We can definitely unmerge the branch, especially if there’s something that really bugs you.

In any case, this is just the same SymmGroup stuff as was already implemented on Grid.py and a few more bug fixes and more rigorous testing, it doesn’t really change the API at all. 

I will say this, we may really want to think hard about floating point errors and come up with a more systematic way to deal with them (this float-> round-> int-> hash solution seems pretty robust).

attila-i-szabo:
We do add a bit of API with `planar_rotations` etc, and most of the improvements were really improvements to that. The question is whether we mind putting something on `master` that is meant to be short-lived (because of #703), or whoever needs it urgently can have it from this branch. In any case, we don't want this to be an integral part of v3.0, because that would tie our hands with bigger API changes.

chrisrothUT:
`We do add a bit of API with planar_rotations etc`

True, I just meant that in the sense that this stuff already exists for other subclasses of NetworkX i.e. grid.

IMO these features seem useful enough that they’re worth pushing through even in their somewhat unfinished state. Anyone who wants to use VMC with NQS to understand real systems is going to want to explore excited states. I certainly plan to use them immediately.

netket/graph/lattice.py
```diff
@@ -20,14 +20,68 @@
...
+from math import pi
+from itertools import product
+from dataclasses import dataclass
 
-cutoff_tol = 1e-5
+from netket.utils.semigroup import Identity, Element
+from .symmetry import SymmGroup
+from netket.utils import HashableArray
+
+tol_digits = 5
```
PhilipVinc, line 31: It's confusing to have two quantities that are botgh 'editable'. We should stick with one.
If it makes sense to edit `tol_digits` instead of `cutoff_tol` due to implementation details, i'm okay with switching to that but then the other one should disappear or be made internal (prepend an underscore).

netket/graph/lattice.py
```diff
@@ -176,7 +230,10 @@ def __init__(
                 for atom_coord in atoms_coord
             ]
         )
-        if atoms_coord_fractional.min() < 0 or atoms_coord_fractional.max() >= 1:
```
PhilipVinc, line 179: this is about the constructor of Lattice.

If we are moving away from cutoff_tol and towards digits, then we should change this in the constructor too and use the one passed in the constructor throughout the class methods.

However, are we sure we want to use digits instead of tolerance?
We can always extract the number of digits by using a `log10` of the tolerance, and it's more familiar for people to use then the number of digits.

I'd prefer to stick with `cutoff_tol` and compute the number of digits unless there is a compelling reason not to do so.

netket/graph/lattice.py
```diff
@@ -176,7 +230,10 @@ def __init__(
...
+        if (
+            atoms_coord_fractional.min() < -cutoff_tol
```
PhilipVinc, line 234: I don't undestand this.
 the check before was `atoms_coords_fractional.min < 0` equivalent to saying if `atom_cpoords_frac` is negative.
Now you are checking that it should be larger than a cutoff.

Why is that? In general floating point math has rounding errors but does not get the sign wrong easily. 
Did you have a case that would break otherwise? (if so, is it in the tests?)

netket/graph/lattice.py
```diff
@@ -226,13 +283,25 @@ def __init__(
...
+        self._lattice_dims = _np.expand_dims(self.extent, 1) * self.basis_vectors
+
         # Order node names
         nodes = sorted(graph.nodes())
         edges = list(graph.edges())
         graph = _nx.MultiGraph()
         graph.add_nodes_from(nodes)
         graph.add_edges_from(edges)
 
+        self._inv_dims = _np.linalg.inv(self._lattice_dims)
```
PhilipVinc, line 296: can we get a docstring on this field, even if internal?

netket/graph/lattice.py
```diff
@@ -226,13 +283,25 @@ def __init__(
         new_nodes = {old_node: new_node for new_node, old_node in enumerate(old_nodes)}
         graph = _nx.relabel_nodes(graph, new_nodes)
 
+        self._coords = dicts_to_array(self._atoms, "r_coord")
```
PhilipVinc, line 286: Docstring. Maybe specify What is the basis of those coords

netket/graph/lattice.py
```diff
@@ -226,13 +283,25 @@ def __init__(
...
         edges = list(graph.edges())
         graph = _nx.MultiGraph()
         graph.add_nodes_from(nodes)
         graph.add_edges_from(edges)
 
+        self._inv_dims = _np.linalg.inv(self._lattice_dims)
+        frac_positions = _np.matmul(self._coords, self._inv_dims) % 1
+        int_positions = _np.around(frac_positions * 10 ** tol_digits).astype(int) % (
+            10 ** tol_digits
+        )
```
PhilipVinc, line 300: If I understand this correctly this is a poor's man fixed point decimal number implementaiton. Is that right?
I have the following comments:

Quoting from numpy documentation 
```python
np.around uses a fast but sometimes inexact algorithm to round floating-point datatypes. 
```
Can we have an *exact* algorithm? Or, if not, can you tell me that this won't introduce errors or bugs? Or, if not, can't we add an extra digit of precision and then throw it away just to be safe?
Is this maybe linked to your 1 % 1 error?

Regardless, a comment saying what you are doing for whoever reads the code in the future without access to the PR discussion is welcome.
Especfially why we are doing that.

netket/graph/lattice.py
```diff
@@ -246,6 +315,114 @@ def atoms_coord(self):
         """
         return self._atoms_coord
 
+    @property
+    def coords(self):
```
PhilipVinc, line 319: type annotation missing

netket/graph/lattice.py
```diff
@@ -246,6 +315,114 @@ def atoms_coord(self):
...
+        """
+        Returns list of coordinates of lattice points
+        """
+
+        return self._coords
+
+    def translation_perm(self):
```
PhilipVinc, line 326: Docstring missing and type annotation missing. 
Can you write what this method does?

Regardless, i think this is an internal method? If so it should be prepended with an underscore.

netket/graph/lattice.py
```diff
@@ -246,6 +315,114 @@ def atoms_coord(self):
...
+        perms = []
+        for vec in self.basis_vectors:
+            perm = []
+            for coord in self._coords:
+                hash_coord = coord.copy() + vec
+                hash_coord = _np.matmul(hash_coord, self._inv_dims) % 1
+                # make sure 1 and 0 are treated the same
```
PhilipVinc, line 333: Great adding a comment, but i don't get it.
Can you explain better?

PhilipVinc: I think you mean that you are trying to keep only the decimal part?

netket/graph/lattice.py
```diff
@@ -246,6 +315,114 @@ def atoms_coord(self):
...
+                hash_coord = _np.around(hash_coord * 10 ** tol_digits).astype(int) % (
+                    10 ** tol_digits
+                )
+                hash_coord = HashableArray(hash_coord)
+                perm.append(self._hash_positions[hash_coord])
+
+            perms.append(tuple(perm))
+        return tuple(perms)
+
+    def rotation_perm(self, period, axes=[0, 1]):
```
PhilipVinc, line 343: Like above.
Type annotations for inputs/output. Docsrting. and if internal and not meant for public API prepend with an underscore.

netket/graph/lattice.py
```diff
@@ -246,6 +315,114 @@ def atoms_coord(self):
...
+        rot_coords[:, axes] = _np.matmul(rot_coords[:, axes], rot_mat)
+
+        for hash_coord in rot_coords:
+            hash_coord = _np.matmul(hash_coord, self._inv_dims) % 1
+            hash_coord = _np.around(hash_coord * 10 ** tol_digits).astype(int) % (
+                10 ** tol_digits
+            )
+            hash_coord = HashableArray(hash_coord)
+            if hash_coord in self._hash_positions:
+                perm.append(self._hash_positions[hash_coord])
```
PhilipVinc, line 361: You are using this hashing trick in a few places.
I think all this logic should go in it's own internal function (not method) , documented and used throughout.
So that if we decide to change the hasing logic it's easier to do as it's all in one place.
And it avoids duplciation.
and it can be tested stand-alone.

netket/graph/lattice.py
```diff
@@ -246,6 +315,114 @@ def atoms_coord(self):
...
+            else:
+                raise ValueError(
+                    "Rotation with the specified period and axes does not map lattice to itself"
+                )
+
+        return tuple(perm)
+
+    def reflection_perm(self, axis=0):
```
PhilipVinc, line 369: As above

netket/graph/lattice.py
```diff
@@ -246,6 +315,114 @@ def atoms_coord(self):
...
+        """
+
+        perm = self.rotation_perm(period, axes)
+        rotations = [PlanarRotation(perm, n) for n in range(1, period)]
+
+        return SymmGroup([Identity()] + rotations, graph=self)
+
+    def basis_translations(self) -> SymmGroup:
+        """
+        Returns SymmGroup corresponding to translations by basis vectors
```
PhilipVinc, line 405: It's not very important but add a dot at the end of the sentence?

netket/graph/lattice.py
```diff
@@ -246,6 +315,114 @@ def atoms_coord(self):
...
+        translations = [Translation(perms, i) for i in translations]
+
+        return SymmGroup([Identity()] + translations, graph=self)
+
+    def reflections(self, axis=0) -> SymmGroup:
+        """
+        Returns SymmGroup corresponding to reflection about axis
+        args:
+          axis: Generated reflections about specified axis
+        """
```
PhilipVinc, line 421: `axis: int = 0`

Dot at the end of the sentence.

Newline between the docstring and the arguments, otherwise the website breaks.
`args:` should be `Args:`, otherwise the website wont work

PhilipVinc (commented):
I'm sorry for arriving late to the party, but @chrisrothUT can you address the comments i left here in a followup PR?

There is nothing big about it except i would like the logic of the hashing factored out so we can change it later, and the `np.around` being imprecise so i'd like it changed or addressed.

Also, it would be great if you could add a test case where we have precision issues, so that if we refactor this code in the future we can have a test catching errors. 

-- 
I'm still convinced we are duplicating the implemnetation of those operations (Translation/Reflection...), because, while the ones of Grid won't work on lattice, the ones of lattice will work on grid because they are more general and work with permutations.
Am i right @femtobit ? (also, can you tell me why in the end you were ok with not subclassing `netket.utils.semigroup.Permutation`? I think the discussion diverged and i don't understand what was the answer in the end)

-- 
@attila-i-szabo Next time can you please wait a moment before merging? I mean, while we don't require everyone to approve a PR there were still point raised without answer in the discussion to keep things lean, this does not mean that it's ok to do it as soon as there is one green light.  
If a discussion involved quite a few people (and maybe, some issues are still not addressed), just asking and saying 'i'll merge in 24 hours if nobody has more issues is the usually polite way to approach this.

Also, in case you want to merge again, never _merge_. This is particularly true when working on master.
It makes the history impossible to read, and it makes it hard to find the PR correspeoning to a commit.  Also, if there are bugs we need to revert it's hard.
'Squash and Merge', so everything becomes 1 commit, making surethere is a reference to the PR number.

femtobit:
>  Am i right @femtobit ? (also, can you tell me why in the end you were ok with not subclassing netket.utils.semigroup.Permutation? I think the discussion diverged and i don't understand what was the answer in the end)

I only agreed with it because of @attila-i-szabo's plans to change all that again anyways based on the ideas in #703. Maybe a more appropriate way to go about this would be to merge the code here into a separate branch in the NetKet repo, so that we all could make PRs against that branch here without having it in master already. (While I approved, I did not expect this to be merged into master right afterwards, TBH. :sweat_smile:)

> I'm still convinced we are duplicating the implemnetation of those operations (Translation/Reflection...), because, while the ones of Grid won't work on lattice, the ones of lattice will work on grid because they are more general and work with permutations.

As long as `Grid` provides a way to get the coordinates of each node, that should be possible. So yes, we should probably go for that.

netket/graph/lattice.py
femtobit: I would also remove all the references to "hashing". That's an implementation detail. The important thing is that we have a mapping from (rounded) position to lattice site index.

attila-i-szabo:
> @attila-i-szabo Next time can you please wait a moment before merging? I mean, while we don't require everyone to approve a PR there were still point raised without answer in the discussion to keep things lean, this does not mean that it's ok to do it as soon as there is one green light.
> If a discussion involved quite a few people (and maybe, some issues are still not addressed), just asking and saying 'i'll merge in 24 hours if nobody has more issues is the usually polite way to approach this.

@PhilipVinc @femtobit sorry about that, I'm still learning collaborative github etiquette...
I suppose we can still unmerge it without breaking anything, since it only affects a single file.

> As long as `Grid` provides a way to get the coordinates of each node, that should be possible. So yes, we should probably go for that.

Both for this and a couple of @PhilipVinc's comments above that aim to polish this interface, I would say that it would make more sense to deal with this as we are developing the final API. For instance, I'm not sure `Grid` is providing much extra functionality compared to a smart `Lattice` constructor/subclass that is preloaded with an n-dimensional cubic lattice, so we could simplify stuff there.

In any case, I would like to draft this API in a PR, but it will take a bit of time. Is that OK with you guys?

PhilipVinc:
Don't worry. Everyone gets stuff wrong form time to time.
Just keep in mind for the next time. :-)

In general it's a bit a mess to unmerge things from master (because you have to edit history, and it can break stuff and open PRs) so it's something nobody really does.
The traditional way to 'unmerge' is to revert a commit (which creates a negating commit), but in this case it's hard to do because the commits were not squashed, so we have to revert N commits.

Anyhow, it's ok if it stays there as long as we can merge a PR to fix the issues i raised above.

My main reservation for merging was that if in the end we will change the implementation, there is little sense to merge it in master before it's final form. Kind of what happened with the other huge PR.

femtobit:
> Kind of what happened with the other huge PR.

I think for longer-running discussions we should move from having one big PR to having a slightly longer running development side-branch that people can make PRs on in return. Technically that's possible for most of our PRs already (as long as the source branch is part of the `netket/netket` repo), but we haven't really used this workflow except in very few cases.

netket/graph/lattice.py
chrisrothUT: This is fine, we can still with cutoff tol

netket/graph/lattice.py
chrisrothUT: Floating points regularly get the sign wrong. For example 1.0 - 1.0 is sometimes positive and sometimes negative. The old code broke any lattice where the atoms were specified to be on the boundary of the unit cell. 

I don’t get the point of this line tbh, the code checks for duplicate atoms somewhere else anyway.

netket/graph/lattice.py
chrisrothUT: > If I understand this correctly this is a poor's man fixed point decimal number implementaiton. Is that right?

Yeah that’s correct. 

> Or, if not, can you tell me that this won't introduce errors or bugs?

I don’t think it matters. It seems like the numpy bug is related to producing extra digits which we truncate away immediately. 

> Is this maybe linked to your 1 % 1 error?

As for the 1.0 % 1.0 bug it’s similar to the 1.0-1.0 bug, i.e. it produces a qualitatively different result based on which one is larger

netket/graph/lattice.py
chrisrothUT: This was a comment from a previous version that no longer makes sense.

chrisrothUT:
I strongly agree on this point. One of the issues we run into is that we all have different areas of expertise. So it seems pretty rare that only one person should make edits to the code when we make these API changes

chrisrothUT:
Anyway I’ll submit a small PR soon that fixes the documentation issues, and moves all of the hashing logic to nk.utils

attila-i-szabo:
> I strongly agree on this point. One of the issues we run into is that we all have different areas of expertise. So it seems pretty rare that only one person should make edits to the code when we make these API changes. So single pull requests to master don’t make a lot of sense

On the other hand, I think it would help here to start with one larger-scale redesign, which we then go on to refine, and that might be easier to start with one PR than trying to do it in many steps so that it works well at all intermediate points. So if you all like the outline in #703 (and your new PR and #709 are settled), I would implement it, and then we can improve that implementation.

netket/graph/lattice.py
attila-i-szabo: > Quoting from numpy documentation
> 
> ```python
> np.around uses a fast but sometimes inexact algorithm to round floating-point datatypes. 
> ```

It also says that it always returns `np.true_divide(np.rint(a * 10**decimals), 10**decimals)`. Since we always use `decimals=0`, our `around` is just `rint` which doesn't come with a similar caveat in the docs.

attila-i-szabo: We might want to write `rint` though for peace of mind and making a bit clearer what we're doing

PR #722, opened 2021-05-20
Test docstring examples as part of CI runs

femtobit:
This PR adds a step to out GitHub Actions CI runs which executes all example sections in our docstring comments in order to make sure they are up-to-date and working as expected (f221a30).

Unsurprisingly, this has unearthed a lot of failing examples (mostly because of using old API or incomplete imports) which are fixed here as well (10b1baa).

Note that if a docstring example really should not be tested, it can be skipped by adding
```
>>> import pytest; pytest.skip("reason")
```
as I've done here (only) for some `NETKET_EXPERIMENTAL` code.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -21,15 +21,23 @@ jobs:
 
           - os: ubuntu-latest
             python-version: 3.8
-            mpi: "mpich"
-
-          - os: macos-latest
-            python-version: 3.8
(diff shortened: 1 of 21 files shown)


gcarleo:
Oh very nice! Once we had these tests running but at some point they were removed, it is really good to have them back and fix the wrong examples...

gcarleo (approved):

PR #808, opened 2021-07-09
Gpescia/continuous support

gpescia:
Hi everyone,

I added the continuous Hilbert as discussed with signature `nk.hilbert.Particles(N=10, L=(3,3,3), pbc=(True,True,False),type="boson")`. 
Additionally there is an implementation of random_state():
For all directions where there are periodic boundary conditions I initialize the particle positions evenly between 0 and L and give them a Gaussian noise. For all directions where there are no boundaries, the positions are initialized with a Gaussian. 
Please comment on this if you have better ideas for initialization. 

I also added a rule for the MetropolisSampler. The rule takes the old positions and proposes a new position for each particle position according to a Gaussian distribution. If there are periodic boundary conditions I take the new positions modulus L (the extend in that spatial direction) and otherwise no modulus operation is done.


netket/hilbert/continuous_hilbert.py
```diff
@@ -0,0 +1,67 @@
...
+        return self._type
+
+    @property
+    def _attrs(self):
+        return (
+            self.size,
+            self.L,
+            self.pbc,
+            self.type,
+        )
```
PhilipVinc, line 67: newline missing

netket/hilbert/__init__.py
```diff
@@ -15,6 +15,8 @@
 from .abstract_hilbert import AbstractHilbert
 from .discrete_hilbert import DiscreteHilbert
 
+from .continuous_hilbert import Particles
```
PhilipVinc, line 18: I'd stop calling things in `/hilbert` with `_hilbert` because that's obvious (of course it is our bad).
I'd call this file `continuous_particles` or something like that instead of `continuous_hilbert` (we might have a continuoushilbert later on, and it's not particles...)

netket/hilbert/continuous_hilbert.py
```diff
@@ -0,0 +1,67 @@
+from typing import Tuple
+
+from .abstract_hilbert import AbstractHilbert
+
+class Particles(AbstractHilbert):
+    """Class for continuous Hilbert space.
```
PhilipVinc, line 6: This is not for continouus hilbert space.
It's rather for particles in a continous hilbert space.

again, mayube one day we'll have a proper base class for abstracthilberts (and particles ain't it).

PhilipVinc: Please change this class docstring to something more proper

netket/hilbert/continuous_hilbert.py
```diff
@@ -0,0 +1,67 @@
...
+
+    This class definese the common interface that can be used to
+    interact with hilbert spaces in continuum.
+    """
+
+    def __init__(self, N: int, L: Tuple[float, ...], pbc: Tuple[bool, ...], ptype: str):
+        """
+        Initializes a continuous Hilbert space.
```
PhilipVinc, line 14: .

netket/hilbert/continuous_hilbert.py
```diff
@@ -0,0 +1,67 @@
...
+
+        :param N: Number of particles
+        :param L: spatial extension in each spatial dimension
+        :param pbc: Whether or not to use periodic boundary conditions for each spatial dimension
+        :param type: Type of particles (bosonic,..)
```
PhilipVinc, line 19: We don't use numpy style but google style
```suggestion
    Args:
        N: Number of particles
        L: spatial extension in each spatial dimension
        pbc: Whether or not to use periodic boundary conditions for each spatial dimension
        type: Type of particles (bosonic,..)
```

PhilipVinc: Also, about the type of particles... @gcarleo what types do we expect? I think this is better to be a diffrent class (BosonicParticle, FermionicParticle...) than a string in here

netket/hilbert/continuous_hilbert.py
```diff
@@ -0,0 +1,67 @@
...
+            raise AssertionError(
+                """You need to define boundary conditions for each spatial dimension."""
+            )
+
+        super().__init__()
+
+    @property
+    def size(self) -> int:
+        r"""The number number of degrees of freedom in the
+                Hilbert space."""
```
PhilipVinc, line 36: you don't need the docstring in here because it's already defined in the base.
```suggestion
```

netket/hilbert/continuous_hilbert.py
```diff
@@ -0,0 +1,67 @@
...
+        return self._N * len(self._L)
+
+    @property
+    def N(self) -> int:
+        r"""The number of particles
+        """
```
PhilipVinc, line 42:
```suggestion
        r"""The number of particles"""
```

netket/hilbert/continuous_hilbert.py
PhilipVinc: This class (for now) it's more the hilbert space for N particles in a n-dimensional box.

netket/hilbert/random/continuous.py
```diff
@@ -0,0 +1,21 @@
...
+from jax import numpy as jnp
+
+from netket.hilbert import Particles
+from netket.utils.dispatch import dispatch
+
+
+@dispatch
+def random_state(hilb: Particles, key, batches: int, *, dtype):
+    """Positions particles w.r.t. normal distribution if no periodic boundary conditions are applied in a spatial dimension.
+    Otherwise the particles are positioned evenly along the box from 0 to L, with Gaussian noise."""
```
PhilipVinc, line 11: can you try to have lines not crazy long? 
We don't enforce the 89 character limit on docstrings, but it's better if we loosely try to abide by it when possible...
x

netket/hilbert/random/continuous.py
```diff
@@ -0,0 +1,21 @@
...
+    gaussian = jax.random.normal(key, shape=(batches, hilb.size))*0.2
+
+    minL = jnp.min(jnp.where(jnp.array(hilb.L) == None, jnp.inf, hilb.L))
+    uniform = jnp.tile(jnp.linspace(0., minL, hilb.size), (batches, 1))
+
+    boundary  = jnp.tile(jnp.array(hilb.N * hilb.L))
+
+    rs = jnp.where(boundary == None, gaussian, uniform + gaussian)
+
+    return jnp.asarray(rs, dtype=dtype)
```
PhilipVinc, line 21: newline at end of file
```suggestion
    return jnp.asarray(rs, dtype=dtype)

```

netket/hilbert/random/continuous.py
```diff
@@ -0,0 +1,21 @@
...
+
+from netket.hilbert import Particles
+from netket.utils.dispatch import dispatch
+
+
+@dispatch
+def random_state(hilb: Particles, key, batches: int, *, dtype):
+    """Positions particles w.r.t. normal distribution if no periodic boundary conditions are applied in a spatial dimension.
+    Otherwise the particles are positioned evenly along the box from 0 to L, with Gaussian noise."""
+    gaussian = jax.random.normal(key, shape=(batches, hilb.size))*0.2
```
PhilipVinc, line 12: what is this 0.2? can we do something more generic?

netket/sampler/metropolis.py
```diff
@@ -525,3 +525,18 @@ def MetropolisHamiltonian(hilbert, hamiltonian, *args, **kwargs) -> MetropolisSa
...
+       s: The width of the Gaussian proposal distribution
+       n_chains: The number of Markov Chain to be run in parallel on a single process.
+       sweep_size: The number of exchanges that compose a single sweep.
+                   If None, sweep_size is equal to the number of degrees of freedom (n_visible).
+    """
+
+    from .rules import GaussianRule
+
+    rule = GaussianRule(s)
+    return MetropolisSampler(hilbert, rule, *args, **kwargs)
```
PhilipVinc, line 542: For now I'd avoid having this and try to be explicit about the sampler/rule we are using.
This name is really too generic for what this is doing...
And in general we give the sampler the name of the rule (so in this case it would rather be `MetropolisGaussian`, which also ain't a great name)

netket/sampler/rules/__init__.py
```diff
@@ -15,6 +15,7 @@
 from .local import LocalRule
 from .exchange import ExchangeRule
 from .hamiltonian import HamiltonianRule
+from .gaussian_continuous import GaussianRule
```
PhilipVinc, line 18: can you rename it to `continouus_gaussian` so that they are packed together in alphabetical order?

netket/sampler/rules/gaussian_continuous.py
```diff
@@ -0,0 +1,31 @@
...
+        boundary = jnp.tile(jnp.array(hilb.N * hilb.L))
+        modulus = jnp.where(boundary == None, np.inf, boundary)
+
+        prop = jax.random.normal(key, shape=(n_chains, hilb.size)) * rule.s
+        σp = jnp.where(boundary == None, σ + prop, (σ + prop) % modulus)
+
+        return σp, None
+
+    def __repr__(self):
+        return "GaussianRule()"
```
PhilipVinc, line 31: missing newline.
Also, you could put the parameter.
```suggestion
        return f"GaussianRule(s={self.s})"

```

netket/sampler/rules/gaussian_continuous.py
```diff
@@ -0,0 +1,31 @@
...
+from ..metropolis import MetropolisRule
+
+@struct.dataclass
+class GaussianRule(MetropolisRule):
+    r"""
+    A transition rule acting on all particle positions at once.
+
+    New proposals of particle positions are generated according to a Gaussian distribution of width s.
+    """
+    s: float = 1.
```
PhilipVinc, line 16: by the way... `s` ain't a great choice for me.
Can't we have something like `sigma` or `variance`?

gcarleo:
Not sure we need a specific type for Bosons etc, I think strings are fine?

gcarleo:
there is en empty file Examples/HarmonicOsciallator/HO.py

netket/sampler/rules/__init__.py
gcarleo: I would maybe call this just gaussian, I guess that is intrinsically continuous

gcarleo: also if we want to be rigorous, ```GaussianRule``` should just go in ```.local``` I think!

PhilipVinc:
That’s just a file.
Let’s keep a rule per file, so that if we ever split this stuff in a sub package is easier, no?
Il 9 lug 2021, 18:03 +0200, Giuseppe Carleo ***@***.***>, ha scritto:

netket/hilbert/random/continuous.py
gpescia: it is the noise level but since the signature is random_state(self, key, size=None, dtype=np.float32)
there is no place for another parameter to pass.

gpescia:
> there is an empty file Examples/HarmonicOsciallator/HO.py

gpescia:
> there is an empty file Examples/HarmonicOsciallator/HO.py

I added something to the file now.

gpescia:
> Not sure we need a specific type for Bosons etc, I think strings are fine?

I left it as it was now. Filippo suggested to have an abstract continuous_space_hilbert and derive from it bosons and fermions (although I think that the Hilbertspace is the same for both actually) However dispatching gets easier the way Filippo suggests (and that is why he suggested it)

gpescia:
I implemented Filippos suggestions, let me know what you think.

gcarleo:
thanks Gabriel, almost there... I think we need to add at least a few tests, for example check that sampling a gaussian with Metropolis gives what expected etc

netket/hilbert/__init__.py
```diff
@@ -15,6 +15,8 @@
 from .abstract_hilbert import AbstractHilbert
 from .discrete_hilbert import DiscreteHilbert
 
+from .continuous_space_particles import Particles
```
gcarleo, line 18: to be consistent with the other Hilbert spaces, I think this should be called ```Particle``` (singular) and not ```Particles``` (sorry for the previous suggestion, I guess I had suggested the plural before) 

In this way it is actually more natural to do things like 

```python
particles=(Particle(L=(10,10,10))*Spin(1/2))**N 
```
which is the typical use case, with explicit tensor powers, we should push for more and more in the future. 

Of course the argument N can be left/ should be left both for spin and particle, but that's just a convenient short cut, I would say...

gcarleo:
Also, to be consistent with the other Hilbert, I think that the signature for ```Particle```should be more something like

```python
@dataclass
class Particle:
    L: float 
    pbc: Tuple[bool,...] = [True,] 
    N: int = 1
    type: Optional[str] = None
```

what do you think?

netket/sampler/rules/continuous_gaussian.py
```diff
@@ -0,0 +1,31 @@
...
+@struct.dataclass
+class GaussianRule(MetropolisRule):
+    r"""
+    A transition rule acting on all particle positions at once.
+
+    New proposals of particle positions are generated according to a Gaussian distribution of width s.
+    """
+    s: float = 1.
+
+    def transition(rule, sampler, machine, parameters, state, key, σ, sigma):
```
gcarleo, line 18: can we call sigma -> r , it would be more transparent

netket/sampler/rules/continuous_gaussian.py
```diff
@@ -0,0 +1,31 @@
...
+
+        n_chains = σ.shape[0]
+        hilb = sampler.hilbert
+        boundary = jnp.tile(jnp.array(hilb.N * hilb.L))
+        modulus = jnp.where(boundary == None, jnp.inf, boundary)
+
+        prop = jax.random.normal(key, shape=(n_chains, hilb.size)) * rule.sigma
+        σp = jnp.where(boundary == None, σ + prop, (σ + prop) % modulus)
```
gcarleo, line 26: sp -> rp

netket/sampler/rules/continuous_gaussian.py
```diff
@@ -0,0 +1,31 @@
...
+from ..metropolis import MetropolisRule
+
+@struct.dataclass
+class GaussianRule(MetropolisRule):
+    r"""
+    A transition rule acting on all particle positions at once.
+
+    New proposals of particle positions are generated according to a Gaussian distribution of width s.
+    """
+    s: float = 1.
```
gcarleo, line 16: is this s or sigma ? (as used below)

netket/hilbert/continuous_space_particles.py
```diff
@@ -0,0 +1,69 @@
...
+        r"""Spatial extension in each spatial dimension"""
+        return self._L
+
+    @property
+    def pbc(self) -> Tuple[bool, ...]:
+        r"""Whether or not to use periodic boundary conditions for each spatial dimension"""
+        return self._pbc
+
+    @property
+    def ptype(self):
```
gcarleo, line 58: shouldn't this be just ```type``` ?

netket/hilbert/continuous_space_particles.py
```diff
@@ -0,0 +1,69 @@
+from typing import Tuple
+
+from .abstract_hilbert import AbstractHilbert
+
```
gcarleo, line 4: I think it is worth turning this into a dataclass?

netket/hilbert/continuous_space_particles.py
```diff
@@ -0,0 +1,69 @@
...
+class Particles(AbstractHilbert):
+    """Class for the Hilbert space of particles
+    in continuous space.
+    This class defines the common interface that
+    can be used to interact with particles defined
+    in continuous space.
+    """
+
+    def __init__(self, N: int, L: Tuple[float, ...], pbc: Tuple[bool, ...], ptype: str):
```
gcarleo, line 13: Possible signatures that we should consider : 
```python
Particle(N: int =1, L: Optional[Tuple[float,...]] = None, pbc: Optional[Tuple[bool, ...]=None, type: Optional[str]=None)
```

```python
Particle(N: int = 1, Geometry: Optional[GeoType]=None, type: Optional[str]=None) 
```

personally I prefer the second, since I don't like putting too many details on the simulation box in the particle itself, since that's rather a geometry concept?

netket/hilbert/random/continuous.py
PhilipVinc: Yeah, but I mean: why did you pick 0.2? Is it possible to chose a value that depends on the Hilbert space (say, for example, on box size times the number of particles, or anything)

And, if not, at least document it that this 0.2 was chosen completely arbitrarily.

netket/sampler/rules/__init__.py
PhilipVinc: `local.py` is just a file. Those are not seen by the user.
I'd like to keep one Transition rule per file for now.
It will make it easier to split netket in sub packages if it ever happens in the future...

netket/hilbert/continuous_space_particles.py
PhilipVinc: Hilbert spaces are not dataclasses but they are dataclass-compatible if you properly define `_attr`.
`_attr` should return a tuple of all fields that must be used to compute the hash and equality among hillbert spaces of the same type.


You can make it a dataclass but 1) it would be useless (for now), because hilbert is always passed around as a `static_argnum` object, and 2) you should be extra careful not to break comparison between hilbert spaces and hashability.

netket/hilbert/continuous_space_particles.py
PhilipVinc: I like the second better, but maybe we could have `L=Tuple[float,...]` be a shortcut to construct the corresponding `L1xL2xL3..` geometry?
I'd refrain from defining the `GeoType` interface now, that we have no use for it (yet) and leave this for the future...

PhilipVinc: Also, I dislike `ptype` as name for particles and would try for something better.

netket/hilbert/continuous_space_particles.py
PhilipVinc: I guess this stands for particle type (boson/fermion...).
I don't like `ptype` or `type`, to be honest.
They are not very transparent...

That's also why i'd like to keep this information in the type domain..

PhilipVinc:
Gabriel, please check step 3 in https://www.netket.org/docs/contributing.html for how to setup your dev environment.
(mainly you need to set up `pre-commit`) so that code is automatically formatted and it never fails the check

netket/hilbert/continuous_space_particles.py
femtobit: For consistency with the common lattices, maybe a better name for `L` would be `extent`? (`N` is fine as that's used everywhere consistently.)

gpescia:
I commited some new stuff: 
1) As requested by Filippo now there is an abstract particle class + a boson and fermion class derived from this particle class. 
2) I added tests for Hilbert and Sampler.
Also I think I should have captured all your suggestions except for the new signature of particle. I think it would be nice if we decide on one layout.

PhilipVinc:
@gpescia can you click on `resolve issues` on the issues that you think are addressed? it makes revieweing for me a bit easier

netket/hilbert/__init__.py
```diff
@@ -14,13 +14,16 @@
 
 from .abstract_hilbert import AbstractHilbert
 from .discrete_hilbert import DiscreteHilbert
+from .abstract_continuous_space_particle import AbstractParticle
```
PhilipVinc, line 17: sorry to be pedantic, but lets call this file `continuous_space_abstract_particle`

netket/hilbert/abstract_continuous_space_particle.py
```diff
@@ -0,0 +1,78 @@
...
+            N: Number of particles
+            L: spatial extension in each spatial dimension
+            pbc: Whether or not to use periodic boundary
+                conditions for each spatial dimension
+        """
+        self._N = N
+        self._L = L
+        self._pbc = pbc
+
+        if not len(self._L) == len(self._pbc):
```
PhilipVinc, line 31: Can you add, for convenience something like 
```python
    if isinstance(pbc, bool):
        pbc = [pbc] * ndim
    elif not len(self._L) == len(self._pbc):
       raise..

File: netket/hilbert/abstract_continuous_space_particle.py [hunk 24]
```diff
@@ -0,0 +1,78 @@
...
+            raise AssertionError(
+                """You need to define boundary conditions for each spatial dimension."""
```
PhilipVinc, line 33: AThat's a ValueError. And with the change above we can rephrase the error string
```suggestion
            raise ValueError(
                """`pbc` must be either a bool or a tuple indicating the periodicity of each spatial dimension."""
```

netket/hilbert/abstract_continuous_space_particle.py
```diff
@@ -0,0 +1,78 @@
...
+        super().__init__()
+
+    @property
+    def size(self) -> int:
+        return self._N * len(self._L)
+
+    @property
+    def N(self) -> int:
+        r"""The number of particles"""
+        return self._N
```
PhilipVinc, line 45: It is customary for attributes to be lowercase, with a few exceptions.
But it's better to stick to the customs.

Moreover, i'd go for a more explicit name like `n_particles`

netket/hilbert/abstract_continuous_space_particle.py
```diff
@@ -0,0 +1,78 @@
...
+
+    @property
+    def L(self) -> Tuple[float, ...]:
+        r"""Spatial extension in each spatial dimension"""
+        return self._L
```
PhilipVinc, line 50: Idem con patate.
I'd call this `extent`

netket/hilbert/abstract_continuous_space_particle.py
```diff
@@ -0,0 +1,78 @@
...
+        return jnp.array(self.N * self.pbc)
+
+    def L_to_array(self):
+        r"""Returns an array of length self.size
+        with the spatial extension in each
+        spatial dimension. If no periodic boundaries
+        are defined in a spatial dimension
+        returns np.inf"""
+        Ls = jnp.array(self.N * self.L)
+        return jnp.where(jnp.equal(self.pbc_to_array(), False), jnp.inf, Ls)
```
PhilipVinc, line 78: what is this stuff needed for? it's a bit weird to have this things live in hilbert

netket/hilbert/__init__.py
gpescia: In view that the boson/fermion file is called continuous_space_boson/fermion doesn't it make more sense to have the abstract before continuous_space?

netket/sampler/metropolis.py
```diff
@@ -525,3 +525,20 @@ def MetropolisHamiltonian(hilbert, hamiltonian, *args, **kwargs) -> MetropolisSa
...
+    and proposes a new state according to a Gaussian distribution
+    with width sigma.
+    Args:
+       hilber: The continuous Hilbert space
+       sigma: The width of the Gaussian proposal distribution
+       n_chains: The number of Markov Chain to be run in parallel on a single process.
+       sweep_size: The number of exchanges that compose a single sweep.
+                   If None, sweep_size is equal to the number of degrees of freedom (n_visible).
+    """
+
```
PhilipVinc, line 540: maybe check that the hilbert space `isinstance(hlibert, AbstractParticle)` and throw an error otherwise?

PhilipVinc: `ValueError("This sampler only works for Continuous Hilbert spaces, my dear and beloved user")`

netket/sampler/metropolis.py
```diff
@@ -525,3 +525,20 @@ def MetropolisHamiltonian(hilbert, hamiltonian, *args, **kwargs) -> MetropolisSa
 
     rule = HamiltonianRule(hamiltonian)
     return MetropolisSampler(hilbert, rule, *args, **kwargs)
+
+def MetropolisGaussian(hilbert, sigma, *args, **kwargs) -> MetropolisSampler:
```
PhilipVinc, line 529:
```suggestion
def MetropolisGaussian(hilbert, sigma=1.0, *args, **kwargs) -> MetropolisSampler:
```

netket/sampler/metropolis.py
```diff
@@ -525,3 +525,20 @@ def MetropolisHamiltonian(hilbert, hamiltonian, *args, **kwargs) -> MetropolisSa
...
+    r"""This sampler acts on all particle positions simultaneously
+    and proposes a new state according to a Gaussian distribution
+    with width sigma.
+    Args:
+       hilber: The continuous Hilbert space
+       sigma: The width of the Gaussian proposal distribution
```
PhilipVinc, line 535:
```suggestion
       sigma: The width of the Gaussian proposal distribution (default: 1.0)
```

test/hilbert/test_hilbert.py
```diff
@@ -178,6 +208,9 @@ def test_flip_state(hi):
 
         np.testing.assert_allclose(states_np, states_new_np)
 
+    elif isinstance(hi, ContinuousBoson):
+        return
+
```
PhilipVinc, line 213: If you don't test nothing then don't...
```suggestion
```

PhilipVinc (commented):
almost there!

netket/hilbert/__init__.py
PhilipVinc: if `DiscreteHilbert` was called `abstract_hilbert` yes.
maybe i have to rename that one.

PhilipVinc: well ok then

netket/hilbert/abstract_continuous_space_particle.py
gpescia: i need it in the random_state as well as for the transition rule of the sampler...

PhilipVinc: can't you just put the code there to do it?
it's a two liner... and it pollutes a bit the interface.

PhilipVinc:
ps: you are still failing formatting checks so you probably need to run `black .` once

test/hilbert/test_hilbert.py
```diff
@@ -203,11 +236,18 @@ def test_random_states_legacy(hi):
         assert hi.random_state(size=(10,)).shape == (10, hi.size)
         assert hi.random_state(size=(10, 2)).shape == (10, 2, hi.size)
 
+    elif isinstance(hi, ContinuousBoson):
+        return
```
PhilipVinc, line 240: uselss
```suggestion
```

test/hilbert/test_hilbert.py
```diff
@@ -203,11 +236,18 @@ def test_random_states_legacy(hi):
...
+
 
 @pytest.mark.parametrize(
     "hi", [pytest.param(hi, id=name) for name, hi in hilberts.items()]
 )
 def test_hilbert_index(hi):
+
+    if isinstance(hi, ContinuousBoson):
+        return
+
```
PhilipVinc, line 250:
```suggestion
@pytest.mark.parametrize(
    "hi", [pytest.param(hi, id=name) for name, hi in hilberts.items() if isinstance(hi, DiscreteHilbert)]
)
def test_hilbert_index(hi):
```

gpescia:
I added the suggested changes.
(Hopefully the formatting works now o.o)

netket/models/gaussian.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+class Gaussian(nn.Module):
+    r"""Multivariate Gaussain function with mean 0
+    and covariance matrix :math:`\Sigma_{ij}`:
+    :math:`\Psi(x) = \exp(\sum_{ij} x_i \Sigma_{ij} x_j)`.
+    The :math:`\Sigma_{ij} = AA^T` matrix is stored as
+    non-positive definite matrix A.
+    """
+
```
PhilipVinc, line 15: Its good customs to have docstring have a first short line or two and then the rest.
This is used as a preview in the doc system,

```suggestion
    r"""
    Multivariate Gaussain function with mean 0 and parametrised covariance matrix
    :math:`\Sigma_{ij}`.
    
    The wavefunction is given by the formula: :math:`\Psi(x) = \exp(\sum_{ij} x_i \Sigma_{ij} x_j)`.
    The (positive definite) :math:`\Sigma_{ij} = AA^T` matrix is stored as
    non-positive definite matrix A.
    """

```

netket/sampler/metropolis.py
```diff
@@ -525,3 +527,25 @@ def MetropolisHamiltonian(hilbert, hamiltonian, *args, **kwargs) -> MetropolisSa
...
+    Args:
+       hilber: The continuous Hilbert space
+       sigma: The width of the Gaussian proposal distribution (default: 1.0)
+       n_chains: The number of Markov Chain to be run in parallel on a single process.
+       sweep_size: The number of exchanges that compose a single sweep.
+                   If None, sweep_size is equal to the number of degrees of freedom (n_visible).
+    """
+    if not isinstance(hilbert, AbstractParticle):
+        raise ValueError(
+            "This sampler only works for Continuous Hilbert spaces, try another one."
```
PhilipVinc, line 545:
```suggestion
            "This sampler only works for Continuous Hilbert spaces."
```

netket/hilbert/random/particle.py
```diff
@@ -0,0 +1,34 @@
...
+    modulus = jnp.where(jnp.equal(pbc, False), jnp.inf, Ls)
+
+    gaussian = jax.random.normal(key, shape=(batches, hilb.size))
+    width = jnp.min(modulus) / (4.0 * hilb.n_particles)
+    """The width gives the noise level. In the periodic case the
+    particles are evenly distributed between 0 and min(L). The
+    distance between the particles coordinates is therefore given by
+    min(L) / hilb.N. To avoid particles to have coincident
+    positions the noise level should be smaller than half this distance.
+    We choose width = min(L) / (4*hilb.N)"""
```
PhilipVinc, line 28: stuff with `"""` is a docstring.
This is more like a comment...
```suggestion
    width = jnp.min(modulus) / (4.0 * hilb.n_particles)
    # The width gives the noise level. In the periodic case the
    # particles are evenly distributed between 0 and min(L). The
    # distance between the particles coordinates is therefore given by
    # min(L) / hilb.N. To avoid particles to have coincident
    # positions the noise level should be smaller than half this distance.
    # We choose width = min(L) / (4*hilb.N)
```

PhilipVinc (approved):

PR #851, opened 2021-08-06
Fixing typo in serialisation documentation

VolodyaCO:
I was checking the new way of saving and loading models (variational states), and found a typo in the documentation that throws an error when running because the correct writing mode is "wb" and not "w".


gcarleo (approved):
Thank you! I also spotted this some time ago but forgot to report it... thanks for fixing it!

PR #911, opened 2021-09-07
Tutorial on Matrix Models.

erinaldi:
This tutorial follows the netket 3 introduction tutorial but it is based on a different system, useful in High Energy Physics.

I have discussed the possibility of adding this tutorial with @gcarleo and @PhilipVinc 

I welcome feedback for improvements.

- [ ] the pre-commit on the notebooks is a bit controversial. I run `black` on all my code cells.


gcarleo:
Thanks for this nice contribution! 
Just a quick question: here I see 

Optimized energy: 3.150e+00 ± 9.791e-17 [σ²=9.664e-30, R̂=1.0056]  
Relative error: 0.51% 

but it looks like the energy variance is exactly zero, so you landed (exactly) on an excited state?

erinaldi:
> Optimized energy: 3.150e+00 ± 9.791e-17 [σ²=9.664e-30, R̂=1.0056]
> Relative error: 0.51%
> 
> but it looks like the energy variance is exactly zero, so you landed (exactly) on an excited state?

Hi Giuseppe,
Indeed I just talked to Filippo about this. I had not noticed the zero variance and the problems related to it in the VMC setup.
When I do the full exact diagonalization, I know that there is a large gap between the ground state and the first excited state (3.13 -> 4.21).
The energy 3.15 is not something I would expect from an excited state, but maybe it can be a superposition of the ground state and some excited state.
I have note used the symmetry of the model to define my ansatz: this is something I want to try next.

gcarleo:
yes I see, so since you did not land on an exact excited state there might be something else wrong going on, maybe with the sampling!

erinaldi:
@gcarleo I am not familiar with diagnosing this type of VMC systems, so I started from looking at the properties of the MCMC sampling, which I am more comfortable with.
I noticed some of them have `nan` autocorrelation time, which would indicate (I believe) that the sampling gave the same value repeatedly along the chain.
With the parameters of the optimizers that are currently in that cell, the appearance of zero variance chains starts around iteration 170.
However, there is also another "plateaux" in the energy before that, at energy 6.25 (not an excited state energy, but close to one, 6.19) which is rather pronounced (long) if I decrease the learning rate and also displays  the zero variance phenomenon.

I would appreciate so directions in what to use in terms of diagnostic tools. 🙏🏼

gcarleo:
Most likely the acceptance is going to zero hero, you will need moves that are not just local single Boson move otherwise I guess as soon as you start getting the gauge constraint right the sampling will not be ergodic

PhilipVinc:
try to decrease the SGD learning rate.

erinaldi:
@gcarleo I read the documentation a bit more carefully and found out how to look at the samples.
On each site there is only 1 value that is sampled (as suspected from `TauCorr`) 
Screen Shot 2021-09-08 at 19 52 14
I think your suggestion of a different sampler is good. However, how can I check the acceptance? (a search on the website returned nothing...)

@PhilipVinc this result is with a 10x smaller learning rate. You mentioned before the option to visualize the state in the computational basis using `to_array()`. Indeed in this case the state is just 1 single value of 0.999 followed by 1e-5 numbers.
Screen Shot 2021-09-08 at 19 58 32

gpescia:
> However, how can I check the acceptance?

Try to add this callback in the run-method:

`def mycb(step, logged_data, driver):
        logged_data["acceptance"] = float(driver.state.sampler_state.acceptance)
        return True`

erinaldi:
@gpescia Thanks for the suggestion.
When the energy _gets stuck_ in a _plateaus_ during the optimization process (there is an intermediate plateau before reaching the smallest energy when using a fully-connected ansatz), the acceptance rate goes to zero.
Screen Shot 2021-09-10 at 12 36 47

erinaldi:
and I should add the plot of the energy corresponding to the plot of the acceptance rate above.
Screen Shot 2021-09-10 at 12 42 51

the result of `FFN` is from running:
```python
optimizer = nk.optimizer.Sgd(learning_rate=0.01)
preconditioner = nk.optimizer.SR(diag_shift=0.1)

# Notice the use, again of Stochastic Reconfiguration, which considerably improves the optimisation
gs = nk.driver.VMC(
    H, optimizer, variational_state=vstate, preconditioner=preconditioner
)

# logging and running
log = nk.logging.RuntimeLog()
gs.run(n_iter=3000, out=log, callback=acc_cb)
```
(and then plotting the iterations on the same scale used by the Jastrow ansatz. Also the Jastrow ansatz gets stuck, albeit minimally if the iterations are kept low: but if the energy is close to 3.15, the σ blows up.

gcarleo:
yes if the acceptance goes to zero it means that essentially all the results you have after that are wrong (unless the quantum state is a purely classical state with overlap only on one basis ket). Maybe I would suggest to try ExactSampler (if the system is small enough) to see what happens with a sampler that does not have ergodicity issues? If that works then one can think of a MCMC sampler that scales better but more ergodic than MetropolisLocal in this case

erinaldi:
> yes if the acceptance goes to zero it means that essentially all the results you have after that are wrong (unless the quantum state is a purely classical state with overlap only on one basis ket). Maybe I would suggest to try ExactSampler (if the system is small enough) to see what happens with a sampler that does not have ergodicity issues? If that works then one can think of a MCMC sampler that scales better but more ergodic than MetropolisLocal in this case

@gcarleo from the exact diagonalization procedure I can see that the ground state is indeed very similar to what comes out of the variational ansatz: both have a single component with >99% overlap on a single basis ket.
Here I am plotting only the first 10 basis kets:
Screen Shot 2021-09-10 at 14 59 44

erinaldi:
@gcarleo and @PhilipVinc I have re-run the entire tutorial using the `ExactSampler` instead of the `MetropolisLocal` sampler and the results do not change much.
With the `ExactSampler` the convergence to the ground state energy for the fully-connected NN ansatz is much faster.
There is no issue with the acceptance now, since the samples are exact.
The ground state is almost 100% one single basis ket so the sampling will favor that one and this is just how this system is defined.

One way to improve is to implement symmetries, but I haven't done it yet.

Is there anything else you would suggest?

@gpescia The benchmark I have for this matrix model is done using continuous variables. I was wondering if I could use the new `ContinuousBoson` but there is very limited documentation related to that. In our paper we use an autoregressive flow to model the wave function and the basis coordinates are just real numbers in a 6-dim vector. Since we use a generative model for the wf, our sampling is also exact. With that setup we get perfect agreement with the exact diagonalization results.

erinaldi:
@PhilipVinc sorry for the long silence on this.
Do you have a suggestion on how to improve this tutorial?
Would it be worth adding it to the tutorials of NetKet, or should I close the PR?

docs/index.md
```diff
@@ -56,6 +56,8 @@ tutorials/gs-ising
 tutorials/gs-heinsenberg
 tutorials/gs-j1j2
 tutorials/gs-gcnn-honeycomb
+tutorials/gs-free-matrix-models
+tutorials/gs-matrix-models
```
PhilipVinc, line 60:
```suggestion
tutorials/gs-matrix-models
tutorials/gs-free-matrix-models
```

PhilipVinc, line 60:
```suggestion
tutorials/gs-matrix-models
tutorials/gs-gcnn-honeycomb
```

PhilipVinc (approved):
I really love this and look forward to you coming back to Lausanne to harass me into approving your pull requests.

fabienalet:
GccfccvgcvtcEnvoyé depuis mon smartphone Samsung Galaxy.
-------- Message d'origine --------De : Filippo Vicentini ***@***.***> Date : 15/07/2022  17:54  (GMT+01:00) À : netket/netket ***@***.***> Cc : Subscribed ***@***.***> Objet : Re: [netket/netket] Tutorial on Matrix Models. (#911) 

@PhilipVinc commented on this pull request.







In docs/index.md:

>  tutorials/gs-gcnn-honeycomb

+tutorials/gs-free-matrix-models

+tutorials/gs-matrix-models



⬇️ Suggested change

-tutorials/gs-gcnn-honeycomb

-tutorials/gs-free-matrix-models

-tutorials/gs-matrix-models

+tutorials/gs-matrix-models

+tutorials/gs-gcnn-honeycomb







—Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you are subscribed to this thread.Message ID: ***@***.***>

[

{

***@***.***": "http://schema.org",

***@***.***": "EmailMessage",

"potentialAction": {

***@***.***": "ViewAction",

"target": "https://github.com/netket/netket/pull/911#pullrequestreview-1040492698",

"url": "https://github.com/netket/netket/pull/911#pullrequestreview-1040492698",

"name": "View Pull Request"

},

"description": "View this Pull Request on GitHub",

"publisher": {

***@***.***": "Organization",

"name": "GitHub",

"url": "https://github.com"

}

}

]

PR #912, opened 2021-09-07
QGT with chunks

inailuig:


netket/jax/_batch_utils.py
```diff
@@ -0,0 +1,47 @@
+from netket.jax import compose
+
+
+def unbatch(x):
+    return x.reshape((-1,) + x.shape[2:])
+
+
+def batch(x, batchsize=None):
```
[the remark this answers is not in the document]
PhilipVinc: very good idea

PhilipVinc:
Let's just wait to see if jax people accept https://github.com/google/jax/pull/7833 .
I'll review after they tell us if it is ok.

By the way, you could also ask them if they could add `linear_call` to the official stable API.
Otherwise they might break this code in the future.

attila-i-szabo:
Two observations
* The batched implementation duplicates all linearisation work. It would be better to perform something like the unbatched `matvec` on every batch and sum the outputs (kinda like the backward pass of #830). I suppose aggregation other than `stack` would be useful for a general-purpose batching interface anyway
* `O_jvp`/`O_vjp`/etc. should probably be jitted as we call it many times in a for loop

inailuig:
> Two observations
> 
> * The batched implementation duplicates all linearisation work. It would be better to perform something like the unbatched `matvec` on every batch and sum the outputs (kinda like the backward pass of [[WIP] minibatches for expect (forward pass) #830](https://github.com/netket/netket/pull/830)). I suppose aggregation other than `stack` would be useful for a general-purpose batching interface anyway

The point is that this is aimed at the setting where storing the linearisation work becomes infeasible because of the memory consumption.

I don't think summing smaller matvec's would be equivalent, since the mean of the Ok's (for computing \Delta Ok) is taken over all the samples. I guess you could precompute said mean (as was done in some previous version of this code) by doing an additional vjp, but I don't see the advantage of doing that?


> * `O_jvp`/`O_vjp`/etc. should probably be jitted as we call it many times in a for loop

The whole matvec and solve functions are jitted...

attila-i-szabo:
> I don't think summing smaller matvec's would be equivalent, since the mean of the Ok's (for computing \Delta Ok) is taken over all the samples. I guess you could precompute said mean (as was done in some previous version of this code) by doing an additional vjp, but I don't see the advantage of doing that?

Good catch, I forgot about that step. I don't think there's a good way around that...

> > * `O_jvp`/`O_vjp`/etc. should probably be jitted as we call it many times in a for loop
> 
> The whole matvec and solve functions are jitted...

I'd still benchmark this. Jitting `matvec` would work by unrolling `for` loops IIRC; depending on the internal workings of `jit`, we might save on not jitting `O_vjp` that many times.

inailuig:
> > The whole matvec and solve functions are jitted...
> 
> I'd still benchmark this. Jitting `matvec` would work by unrolling `for` loops IIRC; depending on the internal workings of `jit`, we might save on not jitting `O_vjp` that many times.
> EDIT I mean we'd jit whatever is inside the for loop separately

The loop body should be compiled only once (actually with the current implementation worst case 2x for the O_vjp since the first batch is computed outside of the loop, but that could easily be fixed by initialising the carry with the correct pytree of zeros...) 
The jax scan's are lowered to while loops (`jax.lax.scan` with `unroll=1` calls this: https://github.com/google/jax/blob/d6a4b2f1b58a7b6141be029351b139a7529a7959/jax/_src/lax/control_flow.py#L1359).
I would assume XLA doesn't try to unroll (and interleave) the loop body( for the vjp there is also a dependency on the carry).

If you mean jitting the loop bodies (with jax.jit(..., inline=False) which is their default) then we can certainly try that.

PhilipVinc:
@inailuig where is the first compilation happening?

Maybe @attila-i-szabo is referring to my old implementation, where I was jax.jitting  only the loop body and looping from Python, so that increasing the number of minibatches would not incur into reconciliation.

however, while a nice idea, it complicates remarkably the implementation and I’d avoid it.

Eventually  jax will add support for ‘arbitrary’ dimensions that don’t trigger recompilation (there’s already something there) so by performing the reshape in Python we might get away with decreasing recompilayopn..

netket/jax/_scanmap.py
```diff
@@ -0,0 +1,39 @@
+import jax
+
+from jax import linear_util as lu
+from jax.api_util import argnums_partial
+
+# from functools import wraps
+
+
+def scan_accum(f, x, op):
+    res0 = f(jax.tree_map(lambda x: x[0], x))
```
inailuig, line 10: @PhilipVinc here f is called once, and then below in the loop. 
```suggestion
    res_init = jax.tree_map(lambda x: jnp.zeros(x.shape), jax.eval_shape(f, jax.tree_map(lambda x: x[0], x)))
```

netket/jax/_scanmap.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def f_(carry, x):
+        return op(carry, f(x)), None
+
+    res, _ = jax.lax.scan(
+        f_, res0, jax.tree_map(lambda x: x[1:], x), unroll=1
```
inailuig, line 16:
```suggestion
        f_, res_init, x, unroll=1
```

netket/jax/_scanmap.py
PhilipVinc: yeah let's try to avoid increasing compilation time, and do everything inside the loop.

PhilipVinc:
@inailuig i'd prefer to do everything in the loop and zero-initialise it independently

netket/jax/_scanmap.py
```diff
@@ -0,0 +1,39 @@
+import jax
```
inailuig, line 1:
```suggestion
import jax
import jax.numpy as jnp
```

PhilipVinc:
check formattin @inailuig

netket/jax/_scanmap.py
inailuig: Thinking about it, the reason I did it this way in the first place was that for a general op the zero element is not necessarily the number 0. 
Computing the first batch first avoids having to specify what the 0 element would be.

netket/optimizer/qgt/qgt_onthefly_logic.py
```diff
@@ -14,10 +14,29 @@
...
+
+# temporary fix to make linear_call jittable
+# TODO remove once in jax
+from jax.interpreters import xla
+from jax._src.custom_derivatives import linear_call_p, _linear_call_impl
+
+xla.initial_style_translations[linear_call_p] = xla.lower_fun_initial_style(
+    _linear_call_impl
+)
+
```
PhilipVinc, line 39: Jax just made a new release which includes your PR, so I'd remove this and you can bump the minimum jax version to 0.2.21

```suggestion
```

netket/optimizer/qgt/qgt_onthefly.py
```diff
@@ -178,7 +186,7 @@ def _to_dense(self: QGTOnTheFlyT) -> jnp.ndarray:
     """
     Npars = nkjax.tree_size(self._params)
     I = jax.numpy.eye(Npars)
-    out = jax.vmap(lambda x: self @ x, in_axes=0)(I)
+    _, out = jax.lax.scan(lambda _, x: (None, self @ x), None, I)
```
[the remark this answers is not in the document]
inailuig: The problem is that the scan in vmap_chunked does not have a jax batching rule (i.e. the thing telling jax how to push down the vmap). 
When we are using the batched stuff to reduce memory consumption we would want to do a scan like this anyway. I think with some effort it might be possible to write a jax batching rule that achieves that, keeping the vmap for the implementation without chunks.

It was my impression that if people want the matrix they can use dense or pytree, but maybe the usecase for this is for many samples and few param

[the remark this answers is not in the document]
PhilipVinc: can we at least have a switch? `if self.chunking: scan else vmap`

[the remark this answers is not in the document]
inailuig: actually it's the linear call which does not have the rule, but I'll just add your workaround for the time being

PhilipVinc (approved):

PhilipVinc:
thanks! 
With this the chunking story is finally finished.

I'll merge when tests pass

PR #1045, opened 2021-12-20
Fix GCNN initializers

attila-i-szabo:
Currently, the initializers used for GCNNs are a bit of a mess. We define several different initializers (some of which are manifestly incorrect, e.g. a `jax.nn.lecun_normal` that refers to the wrong axes), all of which are trumped in the `GCNN` classes by a custom variance-scaled normal distribution, ~which is however a factor of sqrt(2) too large for complex numbers.~ this only works because `jax.random.normal` is now defined to halve the variance of real/imaginary parts for complex dtypes.

This PR clears this up. 
* The one remaining initialiser in `nk.nn` is dropped for it is internal and now available in all versions of JAX that we support. This leaves only deprecations in the `initializers.py` file.
* A single `default_gcnn_initializer` is defined in the same file as `DenseSymm` and `DenseEquivariant`, which is also exported into `GCNN`: this is the plain JAX `lecun_normal` instantiated with the correct axes, the one we ought to use with SELU.

Variance scaling for masks is also fixed by defining a `scaled_mask` whose norm is always the same as a mask full of 1s.

Possible TODOs:
* Check the initialiser of `RBMSymm`. I have no intuition as to whether `lecun_normal` is appropriate for that too, so please weigh in; we can probably do better than the current `normal(0.1)`
* More generally, define similar initialisers for non-GCNN layers like `Dense`


PhilipVinc:
is this equivalent to the old?

attila-i-szabo:
No, the point is that the initializers are changed to do variance scaling correctly

PhilipVinc:
Last time I tried to do that in https://github.com/netket/netket/pull/863 I found out after a little benchmarking that the scale invariant initialisation was working great with SGD and quite poorly with SR that then needed more tuning, so we did not merge it.

Can you try to run a bunch of benchmark simulations both with SGD and SR for different sizes before and after this PR to see how things change?

attila-i-szabo:
Thanks for this. I know nothing about RBM, so if you say it wasn't working great, I won't push to change it in any way.

As I said, deeper GCNNs have been scale-invariant already (the `unit_normal_scaling` that I've deleted is equivalent to `lecun_normal` without truncation), and it has to be if it is not to blow up. I've tried this code in production (on a 4-layer network) and it seems to work just as well as the previous one. SR usually gets slow later in the training when it's near the ground state where the S matrix is the worst-conditioned.

PhilipVinc:
I don't want to stop you from doing this! If this works, i'm all in for merging it. I'm just asking for a bunch of plots like https://github.com/netket/netket/pull/863#issuecomment-899335451 but both with and without sr

netket/nn/symmetric_linear.py
```diff
@@ -20,58 +20,18 @@
...
-
-    return init
-
-
-def unit_normal_scaling(key, shape, dtype):
-    return jax.random.normal(key, shape, dtype) / jnp.sqrt(
-        jnp.prod(jnp.asarray(shape[1:]))
-    )
+# All layers defined here have kernels of shape [out_features, in_features, n_symm]
+default_gcnn_initializer = lecun_normal(in_axis=1, out_axis=0)
```
chrisrothUT, line 34: I'm pretty sure `in_axis = (1,2)` here. And don't we have to specify `mode="fan_in"`?

chrisrothUT (changes requested):

netket/nn/symmetric_linear.py
attila-i-szabo: Yeah, that's confusing, `in_axis` is the axes that **only** appear in the input but not the output, so axis 2 is neither. And `lecun_normal` implies `fan_in`

attila-i-szabo: See the code [here](https://jax.readthedocs.io/en/latest/_modules/jax/_src/nn/initializers.html)

chrisrothUT: Ahh I see, yes the code as written should work.

chrisrothUT (approved):

attila-i-szabo:
OK. Benchmarks. As predicted, the performance of GCNNs doesn't noticeably change (I think any difference in these curves is within the range of outcomes for one initializer).

I used two Hamiltonians, both on the square lattice
* NN antiferromagnet with Marshall sign rule → use real-parameter network
* J1-J2 antiferromagnetic (J2=J1/2) → use complex-parameter network

In all figures, orange is the new initialiser, blue is the old one.

SR, NN model, 4×4 lattice, 6 layers, 4 features
image

SR, J1-J2, 4×4 lattice, 6 layers, 4 features
image

SR, J1-J2, 6×6 lattice, 4 layers, 4 features
image

SGD, NN model, 6×6 lattice, 2 layers, 4 features
image

SGD was notoriously hard to stabilise with either initializer, it didn't work at all for the J1-J2 model, nor for deeper networks with the NN model.

femtobit (approved):
Looks good to me, I have no objections. I'll leave the decision to merge to @PhilipVinc.

chrisrothUT:
There's also a minor bug in the GCNN models which is that this line

`x = jnp.expand_dims(x, -2) ` # add a feature dimension

should really be  

```
if x.ndim < 3:
  x = jnp.expand_dims(x, -2)
```

so that we only expand the dimension if we don't have a onehot representation. Maybe we could sneak that minor change into this PR

attila-i-szabo:
Good catch. There's an equivalent line in `RBMSymm`, I changed that too.

We happy, @PhilipVinc'ent?

PhilipVinc:
yes I am happy though can you add a changeling notice for both the bug fix and the PR itself? (While you are at it, since it's the first commit, can you change the version 3.4 banner to 3.3.1 in the changelog and remove all sections other than  bug fix?

PhilipVinc (approved):
I'll merge when a changelog notice is added

netket/nn/symmetric_linear.py
```diff
@@ -20,58 +20,18 @@
...
-            return _complex_truncated_normal(key, 2, shape, dtype) * stddev
-
-    return init
-
-
-def unit_normal_scaling(key, shape, dtype):
-    return jax.random.normal(key, shape, dtype) / jnp.sqrt(
-        jnp.prod(jnp.asarray(shape[1:]))
-    )
+# All layers defined here have kernels of shape [out_features, in_features, n_symm]
```
PhilipVinc, line 33: I just have an extra comment: This file is called `symmetric_linear` and contains equivariant layers. GCNN is a NN architecture built on top of equivariant layers that, in principle, we should be able to use alone without the gcnn.

It's therefore a bit weird for me to find here `default_gcnn_initializer` defined. I'd much more like an initialiser named `default_gcnn_initializer`. Probably a better name could be something like `default_equivariant_initializer` or `default_symmetry_initializer`? Maybe @jwnys has a better idea for a name.

netket/models/equivariant.py
```diff
@@ -33,19 +33,14 @@
     DenseSymmFFT,
     DenseEquivariantFFT,
     DenseEquivariantIrrep,
+    default_gcnn_initializer,
```
PhilipVinc, line 36: Ah, I just noticed now, but...
in principle all things you are importing here from `symmetric_linear` are 'implementation details. You should only access `netket.nn.DenseSymm` and `netket.nn.DenseEquivariant`.

While this is intra-netket and using internals is fine, I'd like to avoid it if possible.

For what regards this particular PR, I'd be fine with not importing the default initialiser and just redefining it here + a comment below.

Then, it would be great if in a follow up PR you could just use the standard API `DenseSymm(mode="FFT")` or `DenseEquivariant(mode=irrep)` instead of using the type directly.

This (1) tests the public API, so it's good, (2) makes us use our public API and (3) disentangles the internals of the two modules.

Moreover, maybe by doing this we can get rid of having 4 different GCNN classes?

netket/nn/symmetric_linear.py
PhilipVinc, line 34: By the way, is this supported in all jax versions that we support? I remember talking about raising the jax version with @attila-i-szabo ?

netket/nn/symmetric_linear.py
```diff
@@ -105,14 +65,18 @@ class DenseSymmMatrix(Module):
...
-    """Initializer for the bias. Defaults to zero initialization"""
+    """Initializer for the bias. Defaults to zero initialization."""
 
     def setup(self):
         # pylint: disable=attribute-defined-outside-init
         self.n_symm, self.n_sites = np.asarray(self.symmetries).shape
+        if self.mask is not None:
+            self.scaled_mask = (
+                self.mask / jnp.linalg.norm(self.mask) * self.n_sites ** 0.5
+            )
```
PhilipVinc, line 79: The mask must be an hashablearray, right? `jnp.linalg.norm(self.mask)` works because it's doing essentially `jnp.linalg.norm(jnp.array(np.asarray(self.mask)))`. 
But you are doing

```python
self.mask / jnp.linalg.norm(Self.mask)
```
which is converted to
```
jnp.array(self.mask) / jnp.linalg.norm(jnp.array(np.asarray(self.mask)))
```
which captures the mask twice.

I think a better implementation would be to capture only the rescaled mask once:

```suggestion
        if self.mask is not None:
            self.scaled_mask = jnp.asarray(
                self.mask / np.linalg.norm(self.mask) * self.n_sites ** 0.5)
```
(probably you need to import numpy and reformat too)

PhilipVinc: Also can we assert that the bask is boolean and not integers or other things? 
Or do we need in some cases the mask to be a finite number?

netket/nn/symmetric_linear.py
```diff
@@ -212,6 +170,11 @@ def setup(self):
         self.n_point = len(sg) // self.n_cells
         self.sites_per_cell = sg.shape[1] // self.n_cells
 
+        if self.mask is not None:
```
PhilipVinc, line 173: as above

netket/nn/symmetric_linear.py
```diff
@@ -444,13 +399,18 @@ class DenseEquivariantIrrep(Module):
...
+    """Initializer for the kernel. Defaults to Lecun normal."""
     bias_init: NNInitFunc = zeros
-    """Initializer for the bias. Defaults to zero initialization"""
+    """Initializer for the bias. Defaults to zero initialization."""
 
     def setup(self):
         self.n_symm = self.irreps[0].shape[0]
+        if self.mask is not None:
+            self.scaled_mask = (
+                self.mask / jnp.linalg.norm(self.mask) * self.n_symm ** 0.5
```
PhilipVinc, line 411: as above. maybe you can make a function ?

netket/nn/symmetric_linear.py
```diff
@@ -591,14 +551,17 @@ class DenseEquivariantMatrix(Module):
...
+    kernel_init: NNInitFunc = default_gcnn_initializer
+    """Initializer for the kernel. Defaults to Lecun normal."""
     bias_init: NNInitFunc = zeros
-    """Initializer for the bias. Defaults to zero initialization"""
+    """Initializer for the bias. Defaults to zero initialization."""
 
     def setup(self):
-
         self.n_symm = np.asarray(self.product_table).shape[0]
+        if self.mask is not None:
```
PhilipVinc, line 561: .

test/models/test_nn.py
```diff
@@ -187,7 +187,8 @@ def test_DenseEquivariant(symmetries, use_bias, lattice, mode, mask):
     n_symm = np.asarray(perms).shape[0]
 
     if mask:
-        mask = np.random.randint(0, 2, [n_symm])
+        mask = np.zeros(n_symm)
```
PhilipVinc, line 190: do we test that the mask actually masks?

PhilipVinc (changes requested):

netket/nn/symmetric_linear.py
attila-i-szabo: Yeah, this is fine for both `float` and `complex`. It would've been a problem if we kept 2D kernels in `DenseSymm` because that's only supported from 2.26

test/models/test_nn.py
attila-i-szabo: wdym?

netket/nn/symmetric_linear.py
attila-i-szabo: I've never used this feature so no strong feelings, but it might be useful to control the initialisation through this mask. The scaling logic is happy with that too.

netket/models/equivariant.py
attila-i-szabo: `DenseSymm()` and `DenseEquivariant()` take things that don't play nicely at all with jitting, so if you put these functions in `setup()` of another module, their construction gets very slow. I seem to recall we tried this at some point (but @chrisrothUT  might remember better) and decided against it when we saw the performance.

netket/nn/symmetric_linear.py
attila-i-szabo: I generally don't like restricting inputs unless they break stuff/lead to something very absurd. This is neither.

PhilipVinc (approved):
Thanks for addressing all my comments @attila-i-szabo .

I think one test that is missing and if you have some spare minutes in the future would be one checking that the mask is actually masking (in the sense that it affects the output in the correct way).

Anyhow, this is great and I'll merge it.

PhilipVinc:
Another thing that might be worth investigating briefly before releasing this is if lecun_normal has the right variance wrt the default activation (reim_selu)

attila-i-szabo:
> I think one test that is missing and if you have some spare minutes in the future would be one checking that the mask is actually masking (in the sense that it affects the output in the correct way).

Fair enough, the masking logic is hardly tested in general, I'll try to tend to this at some point.

> Another thing that might be worth investigating briefly before releasing this is if lecun_normal has the right variance wrt the default activation (reim_selu)

I did check this (in fact that's why I started this PR). `lecun_normal` ensures that the expected |x|^2 remains 1 for both real and complex in all layers. SELU normally wants real/imag. parts with variance 1; we added factors of sqrt(2) to `reim_selu` with complex inputs to correct for this.

PR #1168, opened 2022-04-19
Bump minimum jax version to 0.2.25

femtobit:
Increase the minimum required jax version to fix CI crashes.


femtobit:
Looks like this helps. Can we merge? @PhilipVinc

PhilipVinc (approved):

femtobit:
Great. The tests did actually pass here (whereas they failed reliably in other PRs), but apparently that was not enough: The CI run after merging this, which should run the exact same code, [fails again](https://github.com/netket/netket/runs/6084408327?check_suite_focus=true). 🙁 

This problem is surprisingly persistent.

PR #1170, opened 2022-04-20
Add `MCState.log_gradient` implementation

femtobit:
This is a first draft of a PR which adds a `MCState.log_gradients` function that, for given samples, returns the corresponding scores `O_k(s) = ∂[ln ψ(s)]/∂θ_k`.

This is done internally by calling `qgt_jacobian_pytree_logic.py::prepare_centered_oks` and is thus essentially the same code path `QGTJacobianPyTree` uses to compute its `.O` member. (This PR currently renames `prepare_centered_oks` to `prepare_log_gradients` since centering is now optional.)

`log_gradient` accepts a `mode` option (with the same meaning as for `QGTJacobian*`). There is one difference, though: `mode=complex` in `QGTJacobianPyTree`  creates the log gradients `qgt.O` as real-valued arrays with stacked real and imaginary parts of the complex `O`. `log_gradients` avoids this (since I don't think returning this to users by default is a good choice), but this comes at the cost of adding yet another option to `prepare_log_gradients`.

There are several things to address before this can be merged (like adding tests), but I'd like some feedback on this first draft (@inailuig @PhilipVinc @gcarleo @attila-i-szabo @danielalcalde):

1. Does the general design make sense? Reusing code from `qgt_jacobian_*` seems to me like the best way to prevent code duplication (and we already support chunking etc. in `log_gradients`).
2. Do we need to add an option to obtain the log gradients by generalizing `qgt_jacobian_dense_logic.py::prepare_centered_oks` in the same way this is done for the pytree version right now? That would probably be useful in the same cases where `QGTJacobianDense` is a good choice, right?
3. There is probably a way to clean up all the nested `if` branches in `prepare_centered_oks`/`prepare_log_gradients` and make the logic less convoluted. Any suggestions (@inailuig)?
4. I am not convinced it is worth the additional effort (since these methods are internal anyways), but conceptually it could make more sense to move most of the Jacobian code to something like `netket.jax._jacobian_{dense,pytree}` (from `netket.optimizer.qgt.qgt_jacobian_{dense,pytree}_logic`) as most of it is not actually that QGT specific.

### Example
```python3
import netket as nk
import jax

L = 4
ma = nk.models.RBMModPhase(alpha=1, use_hidden_bias=False, dtype=float)

vs = nk.vqs.MCState(sa, ma, n_samples=1024)
```
```python3
>>> print(jax.tree_map(jax.numpy.shape, vs.parameters))
FrozenDict({
    Dense_0: {
        kernel: (4, 4),
    },
    Dense_1: {
        kernel: (4, 4),
    },
})
>>> print(jax.tree_map(jax.numpy.shape, vs.log_gradient(vs.samples)))
FrozenDict({
    Dense_0: {
        kernel: (1024, 4, 4),
    },
    Dense_1: {
        kernel: (1024, 4, 4),
    },
})
>>> print(vs.log_gradient(vs.samples))
FrozenDict({
    Dense_0: {
        kernel: DeviceArray([[[-3.29284369e-04+0.j,  5.88890101e-04+0.j,
                       -1.15402368e-05+0.j, -1.53835420e-03+0.j],
                      [ 3.29284369e-04+0.j, -5.88890101e-04+0.j,
...
```

### DIFF
--- a/netket/optimizer/qgt/qgt_jacobian_pytree.py
+++ b/netket/optimizer/qgt/qgt_jacobian_pytree.py
@@ -26,10 +26,40 @@ import netket.jax as nkjax
 from ..linear_operator import LinearOperator, Uninitialized
 
 from .common import check_valid_vector_type
-from .qgt_jacobian_pytree_logic import mat_vec, prepare_centered_oks
+from .qgt_jacobian_pytree_logic import mat_vec, prepare_log_gradients
 from .qgt_jacobian_common import choose_jacobian_mode
 
 from netket.nn import split_array_mpi
+
+
+def log_gradients_pytree(
+    vstate,
+    samples,
+    *,
+    centered: bool,
+    pdf=None,
+    mode: str,
+    split_complex: bool,
+    rescale_shift=False,
+):
+    if hasattr(vstate, "chunk_size"):
+        chunk_size = vstate.chunk_size
+    else:
+        chunk_size = None
+
+    O, scale = prepare_log_gradients(
+        vstate._apply_fun,
+        vstate.parameters,
+        samples.reshape(-1, samples.shape[-1]),
+        vstate.model_state,
+        mode,
+        rescale_shift,
+        centered,
+        pdf,
+        chunk_size,
+        split_complex=split_complex,
+    )
+    return O, scale
 
 
 def QGTJacobianPyTree(
@@ -92,20 +122,14 @@ def QGTJacobianPyTree(
     elif holomorphic is not None:
         raise ValueError("Cannot specify both `mode` and `holomorphic`.")
 
-    if hasattr(vstate, "chunk_size"):
-        chunk_size = vstate.chunk_size
-    else:
-        chunk_size = None
-
-    O, scale = prepare_centered_oks(
-        vstate._apply_fun,
-        vstate.parameters,
-        samples.reshape(-1, samples.shape[-1]),
-        vstate.model_state,
-        mode,
-        rescale_shift,
-        pdf,
-        chunk_size,
+    O, scale = log_gradients_pytree(
+        vstate,
+        samples,
+        centered=True,
+        split_complex=mode == "complex",
+        pdf=pdf,
+        mode=mode,
+        rescale_shift=rescale_shift,
     )
 
     return QGTJacobianPyTreeT(
--- a/netket/optimizer/qgt/qgt_jacobian_pytree_logic.py
+++ b/netket/optimizer/qgt/qgt_jacobian_pytree_logic.py
@@ -200,20 +200,35 @@ def _mat_vec(v: PyTree, oks: PyTree) -> PyTree:
 # here the other modes are converted
 
 
-@partial(jax.jit, static_argnames=("apply_fun", "mode", "rescale_shift", "chunk_size"))
-def prepare_centered_oks(
+@partial(
+    jax.jit,
+    static_argnames=(
+        "apply_fun",
+        "mode",
+        "rescale_shift",
+        "chunk_size",
+        "centered",
+        "split_complex",
+    ),
+)
+def prepare_log_gradients(
     apply_fun: Callable,
     params: PyTree,
     samples: Array,
     model_state: Optional[PyTree],
     mode: str,
     rescale_shift: bool,
+    centered: bool,
     pdf=None,
     chunk_size: int = None,
+    split_complex: bool = False,
 ) -> PyTree:
     """
-    compute ΔOⱼₖ = Oⱼₖ - ⟨Oₖ⟩ = ∂/∂pₖ ln Ψ(σⱼ) - ⟨∂/∂pₖ ln Ψ⟩
-    divided by √n
+    compute Oⱼₖ = Oⱼₖ = ∂/∂pₖ ln Ψ(σⱼ)
+    or      ΔOⱼₖ = Oⱼₖ - ⟨Oₖ⟩ = ∂/∂pₖ ln Ψ(σⱼ) - ⟨∂/∂pₖ ln Ψ⟩
+    divided by √n.
+
+    The `centered` argument determines whether Oⱼₖ or ΔOⱼₖ is computed.
 
     In a somewhat intransparent way this also internally splits all parameters to real
     in the 'real' and 'complex' modes (for C→R, R&C→R, R&C→C and general C→C) resulting in the respective ΔOⱼₖ
@@ -226,8 +241,10 @@ def prepare_centered_oks(
         model_state: untrained state parameters of the model
         mode: differentiation mode, must be one of 'real', 'complex', 'holomorphic'
         rescale_shift: whether scale-invariant regularisation should be used (default: True)
+        centered: Whether to center the returned gradients around zero by subtracting the sample mean.
         pdf: |ψ(x)|^2 if exact optimization is being used else None
         chunk_size: an int specfying the size of the chunks the gradient should be computed in (default: None)
+        split_complex: If mode=="complex", split result into stacked real and imaginary part.
 
     Returns:
         if not rescale_shift:
@@ -238,6 +255,9 @@ def prepare_centered_oks(
             pytree containing the norms that were divided out (same shape as params)
 
     """
+    if mode != "complex" and split_complex:
+        raise ValueError("split_complex=True can only be used with mode='complex'")
+
     # un-batch the samples
     samples = samples.reshape((-1, samples.shape[-1]))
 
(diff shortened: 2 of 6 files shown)


danielalcalde:
What do you think of also adding @attila-i-szabo suggestion from #1154 to this PR:
>I think a better solution would be adding a jacobian: Optional[PyTree] field to MCState (I would keep it as a pytree, that's how it's first generated anyway and Dense is already set up to convert it to a matrix). Then, QGTJacobian* can check if there is something in this field, use it if yes or compute it from scratch (and memoise it in the vstate) if no. The Jacobian only depends on things that are part of the vstate, so it makes sense to keep it there. Then, the only thing that needs to be special-cased is letting the driver know it should compute the Jacobian together with expect_and_grad.

attila-i-szabo:
> 1\. Does the general design make sense? Reusing code from `qgt_jacobian_*` seems to me like the best way to prevent code duplication (and we already support chunking etc. in `log_gradients`).
>
> 4\. I am not convinced it is worth the additional effort (since these methods are internal anyways), but conceptually it could make more sense to move most of the Jacobian code to something like `netket.jax._jacobian_{dense,pytree}` (from `netket.optimizer.qgt.qgt_jacobian_{dense,pytree}_logic`) as most of it is not actually that QGT specific.

Strongly agree with both. Also, I don't like that the Jacobians for QGTs are not routed through `vstate.log_gradient`, so if we improve the implementation there (e.g. by memoising the Jacobian, which kicked this project off), that wouldn't carry over to QGT automatically. My ideal design would roughly be:
* Jacobian computation routines in `nk.jax` (I doubt it takes much more than moving the files)
* `log_gradient` in `MCState`
* QGT constructors that either take the Jacobian directly (for `PyTree`) or flatten it (for `Dense`)

> 2\. Do we need to add an option to obtain the log gradients by generalizing `qgt_jacobian_dense_logic.py::prepare_centered_oks` in the same way this is done for the pytree version right now? That would probably be useful in the same cases where `QGTJacobianDense` is a good choice, right?

It would be nice to offer it. Also, we could use this to refactor the `dense` logic—atm, dense gradients are prepared on a completely distinct code path, instead of ravelling and concatenating the leaves of the `pytree` gradients...

> 3\. There is probably a way to clean up all the nested `if` branches in `prepare_centered_oks`/`prepare_log_gradients` and make the logic less convoluted. Any suggestions (@inailuig)?

`centered_*` and `*` differ by an additional call to `tree_subtract_mean`:
https://github.com/netket/netket/blob/7c50aeae3b37a571b75909bf731c2635a4a5c91a/netket/optimizer/qgt/qgt_jacobian_pytree_logic.py#L108-L109
we could have something like this inside `prepare_centered_oks` to eliminate all the `if` statements about `centered`:
```python
if centered:
    centered_jacobian_fn = compose(tree_subtract_mean, centered_jacobian_fun)
```
BTW, what is the whole alternate path with `pdf` for? Why do we need `jacobian_fun for that separately?

> `log_gradient` accepts a `mode` option (with the same meaning as for `QGTJacobian*`). There is one difference, though: `mode=complex` in `QGTJacobianPyTree` creates the log gradients `qgt.O` as real-valued arrays with stacked real and imaginary parts of the complex `O`. `log_gradients` avoids this (since I don't think returning this to users by default is a good choice), but this comes at the cost of adding yet another option to `prepare_log_gradients`.

This seems sensible to me. In any case, we'll have to document very carefully what exactly `log_gradient` returns. For example, we do split the real&imag. parts of the parameters, which is not a trivial change, and users need to know how the pytree structure changes.

> (This PR currently renames `prepare_centered_oks` to `prepare_log_gradients` since centering is now optional.)

`centered` controls only whether the mean is subtracted from the Jacobian, not the `sqrt(n_samp)` scaling, right? I would change the logic s.t. we only do that if `centered is True`. Otherwise we can add _yet another switch_ but it doesn't make sense to do it if one just wants the gradients.

femtobit:
Adding some caching of the log-gradients could be useful. The first step would be to make the `s` argument optional and have `log_gradient` return the gradient for the current `MCState.samples` and cache the result inside `MCState` too. That is easy to implement. 

Unfortunately, I think the subsequent steps are not as straightforward:
At the moment, the computed gradients be used to construct `QGTJacobianPyTree` and, with some additional code, `QGTJacobianDense`, but only if the other options such as `mode` and `rescale_shift` match - so already we have to memorize both the gradients and other options so the QGT could check whether its actually compatible. If we support dense and pytree gradients, we would also need a conversion the other way around.

I am not sure this added complexity is worth it. A simpler alternative is to make it easy to create `QGTJacobian*` from the pre-computed gradients outside of `MCState` (say by adding a `.from_gradients` static method), so that something like this is possible:
```
grads = vstate.log_gradient(mode=mode, centered=True)
qgt = QGTJacobianPyTreeT.from_gradients(grads, mode=mode)
```
At least for the purpose of the TDVP, this would be sufficient and is simpler to implement and maintain.

femtobit:
@attila-i-szabo Thanks for the suggestions, they sound very good to me. I'll simplify the implementation for pytree.

By the way, is the current implementation of the dense Jacobian code more efficient than ravelling the result of the pytree code afterwards? That would be a valid reason to not merge the implementations fully.

attila-i-szabo:
> I am not sure this added complexity is worth it. A simpler alternative is to make it easy to create `QGTJacobian*` from the pre-computed gradients outside of `MCState` (say by adding a `.from_gradients` static method), so that something like this is possible:

The starting motivation for doing this was to be able to get local energies, energy gradients, and Jacobians in one go from a `MCState.expect_grad_jacobian` function, which @danielalcalde pointed out is much more efficient than `expect_and_grad` and a separate round of computing the gradients. In the current implementation of the drivers, this could only be done if `MCState` can hold on to the Jacobian from the point it's computed in `expect_and_...` and then feed it into the QGT constructor. 

But I agree that storing the Jacobian in the vstate is probably not the best. Other than having to deal with all the options (in the worst case you could memoise those too and recompute if they don't match), the bigger problem is that if we use `QGTDense`, the same Jacobian would be stored once as a pytree in the vstate and once as an array in the QGT, and the Jacobian is too big for to duplicate it...

I guess the real solution for this would be to have a separate class for computing updates (as I suggested here: https://github.com/netket/netket/pull/1142#issuecomment-1102276885) that can handle the Jacobian as it wants to without storing it in the vstate.

> EDIT: We could probably also change the `QGTJacobianPyTree` function directly, it already deals with different kinds of input so it could additionally check whether `MCState` or an array is passed too I guess.)

IIRC we'd still need an `MCState` field to have access to pytree shapes, etc. The best pattern would probably be an optional `jacobian` argument in the "constructor"

> At least for the purpose of the TDVP, this would be sufficient and is simpler to implement and maintain.

I'm never gonna tire of repeating that we can't just implement things as central as this for one or the other use case of netket...

> By the way, is the current implementation of the dense Jacobian code more efficient than ravelling the result of the pytree code afterwards? That would be a valid reason to not merge the implementations fully.

No clue, I don't think the implementation I suggest ever existed. @PhilipVinc, do you know of any benchmarks for this?
I don't think it would be slower though: right now we ravel sample by sample and `vmap` over it, which doesn't sound any better than reshaping the full arrays and concatenating them.

femtobit:
> The starting motivation for doing this was to be able to get local energies, energy gradients, and Jacobians in one go from a MCState.expect_grad_jacobian function, which @danielalcalde pointed out is much more efficient than expect_and_grad and a separate round of computing the gradients.

Yes, indeed, an `expect_grad_jacobian` function could be added. That doesn't need the caching, though: it could just compute all three internally (without calling `expect_and_grad` in the default case) and return them. Not relying on `expect_and_grad` for that is necessary since it doesn't actually compute `O`s which could be cached but uses a vjp:
https://github.com/netket/netket/blob/1ba051e332614b23cf4d876becdad92904d2d017/netket/vqs/mc/mc_state/expect_grad.py#L149-L155
It could work the other way round (caching the gradients from `log_gradient` and/or `.quantum_geoemtric_tensor` to speed of `expect_and_grad`), but this would need a different implementation compared to the current `expect_and_grad` too.

Note also that the gradient returned from `expect_and_grad` cannot always be used for time propagation: In the real-time and real-parameter case, `expect_and_grad` returns `∇E = 2Re[cov(O, H)]` which is real but for the real-time TDVP (at least the McLachlan version of it) we need the complex `F = cov(O, H)` (to compute `Re[-i F] = Im[F]`) which is not returned at the moment. So in that case, unless we want to change more parts of the `expect_and_grad` code, going the other way and computing everything from the explicitly obtained `log_gradient` seems easier.

PR #1196, opened 2022-05-09
Propagator of a Hamiltonian

VolodyaCO:
This PR introduces an experimental feature analogous to the `.expm()` method in qutip, that allows the construction of the `e^{-iA}` for some `LocalOperator` `A`. The implementation is based on approximating the complex exponential with Chebysheb polynomials, which seem to be quite performant in terms of the number of products that need to be carried out.

If you think that this is a valuable addition, please tell me where to place it and I can take care of testing. For the moment I see that product of two operators is quite costly in netket. Why is that? Is there a plan to accelerate this? Has this been diagnosed before?

### Integrating Jaynes-Cummings
I have been trying, for the good part of last month, to integrate the Jaynes Cummings hamiltonian (which describes an optical mode coupled to an exciton, or a two-level-system) and show some evolution with TDVP integration with an initial state consisting of 0 photons and 1 exciton.

Apparently, since the spread of the probability is non-existing, i.e., all probability is assigned to a single state of the bare basis, TDVP has a lot of difficulties making the probability flow from that state to others, theoretically requiring very small integration time steps. Experimentation didn't yield good results and I was not able to ever use TDVP to integrate the JC hamiltonian.

On the other hand, I also tried including a random rotation of the basis so that the probability is more distributed in this new basis. This, however, induces a dense Hamiltonian, which completely thrashes the `LocalOperator` idea of compactly writing an operator.

**Note**: I prepared the initial state of 0 photons and 1 exciton in a full-forward neural network, and then I used the variational state to compute the expected value of the 0photon-1exciton projector, which yielded the correct Rabi oscillations

### Reproduction of https://arxiv.org/abs/2204.03454

I am also trying to reproduce the results in this paper, which uses the Feynman-Kitaev formalism to encode the history of a quantum system in the ground state of a modified problem. For this, I need to compute `Udt = exp(-i dt H)`, for some hamiltonian H, and then I'm able to construct the Feynman-Kitaev hamiltonian.

In general, the propagator is not sparse, even if the Hamiltonian is sparse. But that's the price we have to pay if we want to use the Feynman-Kitaev hamiltonian.

### DIFF
--- a/netket/experimental/operator/__init__.py
+++ b/netket/experimental/operator/__init__.py
@@ -17,6 +17,7 @@ __all__ = ["FermionOperator2nd"]
 
 from ._fermions_2nd import FermionOperator2nd
 from . import fermion
+from .propagator import propagator
 
 from netket.utils import _auto_export
 
--- a/netket/experimental/operator/propagator.py
+++ b/netket/experimental/operator/propagator.py
@@ -15,6 +15,30 @@
 from netket.operator import LocalOperator
 import math
 import numpy as np
+import time
+
+_chebysheb18_constants = {
+    "a11": 3 / 25,
+    "a21": -0.00877476096879703859j,
+    "a31": -0.00097848453523780954,
+    "b11": -0.66040840760771318751j,
+    "b21": -1.09302278471564897987,
+    "b31": 0.25377155817710873323j,
+    "b61": 0.00054374267434731225,
+    "b02": -2.58175430371188142440,
+    "b12": -1.73033278310812419209j,
+    "b22": -0.07673476833423340755,
+    "b32": -0.00261502969893897079j,
+    "b62": -0.00003400011993049304,
+    "b03": 2.92377758396553673559,
+    "b13": 1.44513300347488268510j,
+    "b23": 0.12408183566550450221,
+    "b33": -0.01957157093642723948j,
+    "b63": 0.00002425253007433925,
+    "b24": -0.123953695858283131480j,
+    "b34": -0.011202694841085592373,
+    "b64": -0.000012367240538259896j,
+}
 
 
 def _one_norm(op: LocalOperator) -> float:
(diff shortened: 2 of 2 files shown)


VolodyaCO:
Oh yeah I got the reference wrong from the clipboard: it's correct now.

> Could you please tell see more about the function you implemented and what would be the use-case?
If I understand correctly, you see C. approximation to write down a (less dense) version of the propagator, but that will still be dense...?

I think that the most sensible application would be to compute U(dt) for small (dt), which ensures that the operator is "not very dense", and use U(dt) as a building block for time evolution as a ground-state problem.

PhilipVinc:
can you show how much 'not dense' this turns out to be as a function of dt?

VolodyaCO:
Hey @PhilipVinc sorry for the delay. For the Bose Hubbard model (4 sites and local size 4) I count the nonzero values of the propagator (I tidy up values with a magnitude less than 1e-12 as qutip does) and I see this behaviour
Screenshot 2022-05-18 at 5 38 18 PM

The value quickly grows and stabilises.

PhilipVinc:
I'm more interested in the what you find in `operator.acting_on` and `np.max(operator._n_conns, axis=1)` for a bunch of different values of dt

PhilipVinc:
that's only present on local operators.
you need to call `._setup()` to generate those structures, or perform any `get_conn` operations.

VolodyaCO:
`.setup()` and `.setup(force=True)` didn't work. Calling `get_conn` did not work either, to produce the attribute `_n_conns`.

PhilipVinc:
```python
>>> n_sites = 4
>>> hi = nk.hilbert.Fock(N=n_sites, n_max=3)
>>> op = nk.operator.BoseHubbard(hi, nk.graph.Chain(n_sites), U=1)
>>> oplocal = op.to_local_operator()
>>>
>>> oplocal._setup()
>>> oplocal._n
oplocal._n_conn_from_sections(  oplocal._n_conns                oplocal._nonzero_diagonal
>>> oplocal._n_conns
array([[ 0,  0,  0,  0, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1],
       [ 0,  0,  0,  0, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1],
       [ 0,  0,  0,  0, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1],
       [ 0,  0,  0,  0, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1],
       [ 0,  1,  1,  1,  1,  2,  2,  1,  1,  2,  2,  1,  1,  1,  1,  0],
       [ 0,  1,  1,  1,  1,  2,  2,  1,  1,  2,  2,  1,  1,  1,  1,  0],
       [ 0,  1,  1,  1,  1,  2,  2,  1,  1,  2,  2,  1,  1,  1,  1,  0],
       [ 0,  1,  1,  1,  1,  2,  2,  1,  1,  2,  2,  1,  1,  1,  1,  0]])
```

VolodyaCO:
I don't know why it wasn't working. Looks like my branch was way behind in commits, but doesn't make much sense. Anyhow, here's the plot for `_n_conns` (for each dt I saved the whole `np.max(operator._n_conns, axis=1)` and plot the series):
Screenshot 2022-05-20 at 8 38 24 PM

and for `len(U.acting_on)` it is constant.

PhilipVinc:
The number of nonzero connected elements per term in the hamiltonian as a function of dt.
This basically shows that for dt $10^{-2}$ you're already almost saturated to a K=Nsites-local matrix so not really local...

@VolodyaCO from this plot I get to estimate that unless you're using $dt<10^{-3}$ this method is going to be exponentially expensive...

VolodyaCO:
Yes @PhilipVinc I think that the most sensible application of this is to solve time dynamics as a ground state problem with a small dt (with the Feynman Kitaev formalism). In general, for large enough dt's the propagator will no longer be sparse, and the `LocalOperator` structure will succumb.

PhilipVinc:
Ultimately unplanned

PR #1199, opened 2022-05-13
Example + DeepSet

gpescia:
This adds the DeepSet model and the example of Helium in 1d (and removes old calogero).


netket/models/deepset.py
```diff
@@ -0,0 +1,104 @@
...
+from flax import linen as nn
+from netket.utils.types import NNInitFunc
+from jax.nn.initializers import (
+    zeros,
+    ones,
+    lecun_normal,
+)
+
+
+class DeepSet(nn.Module):
```
PhilipVinc, line 14: Can you add a doc string with a reference to literature and a formula or two?

netket/models/deepset.py
```diff
@@ -0,0 +1,104 @@
...
+    L: jnp.float64  # boxsize
+    sdim: int  # nbr of spatial dimensions
+
+    """Number of layers in phi/rho network"""
```
PhilipVinc, line 18: The doc strings should be one per attribute, in the line right below (not above)

docs/docs/changelog.md
```diff
@@ -1,5 +1,8 @@
 (changelog)=
 
 ```{include} ../../CHANGELOG.md
+Added permutation equivariant DeepSet architecture.
```
PhilipVinc, line 4: Not here! In the file in the root directory, named CHANGELOG.md

This one here is only including it in the website but it is not the changelog itself

netket/models/deepset.py
```diff
@@ -0,0 +1,129 @@
...
+    """Number of layers in phi network"""
+    layers_rho: int
+    """Number of layers in rho network"""
+
+    features_phi: Union[Tuple, int]
+    """Number of features in each layer for phi network."""
+    features_rho: Union[Tuple, int]
+    """Number of features in each layer for rho network."""
+
+    cusp_exponent: int = 0.0
```
PhilipVinc, line 47: Can we make this an `Optional[int]` and have it None by default instead of 0? 
And in the apply method check that if cusp is None, then don't apply it (so it won't compute distances, etc)

netket/models/deepset.py
```diff
@@ -0,0 +1,129 @@
...
+
+        \psi(x_1,...,x_N) = \rho\left(\sum_i \phi(d_{\sin}(x_i,x_j))\right) \cdot \exp\left[-\frac{1}{2}\left(b/d_{\sin}(x_i,x_j)\right)^5\right]
+
+    """
+    L: jnp.float64
+    """boxsize"""
+    sdim: int
+    """number of spatial dimensions"""
+
+    layers_phi: int
```
PhilipVinc, line 37: We don't need to specify the layers

netket/models/deepset.py
```diff
@@ -0,0 +1,129 @@
...
+
+    kernel_init: NNInitFunc = lecun_normal()
+    """Initializer for the Dense layer matrix"""
+    bias_init: NNInitFunc = zeros
+    """Initializer for the hidden bias"""
+    params_init: NNInitFunc = ones
+    """Initializer for the parameter in the cusp"""
+
+    def setup(self):
+        self.phi = [
```
PhilipVinc, line 67: Can you add some check like this?

```python
    def setup(self):
        if isinstance(self.features, int):
            features = [self.features] * (self.layers - 1)
        else:
            features = self.features
``

**Review by PhilipVinc — COMMENTED** [review]
Ok, last remarks

File: netket/models/deepset.py [hunk 7]
```diff
@@ -0,0 +1,129 @@
...
+    For helium the Ansatz reads:
+
+    .. math ::
+
+        \psi(x_1,...,x_N) = \rho\left(\sum_i \phi(d_{\sin}(x_i,x_j))\right) \cdot \exp\left[-\frac{1}{2}\left(b/d_{\sin}(x_i,x_j)\right)^5\right]
+
+    """
+    L: jnp.float64
+    """boxsize"""
+    sdim: int
```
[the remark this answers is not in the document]
gpescia: yes true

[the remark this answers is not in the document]
PhilipVinc: ah right. ok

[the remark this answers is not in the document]
PhilipVinc: shouldn't we just pass the Hilbert space here, then?

PhilipVinc:
I fixed it

PhilipVinc (approved):

PR #1221, opened 2022-06-01
Fix inline math directive in docstrings

nikosavola:
I noticed some math in the API was not rendering correctly so this PR fixes those inline math directives in docstrings.


PhilipVinc (approved):
Thanks!

PR #1248, opened 2022-06-15
states_to_local_indices for fermions

jwnys:
As the title says, solves #1246  for fermions


PhilipVinc (approved):
I liked the previous PR title..🥲

PR #1259, opened 2022-06-24
Invalid Loss Stopping callback

VolodyaCO:
This PR implements the invalid loss stopping callback and tests it, as required in #1239

### DIFF
--- a/netket/callbacks/invalidlossstopping.py
+++ b/netket/callbacks/invalidlossstopping.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 from dataclasses import dataclass
+from typing import Union
 
 import numpy as np
 
@@ -24,6 +25,11 @@ class InvalidLossStopping:
 
     monitor: str = "mean"
     """Loss statistic to monitor. Should be one of 'mean', 'variance', 'sigma'."""
+    patience: Union[int, float] = 0
+    """Number of epochs with invalid loss after which training will be stopped."""
+
+    def __post_init__(self):
+        self._last_valid_iter = 0
 
     def __call__(self, step, log_data, driver):
         """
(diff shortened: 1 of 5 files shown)


netket/callbacks/invalidlossstopping.py
```diff
@@ -0,0 +1,52 @@
...
+        Args:
+            step: An integer corresponding to the step (iteration or epoch) in training.
+            log_data: A dictionary containing log data for training.
+            driver: A NetKet variational driver.
+
+        Returns:
+            A boolean. If True, training continues, else, it does not.
+        """
+        loss = np.real(getattr(log_data[driver._loss_name], self.monitor))
+        if np.isnan(loss):
```
femtobit, line 47:
```suggestion
        if np.isnan(loss):
```
I wonder if `np.isfinite` would be better? It tests for `NaN`, but also for `±Infinity`, which also seems like a problematic expectation that might warrant aborting the simulation.

What do you think? @VolodyaCO @PhilipVinc

VolodyaCO: Sure thing!
```suggestion
        if np.isfinite(loss):
```

netket/callbacks/invalidlossstopping.py
```diff
@@ -0,0 +1,52 @@
...
+        Args:
+            step: An integer corresponding to the step (iteration or epoch) in training.
+            log_data: A dictionary containing log data for training.
+            driver: A NetKet variational driver.
+
+        Returns:
+            A boolean. If True, training continues, else, it does not.
+        """
+        loss = np.real(getattr(log_data[driver._loss_name], self.monitor))
+        if np.isfinite(loss):
```
femtobit, line 47:
```suggestion
        if not np.isfinite(loss):
```

femtobit (approved):
Looks good to me now. Thanks @VolodyaCO.

PR #1352, opened 2022-10-09
Add invariant and constant diagonal shift at the same time in `QGTJacobian*`

attila-i-szabo:
This PR allows specifying both a scale-invariant and a constant diagonal offset in `QGTJacobian*` at the same time.

@chrisrothUT and I discovered that adding a diagonal shift of the form `diag_shift + diag_scale * S_ii` is much more stable than pure scale-invariant shifting but remains faster and better converging than simply using `diag_shift`. To accommodate both, this PR deprecates `rescale_shift` and introduces `diag_scale`, the coefficient of S_ii in the shift.

Defaults:
* If nothing is specified, `diag_shift=0.01, diag_scale=0.0` to recover the original behaviour
* If only `diag_scale` is specified, `diag_shift=0.0`
* If `rescale_shift` is specified, it behaves as it used to, but there is a deprecation warning
* Specifying `rescale_shift` and `diag_scale` together leads to an error

Internally, `diag_scale` is implemented the same way `rescale_shift` once was; `diag_shift` is added to it by adding `offset=diag_shift/diag_scale` to the scale factors used to rescale rows/columns of the S matrix. 

There are some bits to iron out, most importantly whether the defaults above are good and `diag_scale` is a good name for the new parameter.

### DIFF
--- a/netket/optimizer/qgt/qgt_jacobian_common.py
+++ b/netket/optimizer/qgt/qgt_jacobian_common.py
@@ -115,8 +115,13 @@ def choose_jacobian_mode(afun, pars, state, samples, *, mode, holomorphic):
 def sanitize_diag_shift(diag_shift, diag_scale, rescale_shift):
     """Sanitises different inputs for diag_shift etc.
 
-    Output: diag_shift, diag_scale.
-    Raises deprecation warnings for `rescale_shift`."""
+
+    Also raises a deprecation warnings for `rescale_shift`.
+
+    Returns:
+        the tuple `(diag_shift, diag_scale)`.
+    """
+
     if diag_shift is None:
         diag_shift = 0.01 if diag_scale is None else 0.0
 
--- a/netket/optimizer/qgt/qgt_jacobian_common.py
+++ b/netket/optimizer/qgt/qgt_jacobian_common.py
@@ -173,7 +173,7 @@ def rescale(centered_oks, offset, *, ndims: int = 1):
 
     scale = jax.tree_map(
         lambda x: (
-            mpi.mpi_sum_jax(jnp.sum((x * x.conj()).real, axis=0, keepdims=True))[0]
+            mpi.mpi_sum_jax(jnp.sum((x * x.conj()).real, axis=axis, keepdims=True))[0]
             + offset
         )
         ** 0.5,
--- a/netket/optimizer/qgt/qgt_jacobian_common.py
+++ b/netket/optimizer/qgt/qgt_jacobian_common.py
@@ -120,7 +120,7 @@ def sanitize_diag_shift(diag_shift, diag_scale, rescale_shift):
     if diag_shift is None:
         diag_shift = 0.01 if diag_scale is None else 0.0
 
-    if rescale_shift == False:
+    if rescale_shift is False:
         warn_deprecation(
             "`rescale_shift` is deprecated, please do not specify `rescale_shift=False`."
         )
@@ -130,7 +130,7 @@ def sanitize_diag_shift(diag_shift, diag_scale, rescale_shift):
             )
 
         return diag_shift, 0.0
-    elif rescale_shift == True:
+    elif rescale_shift is True:
         warn_deprecation(
             f"`rescale_shift` is deprecated, use `diag_scale={diag_shift}, diag_shift=0` instead."
         )
--- a/netket/optimizer/qgt/qgt_jacobian_common.py
+++ b/netket/optimizer/qgt/qgt_jacobian_common.py
@@ -17,8 +17,10 @@ import warnings
 from textwrap import dedent
 
 import jax
+import jax.numpy as jnp
 
 import netket.jax as nkjax
+from netket.utils import mpi
 
 
 @partial(jax.jit, static_argnums=(0, 4, 5))
@@ -107,3 +109,31 @@ def choose_jacobian_mode(afun, pars, state, samples, *, mode, holomorphic):
         return "holomorphic"
     else:
         raise ValueError(f"unknown mode {i}")
+
+
+def rescale(centered_oks, *, ndims: int = 1):
+    """
+    compute ΔOₖ/√Sₖₖ and √Sₖₖ
+    to do scale-invariant regularization (Becca & Sorella 2017, pp. 143)
+    Sₖₗ/(√Sₖₖ√Sₗₗ) = ΔOₖᴴΔOₗ/(√Sₖₖ√Sₗₗ) = (ΔOₖ/√Sₖₖ)ᴴ(ΔOₗ/√Sₗₗ)
+
+    Args:
+        centered_oks: A mean-zero Jacobian.
+        ndims: A number of leading dimensions to use to compute the
+            rescale factor. Those should be all the non-parameters
+            axes in the jacobian (so it should be 1 normally, 2 for
+            non holomorphic stacked jacobians).
+    """
+    # should be (0,) for standard, (0,1) when we have 2 jacobians in complex mode
+    axis = tuple(range(ndims))
+
+    scale = jax.tree_map(
+        lambda x: mpi.mpi_sum_jax(
+            jnp.sum((x * x.conj()).real, axis=axis, keepdims=True)
+        )[0]
+        ** 0.5,
+        centered_oks,
+    )
+    centered_oks = jax.tree_map(jnp.divide, centered_oks, scale)
+    scale = jax.tree_map(partial(jnp.squeeze, axis=axis), scale)
+    return centered_oks, scale
--- a/netket/optimizer/qgt/qgt_jacobian_common.py
+++ b/netket/optimizer/qgt/qgt_jacobian_common.py
@@ -21,6 +21,7 @@ import jax.numpy as jnp
 
 import netket.jax as nkjax
 from netket.utils import mpi
+from netket.utils import warn_deprecation
 
 
 @partial(jax.jit, static_argnums=(0, 4, 5))
@@ -111,7 +112,50 @@ def choose_jacobian_mode(afun, pars, state, samples, *, mode, holomorphic):
         raise ValueError(f"unknown mode {i}")
 
 
-def rescale(centered_oks, *, ndims: int = 1):
+def sanitize_diag_shift(diag_shift, diag_scale, rescale_shift):
+    """Sanitises different inputs for diag_shift etc.
+
+    Output: diag_shift, diag_scale.
+    Raises deprecation warnings for `rescale_shift`."""
+    if diag_shift is None:
+        diag_shift = 0.01 if diag_scale is None else 0.0
+
+    if rescale_shift == False:
+        warn_deprecation(
+            "`rescale_shift` is deprecated, please do not specify `rescale_shift=False`."
+        )
+        if diag_scale is not None:
+            raise ValueError(
+                "`rescale_shift` and `diag_scale` must not be specified together."
+            )
+
+        return diag_shift, 0.0
+    elif rescale_shift == True:
+        warn_deprecation(
+            f"`rescale_shift` is deprecated, use `diag_scale={diag_shift}, diag_shift=0` instead."
+        )
+        if diag_scale is not None:
+            raise ValueError(
+                "`rescale_shift` and `diag_scale` must not be specified together."
+            )
+
+        return 0.0, diag_shift
+    elif rescale_shift is None:
+        if diag_scale is None:
+            diag_scale = 0.0
+        return diag_shift, diag_scale
+    else:
+        raise ValueError("`rescale_shift` must be boolean or None.")
+
+
+def to_shift_offset(diag_shift, diag_scale):
+    if diag_scale == 0.0:
+        return diag_shift, None
+    else:
+        return diag_scale, diag_shift / diag_scale
+
+
+def rescale(centered_oks, offset, *, ndims: int = 1):
     """
     compute ΔOₖ/√Sₖₖ and √Sₖₖ
     to do scale-invariant regularization (Becca & Sorella 2017, pp. 143)
@@ -128,9 +172,10 @@ def rescale(centered_oks, *, ndims: int = 1):
     axis = tuple(range(ndims))
 
     scale = jax.tree_map(
-        lambda x: mpi.mpi_sum_jax(
-            jnp.sum((x * x.conj()).real, axis=axis, keepdims=True)
-        )[0]
+        lambda x: (
+            mpi.mpi_sum_jax(jnp.sum((x * x.conj()).real, axis=0, keepdims=True))[0]
+            + offset
+        )
         ** 0.5,
         centered_oks,
     )
--- a/netket/optimizer/qgt/qgt_jacobian_dense.py
+++ b/netket/optimizer/qgt/qgt_jacobian_dense.py
@@ -81,11 +81,12 @@ def QGTJacobianDense(
     :math:`\\epsilon_{1,2}` are specified using `diag_scale` and `diag_shift`,
     respectively. By default, `diag_scale=0.0`, `diag_shift=0.01`.
     """
+    if mode is not None and holomorphic is not None:
+        raise ValueError("Cannot specify both `mode` and `holomorphic`.")
+    if rescale_shift is not None and diag_scale is not None:
+        raise ValueError("Cannot specify both `rescale_shift` and `diag_scale`.")
+
     if vstate is None:
-        if mode is not None and holomorphic is not None:
-            raise ValueError("Cannot specify both `mode` and `holomorphic`.")
-        if rescale_shift is not None and diag_scale is not None:
-            raise ValueError("Cannot specify both `rescale_shift` and `diag_scale`.")
         if diag_shift is not None:
             kwargs["diag_shift"] = diag_shift
         if diag_scale is not None:
--- a/netket/optimizer/qgt/qgt_jacobian_dense.py
+++ b/netket/optimizer/qgt/qgt_jacobian_dense.py
@@ -87,17 +87,14 @@ def QGTJacobianDense(
         raise ValueError("Cannot specify both `rescale_shift` and `diag_scale`.")
 
     if vstate is None:
-        if diag_shift is not None:
-            kwargs["diag_shift"] = diag_shift
-        if diag_scale is not None:
-            kwargs["diag_scale"] = diag_scale
-        if rescale_shift is not None:
-            kwargs["rescale_shift"] = rescale_shift
         return partial(
             QGTJacobianDense,
             mode=mode,
             holomorphic=holomorphic,
             chunk_size=chunk_size,
+            diag_shift=diag_shift,
+            diag_scale=diag_scale,
+            rescale_shift=rescale_shift,
             **kwargs,
         )
 
--- a/netket/optimizer/qgt/qgt_jacobian_dense.py
+++ b/netket/optimizer/qgt/qgt_jacobian_dense.py
@@ -69,6 +69,9 @@ def QGTJacobianDense(
                     (useful for models where the backward pass requires more
                     memory than the forward pass).
     """
+    if mode is not None and holomorphic is not None:
+        raise ValueError("Cannot specify both `mode` and `holomorphic`.")
+
     diag_shift, diag_scale = sanitize_diag_shift(diag_shift, diag_scale, rescale_shift)
 
     if vstate is None:
@@ -76,8 +79,9 @@ def QGTJacobianDense(
             QGTJacobianDense,
             mode=mode,
             holomorphic=holomorphic,
-            rescale_shift=rescale_shift,
             chunk_size=chunk_size,
+            diag_shift=diag_shift,
+            diag_scale=diag_scale,
             **kwargs,
         )
 
@@ -96,13 +100,11 @@ def QGTJacobianDense(
             mode=mode,
             holomorphic=holomorphic,
         )
-    elif holomorphic is not None:
-        raise ValueError("Cannot specify both `mode` and `holomorphic`.")
-
-    shift, offset = to_shift_offset(diag_shift, diag_scale)
 
     if chunk_size is None and hasattr(vstate, "chunk_size"):
         chunk_size = vstate.chunk_size
+
+    shift, offset = to_shift_offset(diag_shift, diag_scale)
 
     O, scale = prepare_centered_oks(
         vstate._apply_fun,
--- a/netket/optimizer/qgt/qgt_jacobian_dense.py
+++ b/netket/optimizer/qgt/qgt_jacobian_dense.py
@@ -26,7 +26,7 @@ import netket.jax as nkjax
 from ..linear_operator import LinearOperator, Uninitialized
 
 from .common import check_valid_vector_type
-from .qgt_jacobian_dense_logic import prepare_centered_oks, vec_to_real
+from .qgt_jacobian_dense_logic import prepare_centered_oks, vec_to_real, mat_vec
 from .qgt_jacobian_common import choose_jacobian_mode
 
 
@@ -198,10 +198,7 @@ def _matmul(
     if self.scale is not None:
         vec = vec * self.scale
 
-    result = (
-        mpi.mpi_sum_jax(((self.O @ vec).T.conj() @ self.O).T.conj())[0]
-        + self.diag_shift * vec
-    )
+    result = mat_vec(vec, self.O, self.diag_shift)
 
     if self.scale is not None:
         result = result * self.scale
@@ -259,9 +256,11 @@ def _solve(
 def _to_dense(self: QGTJacobianDenseT) -> jnp.ndarray:
     if self.scale is None:
         O = self.O
-        diag = jnp.eye(self.O.shape[1])
+        diag = jnp.eye(self.O.shape[-1])
     else:
         O = self.O * self.scale[jnp.newaxis, :]
         diag = jnp.diag(self.scale**2)
 
-    return mpi.mpi_sum_jax(O.T.conj() @ O)[0] + self.diag_shift * diag
+    # concatenate samples with real/Imaginary dimension
+    O = O.reshape(-1, O.shape[-1])
+    return mpi.mpi_sum_jax(O.conj().T @ O)[0] + self.diag_shift * diag
--- a/netket/optimizer/qgt/qgt_jacobian_dense.py
+++ b/netket/optimizer/qgt/qgt_jacobian_dense.py
@@ -55,21 +55,6 @@ def QGTJacobianDense(
     The details on how the ⟨S⟩⁻¹⟨F⟩ system is solved are contained in
     the field `sr`.
 
-    Args:
-        vstate: The variational state
-        mode: "real", "complex" or "holomorphic": specifies the implementation
-              used to compute the jacobian. "real" discards the imaginary part
-              of the output of the model. "complex" splits the real and imaginary
-              part of the parameters and output. It works also for non holomorphic
-              models. holomorphic works for any function assuming it's holomorphic
-              or real valued.
-        holomorphic: a flag to indicate that the function is holomorphic.
-        diag_scale: Fractional shift :math:`\\epsilon_1` added to diagonal entries (see below)
-        diag_shift: Constant shift :math:`\\epsilon_2` added to diagonal entries (see below)
-        chunk_size: If supplied, overrides the chunk size of the variational state
-                    (useful for models where the backward pass requires more
-                    memory than the forward pass).
-
     Numerical estimates of the QGT are usually ill-conditioned and require
     regularisation. The standard approach is to add a positive constant to the diagonal;
     alternatively, Becca and Sorella (2017) propose scaling this offset with the
@@ -81,6 +66,21 @@ def QGTJacobianDense(
 
     :math:`\\epsilon_{1,2}` are specified using `diag_scale` and `diag_shift`,
     respectively. By default, `diag_scale=0.0`, `diag_shift=0.01`.
+
+    Args:
+        vstate: The variational state
+        mode: "real", "complex" or "holomorphic": specifies the implementation
+              used to compute the jacobian. "real" discards the imaginary part
+              of the output of the model. "complex" splits the real and imaginary
+              part of the parameters and output. It works also for non holomorphic
+              models. holomorphic works for any function assuming it's holomorphic
+              or real valued.
+        holomorphic: a flag to indicate that the function is holomorphic.
+        diag_scale: Fractional shift :math:`\\epsilon_1` added to diagonal entries (see above)
+        diag_shift: Constant shift :math:`\\epsilon_2` added to diagonal entries (see above)
+        chunk_size: If supplied, overrides the chunk size of the variational state
+                    (useful for models where the backward pass requires more
+                    memory than the forward pass).
     """
     if mode is not None and holomorphic is not None:
         raise ValueError("Cannot specify both `mode` and `holomorphic`.")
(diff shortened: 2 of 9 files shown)


netket/optimizer/qgt/qgt_jacobian_dense_logic.py
```diff
@@ -151,21 +151,21 @@ def _rescale(centered_oks):
...
-    )
+        + offset
+    ) ** 0.5
     centered_oks = jnp.divide(centered_oks, scale)
     scale = jnp.squeeze(scale, axis=0)
     return centered_oks, scale
 
 
-@partial(jax.jit, static_argnames=("apply_fun", "mode", "rescale_shift", "chunk_size"))
+@partial(jax.jit, static_argnames=("apply_fun", "mode", "chunk_size"))
```
femtobit, line 161: Shouldn't `offset` be static now?

netket/optimizer/qgt/qgt_jacobian_pytree_logic.py
```diff
@@ -211,14 +212,14 @@ def _mat_vec(v: PyTree, oks: PyTree) -> PyTree:
 # here the other modes are converted
 
 
-@partial(jax.jit, static_argnames=("apply_fun", "mode", "rescale_shift", "chunk_size"))
+@partial(jax.jit, static_argnames=("apply_fun", "mode", "chunk_size"))
```
femtobit, line 215: (see above)

netket/optimizer/qgt/qgt_jacobian_dense_logic.py
attila-i-szabo: No. It jits the same way if it's any `float`, and `None` generates a new jit by type.

attila-i-szabo: Though it makes little difference, it would usually be the same number throughout...

netket/optimizer/qgt/qgt_jacobian_common.py
```diff
@@ -107,3 +108,46 @@ def choose_jacobian_mode(afun, pars, state, samples, *, mode, holomorphic):
...
+
+def sanitize_diag_shift(diag_shift, diag_scale, rescale_shift):
+    """Sanitises different inputs for diag_shift etc.
+
+    Output: diag_shift, diag_scale.
+    Raises deprecation warnings for `rescale_shift`."""
+    if diag_shift is None:
+        diag_shift = 0.01 if diag_scale is None else 0.0
+
+    if rescale_shift == False:
```
attila-i-szabo, line 121:
```suggestion
    if rescale_shift is False:
```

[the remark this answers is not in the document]
attila-i-szabo: for some reason it didn't show up on the changed files page so I duplicated it...

test/optimizer/test_qgt_logic.py
```diff
@@ -391,11 +403,39 @@ def test_scale_invariant_regularization(e, outdtype, pardtype):
...
+        ((None, 0.02, None), (0.0, 0.02)),
+        ((0.01, 0.02, None), (0.01, 0.02)),
+        ((None, None, False), (0.01, 0.0)),
+        ((0.02, None, False), (0.02, 0.0)),
+        ((None, 0.02, False), "error"),
+        ((0.01, 0.02, False), "error"),
+        ((None, None, True), (0.0, 0.01)),
+        ((0.02, None, True), (0.0, 0.02)),
+        ((None, 0.02, True), "error"),
+        ((0.01, 0.02, True), "error"),
```
attila-i-szabo, line 429:
```suggestion
        ((0.01, 0.02, True), "error"),
        ((None, None, 0.0), "error"),
```
because codecov was complaining

PhilipVinc (changes requested):
Hi Attila, thanks for the contribution!

I'm all in favour of taking in this change, but as it stands it's an under-documented feature that is not easily discoverable and I doubt non-core users would easily end up using it.
I think solving this issue would be simple:

- Extensively extend the docstrings of the `QGTJacobian**` constructors by writing down the mathematical formula of the resulting geometric tensor (preferably using latex, so it gets rendered on the API reference) so that those flags are properly documented. Also, you can add brief discussions or references to literature (if any) about in what situations this might help.
 
- Compile a brief python notebook on the subject of "improving convergence with QGT rescalings" where you take a simple model (Ising, Heisenberg... anything that run fast) and show that one rescaling converges faster than another. For example, you could compare the 3, "pure shift", "pure rescaling", "combined".

I know it's less exciting than solving our research needs, but we can't just throw features around without documenting them, because after all the greatest selling point of NetKet is the fact that it is documented and non-super-tech-savy users can use it.

attila-i-szabo:
> I think solving this issue would be simple:

Well, it requires people to start using it, which is hard even if you show it's better.

> * Extensively extend the docstrings of the `QGTJacobian**` constructors by writing down the mathematical formula of the resulting geometric tensor (preferably using latex, so it gets rendered on the API reference) so that those flags are properly documented. Also, you can add brief discussions or references to literature (if any) about in what situations this might help.

How do you put latex in the docstring?

> * Compile a brief python notebook on the subject of "improving convergence with QGT rescalings" where you take a simple model (Ising, Heisenberg... anything that run fast) and show that one rescaling converges faster than another. For example, you could compare the 3, "pure shift", "pure rescaling", "combined".

I won't have time for this anytime soon, so it might be better to merge without this and come back to it later. When I get around it, where would this go? Is there already an SR page somewhere in the docs?

PhilipVinc:
@attila-i-szabo I rebased this on master, as I had landed a lot of changes in order to unify the two jacobian implementations.
I changed a bit your tests of the API to be a bit less likely to fail if we add keyword arguments in the future.

If you could expand the docstrings of the QGTJacobian, I will merge this.

attila-i-szabo:
Thanks for this. I'll do the docstrings asap.

However, I was having second thoughts about doing the sanitisation early. Basically, in the upstream code now, you can say `qgt= QGTJacobianDense(...)` and then change `diag_shift` as a function of where you are in the training (SR and the standard drivers don't support it, but you can always write your own training loop). We'd lose this functionality if we run `sanitize_diag_shift` before returning the `partial`. (We could check for not specifying `diag_scale` and `rescale_shift` the way you've just introduced.)

That is, I'd rewrite the start of the constructor as
```python
    if mode is not None and holomorphic is not None:
        raise ValueError("Cannot specify both `mode` and `holomorphic`.")

    if vstate is None:
        return partial(
            QGTJacobianPyTree,
            mode=mode,
            holomorphic=holomorphic,
            chunk_size=chunk_size,
            diag_shift=diag_shift,
            diag_scale=diag_scale,
            rescale_shift=rescale_shift,
            **kwargs,
        )

    diag_shift, diag_scale = sanitize_diag_shift(diag_shift, diag_scale, rescale_shift)
```

The only downside (unless I'm missing something) is that this will generate a deprecation warning on every iteration, as the sanitisation is now inside the training loop. Is that OK? Or, how can we stop that?

PhilipVinc:
I tried not to change your logics I think I left the sanitisation where you had left it? But maybe I'm mistaken...

Hmm... I understand what you mean but.. how would putting the sanitisation after the partial change anything? You can't modify the partial object (it's a tuple inside) so to change the qgt in the standard driver you'd have to do something like

```python
gs = VMC(..., preconditioner=SR())
gs.preconditioner.qgt = QGTJacobianDense(diag_hisft=0.4)
```

anyway... or am I missing something?

attila-i-szabo:
Oh yeah, you're right, this still imposes some choices of `diag_shift` etc on the `partial`. Do you think it's worth excluding the `None`s from the `partial` like
```python
    if vstate is None:
        if diag_shift is not None:
            kwargs['diag_shift']=diag_shift
        # same for diag_scale and rescale_shift
        return partial(
            QGTJacobianPyTree,
            mode=mode,
            holomorphic=holomorphic,
            chunk_size=chunk_size,
            **kwargs,
        )
```

> I tried not to change your logics. I think I left the sanitisation where you had left it? But maybe I'm mistaken...

Yeah, this is something I want to change

PhilipVinc:
Hmm... can you tell me how you'd like to use this together with the driver? Maybe if we look at this from an higher level it will be more clear.

attila-i-szabo:
as I said you can't use this with the existing drivers (I wanted to add support, but we never got around it), but you can write something like

```python
for i in range(1000):
 _, grad = vstate.expect_and_grad()
 update, _ = qgt(vstate, diag_shift = f(i)).solve(cg, grad)
 params -= rate * update # you know what I mean
```

of course at this point you might as well write the full QGT definition inside the loop, but then what do we have this functionality for? (SR makes its own `partial`s)

PhilipVinc:
The main design point for me is to keep all this regularisation scheduling logic outside `QGT***` because I think those objects should not interact with it. They are 'frozen' objects.

The partial generation is an user-friendly thing so to avoid having them use partials all over the place, but I would like to eventually remove it in favour of something better/not put too much into it.

netket/optimizer/qgt/qgt_jacobian_dense.py
```diff
@@ -63,28 +63,45 @@ def QGTJacobianDense(
...
+    .. math::
+
+        S_{ii} \mapsto S_{ii} + \epsilon_1 S_{ii} + \epsilon_2;
 
+    :math:`\epsilon_{1,2}` are specified using `diag_scale` and `diag_shift`,
+    respectively. By default, `diag_scale=0.0`, `diag_shift=0.01`.
+    """
     if vstate is None:
+        if mode is not None and holomorphic is not None:
+            raise ValueError("Cannot specify both `mode` and `holomorphic`.")
```
PhilipVinc, line 86: this was good where it was?

netket/optimizer/qgt/qgt_jacobian_dense.py
```diff
@@ -63,28 +63,45 @@ def QGTJacobianDense(
...
+        if rescale_shift is not None and diag_scale is not None:
+            raise ValueError("Cannot specify both `rescale_shift` and `diag_scale`.")
+        if diag_shift is not None:
+            kwargs["diag_shift"] = diag_shift
+        if diag_scale is not None:
+            kwargs["diag_scale"] = diag_scale
+        if rescale_shift is not None:
+            kwargs["rescale_shift"] = rescale_shift
```
PhilipVinc, line 94: I'm not necessarily against this (though it's a lot of logic) but I fail to see how this helps us down the line..

attila-i-szabo: see [here](https://github.com/netket/netket/pull/1352#issuecomment-1281473410)

netket/optimizer/qgt/qgt_jacobian_dense.py
PhilipVinc: This way it will never be executed if you run `QGTJacobianDense(vstate, holomorphic=True, mode="complex")` but it should error.

Plus, I think it's better to error early if we can rather than later.

PhilipVinc: Early because if it errors right when you specified those options it's more obvious to an user what was wrong, rather than erroring later on in the optimisation and the user has to understand where he had specified those options.

attila-i-szabo: > This way it will never be executed if you run `QGTJacobianDense(vstate, holomorphic=True, mode="complex")` but it should error.

OK, I was assuming it errors somewhere down the line, but it doesn't...

PhilipVinc: I removed it in my commit when I moved it to the top. I noticed that we could have errored early and thought it was better.

netket/optimizer/qgt/qgt_jacobian_dense.py
PhilipVinc: I think this might be unneeded... See the following:

``python
In [5]: def f(**kwargs):
   ...:      print(kwargs)
   ...:
In [8]: partial(f, ciao=1)
Out[8]: functools.partial(<function f at 0x108665630>, ciao=1)
In [9]: partial(f, ciao=1)(ciao=3)
{'ciao': 3}
```

So you can shadow already defined partial warts by passing them from the top level.
Therefore I think it would be ok to drop this chunk of `if is not None...` and simply chug all keyword arguments in the partial 
```suggestion
```

PhilipVinc: this works because we now sanitise the arguments after the partial, not before, which I think is the correct thing to do.

netket/optimizer/qgt/qgt_jacobian_common.py
```diff
@@ -111,7 +112,50 @@ def choose_jacobian_mode(afun, pars, state, samples, *, mode, holomorphic):
         raise ValueError(f"unknown mode {i}")
 
 
-def rescale(centered_oks, *, ndims: int = 1):
+def sanitize_diag_shift(diag_shift, diag_scale, rescale_shift):
+    """Sanitises different inputs for diag_shift etc.
+
+    Output: diag_shift, diag_scale.
+    Raises deprecation warnings for `rescale_shift`."""
```
PhilipVinc, line 119:
```suggestion

    Also raises a deprecation warnings for `rescale_shift`.

    Returns:
        the tuple `(diag_shift, diag_scale)`.
    """

```

netket/optimizer/qgt/qgt_jacobian_dense.py
attila-i-szabo: > So you can shadow already defined partial warts by passing them from the top level.

OK that's interesting

netket/optimizer/qgt/qgt_jacobian_common.py
```diff
@@ -111,7 +112,55 @@ def choose_jacobian_mode(afun, pars, state, samples, *, mode, holomorphic):
...
+        if diag_scale is not None:
+            raise ValueError(
+                "`rescale_shift` and `diag_scale` must not be specified together."
+            )
+
+        return 0.0, diag_shift
+    elif rescale_shift is None:
+        if diag_scale is None:
+            diag_scale = 0.0
+        return diag_shift, diag_scale
```
PhilipVinc, line 151: can we keep `diag_scale=None` the default? Using 0 here is conceptually wrong, I think.

And then we change the function below as well

```suggestion
        return diag_shift, diag_scale
```

netket/optimizer/qgt/qgt_jacobian_common.py
```diff
@@ -111,7 +112,55 @@ def choose_jacobian_mode(afun, pars, state, samples, *, mode, holomorphic):
...
+    if rescale_shift is False:
+        warn_deprecation(
+            "`rescale_shift` is deprecated, please do not specify `rescale_shift=False`."
+        )
+        if diag_scale is not None:
+            raise ValueError(
+                "`rescale_shift` and `diag_scale` must not be specified together."
+            )
+
+        return diag_shift, 0.0
```
PhilipVinc, line 137:
```suggestion
        return diag_shift, None
```

netket/optimizer/qgt/qgt_jacobian_common.py
```diff
@@ -111,7 +112,55 @@ def choose_jacobian_mode(afun, pars, state, samples, *, mode, holomorphic):
...
+        return diag_shift, diag_scale
+    else:
+        raise ValueError("`rescale_shift` must be boolean or None.")
+
+
+def to_shift_offset(diag_shift, diag_scale):
+    if diag_scale == 0.0:
+        return diag_shift, None
+    else:
+        return diag_scale, diag_shift / diag_scale
```
PhilipVinc, line 160:
```suggestion
def to_shift_offset(diag_shift, diag_scale):
    if diag_scale is not None:
        return diag_scale, diag_shift / diag_scale
    return diag_shift, diag_scale
```

attila-i-szabo:
No, at this point `diag_shift` and `diag_scale` should be the actual coefficients in the diagonal shift. I think it makes the code less transparent if we use different rules for `diag_shift` and `diag_scale`

PhilipVinc:
But a scale of 0 makes no sense

attila-i-szabo:
Makes the same amount of sense as `diag_shift=0.0` but we keep that (and we should, it makes things veyr messy if we make that `None` too)

attila-i-szabo:
Could we merge this or is there anything more to be done?

netket/optimizer/qgt/qgt_jacobian_dense.py
```diff
@@ -58,22 +64,43 @@ def QGTJacobianDense(
...
               or real valued.
         holomorphic: a flag to indicate that the function is holomorphic.
-        rescale_shift: If True rescales the diagonal shift.
+        diag_scale: Fractional shift :math:`\\epsilon_1` added to diagonal entries (see below)
+        diag_shift: Constant shift :math:`\\epsilon_2` added to diagonal entries (see below)
         chunk_size: If supplied, overrides the chunk size of the variational state
                     (useful for models where the backward pass requires more
                     memory than the forward pass).
+
+    Numerical estimates of the QGT are usually ill-conditioned and require
```
PhilipVinc, line 73: You should move this block above/before the args

PhilipVinc: also on the other file

PhilipVinc:
If you fix the formatting of the docstring I can merge.

attila-i-szabo:
Done. Also added changelog.

PhilipVinc:
lovely! let me check if it renders properly as soon as the docs/readthedocs.org:netket worker is done and I can merge.

By the way, @inailuig , is it pobbisble to implement `diag_scale` for `QGTOntheFly` as well? Possible in the sense of good performance.
And how hard would it be? Hard in the sense of keeping the complexity of the code low.

netket/optimizer/qgt/qgt_jacobian_dense.py
```diff
@@ -49,6 +55,18 @@ def QGTJacobianDense(
...
+    regularisation. The standard approach is to add a positive constant to the diagonal;
+    alternatively, Becca and Sorella (2017) propose scaling this offset with the
+    diagonal entry itself. NetKet allows using both in tandem:
+
+    .. math::
+
+        S_{ii} \\mapsto S_{ii} + \\epsilon_1 S_{ii} + \\epsilon_2;
+
+    :math:`\\epsilon_{1,2}` are specified using `diag_scale` and `diag_shift`,
+    respectively. By default, `diag_scale=0.0`, `diag_shift=0.01`.
```
PhilipVinc, line 68:
```suggestion
    :math:`\\epsilon_{1,2}` are specified using `diag_scale` and `diag_shift`,
    respectively.
```

those are only the defaults for SR, not the real defaults of this

netket/optimizer/qgt/qgt_jacobian_dense.py
```diff
@@ -58,22 +76,31 @@ def QGTJacobianDense(
               models. holomorphic works for any function assuming it's holomorphic
               or real valued.
         holomorphic: a flag to indicate that the function is holomorphic.
-        rescale_shift: If True rescales the diagonal shift.
+        diag_scale: Fractional shift :math:`\\epsilon_1` added to diagonal entries (see above)
+        diag_shift: Constant shift :math:`\\epsilon_2` added to diagonal entries (see above)
```
PhilipVinc, line 80:
```suggestion
        diag_scale: Fractional shift :math:`\\epsilon_1` added to diagonal entries (see above).
        diag_shift: Constant shift :math:`\\epsilon_2` added to diagonal entries (see above).
```

netket/optimizer/qgt/qgt_jacobian_pytree.py
```diff
@@ -49,6 +55,18 @@ def QGTJacobianPyTree(
...
+    regularisation. The standard approach is to add a positive constant to the diagonal;
+    alternatively, Becca and Sorella (2017) propose scaling this offset with the
+    diagonal entry itself. NetKet allows using both in tandem:
+
+    .. math::
+
+        S_{ii} \\mapsto S_{ii} + \\epsilon_1 S_{ii} + \\epsilon_2;
+
+    :math:`\\epsilon_{1,2}` are specified using `diag_scale` and `diag_shift`,
+    respectively. By default, `diag_scale=0.0`, `diag_shift=0.01`.
```
PhilipVinc, line 68:
```suggestion
    respectively.
```

netket/optimizer/qgt/qgt_jacobian_pytree.py
```diff
@@ -58,21 +76,31 @@ def QGTJacobianPyTree(
               models. holomorphic works for any function assuming it's holomorphic
               or real valued.
         holomorphic: a flag to indicate that the function is holomorphic.
-        rescale_shift: If True rescales the diagonal shift.
+        diag_scale: Fractional shift :math:`\\epsilon_1` added to diagonal entries (see above)
+        diag_shift: Constant shift :math:`\\epsilon_2` added to diagonal entries (see above)
```
PhilipVinc, line 80:
```suggestion
        diag_scale: Fractional shift :math:`\\epsilon_1` added to diagonal entries (see above).
        diag_shift: Constant shift :math:`\\epsilon_2` added to diagonal entries (see above).
```

PhilipVinc (approved):

PhilipVinc:
thanks!

attila-i-szabo:
> those are only the defaults for SR, not the real defaults of this

They actually are atm. I thought that was part of QGT, so I included that in `sanitize_diag_shift`. Should we change that?

PhilipVinc:
Ah wow. Yes. Yes. The default should be zero for both!

attila-i-szabo:
Okay. Should we revert, fix and remerge?

PhilipVinc:
By the way, I looked again at the code and noticed that you also output, in `to shift offset`, a None for diag scale (which makes sense because we can then avoid a little bit of computations then in jax).

So right now the flow is that `sanitize` converts a `scale=None` to a 0.0. Then `to shift offset` converts 0.0 to None again.

I would like to have sanitize keep a None None so that we avoid potential problems with floating point 0s as long as users do not specify a value for it.

Another thing that makes me think that this approach is better is that it would be entirely jittable, contrarily to converting 0s to None which would not (and a long term goal I have would be to jit much more)

attila-i-szabo:
> None for diag scale (which makes sense because we can then avoid a little bit of computations then in jax).

it's more important than that. None triggers a new jit, which you need because the logic is different if you have no `diag_scale` at all

> So right now the flow is that `sanitize` converts a `scale=None` to a 0.0. Then `to shift offset` converts 0.0 to None again.
> 
> I would like to have sanitize keep a None None so that we avoid potential problems with floating point 0s as long as users do not specify a value for it.

`diag_scale=0` is also a legitimate user input and I don't see any reason to treat it separate from `diag_scale=None`. Also, these are zero literals, which are reliable (floating point numbers are inaccurate, not nuts):
```python
>>> 0.0 == 0
True
>>> 0.0 == -0.0
True
>>> -0.0 == 0
True
>>> -0.0e27 == 0
True
```

> Another thing that makes me think that this approach is better is that it would be entirely jittable, contrarily to converting 0s to None which would not (and a long term goal I have would be to jit much more)

All this conversion logic takes maybe 100 clock cycles and is full of conditionals, so nothing to be jitted. The actual computation is already fully jitted (see above re why `to_shift_offset` produces `None`)

PhilipVinc:
Open another PR, it's easier

netket/optimizer/qgt/qgt_jacobian_common.py
```diff
@@ -111,7 +112,55 @@ def choose_jacobian_mode(afun, pars, state, samples, *, mode, holomorphic):
...
+        )
+        if diag_scale is not None:
+            raise ValueError(
+                "`rescale_shift` and `diag_scale` must not be specified together."
+            )
+
+        return 0.0, diag_shift
+    elif rescale_shift is None:
+        if diag_scale is None:
+            diag_scale = 0.0
```
PhilipVinc, line 150: I'm mainly advocating for removing those two lines

PhilipVinc:
>it's more important than that. None triggers a new jit, which you need because the logic is different if you have no diag_scale at all

Yep I agree with you.

>diag_scale=0 is also a legitimate user input and I don't see any reason to treat it separate from diag_scale=None. Also, these are zero literals, which are reliable (floating point numbers are inaccurate, not nuts):

And I agree also with this.

netket/optimizer/qgt/qgt_jacobian_common.py
attila-i-szabo: that's the opposite of `sanitizing` the inputs. currently, `sanitize...` turns any input into two numbers, so `to_scale_offset` doesn't have to worry about all the possibilities.

PhilipVinc:
>All this conversion logic takes maybe 100 clock cycles and is full of conditionals, so nothing to be jitted. The actual computation is already fully jitted (see above re why to_shift_offset produces None)

I also agree with that. 

The main interest in jitting the QGT constructor would be to run AD on it. I once played around with using AD to determine the best diag_shift, and yuopu need to make the code jai-friendly to do that...

What I tried was not very useful, and maybe it's useless. But being able to do something, rather than prevent it, is very important for me. We cannot know all things someone might want to do so the code must impose as few limitations as possible.

netket/optimizer/qgt/qgt_jacobian_common.py
PhilipVinc: If it was complicated to support both, I'd agree with you. But as our QGT logic already supports both cases, I don't see why an intermediate function (which would literally not even add another line) should not support it.

attila-i-szabo:
> The main interest in jitting the QGT constructor would be to run AD on it. I once played around with using AD to determine the best diag_shift, and yuopu need to make the code jai-friendly to do that...
> 
> What I tried was not very useful, and maybe it's useless. But being able to do something, rather than prevent it, is very important for me. We cannot know all things someone might want to do so the code must impose as few limitations as possible.

You can't jit `diag_scale=0` the same way as `diag_scale=0.01` because they trigger different code paths, and this doesn't change with different implementations of the intermediate functions. And if you decide whether `diag_shift` and `diag_scale` are zero or not, everything goes through the same conditional branches, so you can jit.

PhilipVinc:
Which is why the correct solution in my opinion is to do what jax does: "type-stable". If you get a 0, it's a 0. If you get a None, it's a None. Types have semantical meaning. Even more so as Jax uses them to dispatch to different compiled implementations...

Well this makes me think now. This code jitted fine before this PR. Now this doesn't.

PhilipVinc:
Technically the problem was that, before, we had a 'static' flag (`rescale_norm`). Now we don't anymore. The type of `diag_scale` can be an implicit flag.

attila-i-szabo:
> This code jitted fine before this PR. Now this doesn't.

WDYM?

attila-i-szabo:
> The type of `diag_scale` can be an implicit flag.

The constructor as a whole was NEVER jitted. `prepare_centered_oks` was and is, so the relevant implicit flag is the type of `offset`. That's why `to_scale_offset` MUST return a `None` if `diag_scale=0`

PhilipVinc:
> WDYM

I mean that, given that preamble:
```
import jax
import netket as nk

hi = nk.hilbert.Spin(s=1 / 2, N=10)

vs = nk.vqs.MCState(sa, ma, n_samples=1000, n_discard_per_chain=100)
```

The function below jitted correctly before and jits now.
``` 
@jax.jit
def fun1(pars, sampler_state, shift):
    _vs = nk.vqs.MCState(sa, ma, variables={"params":pars})
    _vs.sampler_state = sampler_state
    S = nk.optimizer.qgt.QGTJacobianDense(_vs, diag_shift=shift, mode="real")
    return S@pars

fun1(vs.parameters, vs.sampler_state, 0.01)
```

The function below jitted correctly before, but does not jit now.
```python
@jax.jit
def fun2(pars, sampler_state, shift):
    _vs = nk.vqs.MCState(sa, ma, variables={"params":pars})
    _vs.sampler_state = sampler_state
    S = nk.optimizer.qgt.QGTJacobianDense(_vs, diag_shift=shift, mode="real", rescale_shift=True)
    return S@pars

fun2(vs.parameters, vs.sampler_state, 0.01)
```

The function below jitted correctly before, but does not jit now.
```python
@jax.jit
def fun3(pars, sampler_state, shift, scale):
    _vs = nk.vqs.MCState(sa, ma, variables={"params":pars})
    _vs.sampler_state = sampler_state
    S = nk.optimizer.qgt.QGTJacobianDense(_vs, diag_shift=shift, diag_scale=scale, mode="real")
    return S@pars

fun3(vs.parameters, vs.sampler_state, 0.01, 0.02)
```
this does not jit unless one declares `scale` a `static argument`


Now, you might ask why I wanna do that. I don't really care about jitting. 
But I do care about taking gradients of the whole `fun3` (sampling included), which requires tracing.

Now, I am not myself using `diag_scale`, but see no reason to **limit** arbitrary experimentation of users when we can do it without performance loss.

PhilipVinc:
> That's why to_scale_offset MUST return a None if diag_scale=0 and I don't want you to break that. 

I understand that you might want this performance optimisation. if you use schedules, I guess? But I would be curious to know if you have one or more specific reasons for wanting it. 

The problem is that this is, from a high level point of view, at clash with having netket fully traceable/jittable. 
In my opinion, NetKet to remain relevant and grow needs to allow users to do pretty much anything on top of its structures. We impose several limitations and what one can and cannot do, and I wish for all of those to disappear through various improvements. 

Now, I don't know exactly **why** one might want to trace the whole netket. I have a few ideas (optimal control on NQS, is the easiest answer). But I want to lift all limitations we have. And the main source of non-traceability are this kind of so called _Type Instabilities_ because they can't be traced.

Now, you might not agree with me about what should be the long term goal of netket, and I would be happy to discuss about it.

But if you don't disagree, this means that the base-logic of NetKet must support this direction. Even at the cost of losing a bit of performance when `diag_scale=0` instead of `diag_scale=None` (unless I'm wrong and the two are not algebraically the same?)

Now, having a type-stable code aka, keeping None-s None-s and floats-floats **does not** mean that `float-0` cannot be optimised away, when **not tracing**, to Nones. And I am wholly in favour of this kind of performance optimisations! My whole point of view, however, is that this is a specific performance improvement, that has to built on top of a type-stable code.

attila-i-szabo:
This is not a matter of optimisation or me hating beautiful JAX things. This is about how the code works. diag_scale=0 or None or call it what you want USES DIFFERENT LOGIC to diag_scale=0.01. The two are NOT compatible and if you try to run diag_scale=0 with code jitted on diag_scale=0.01, it will generate NaNs. No matter what beautiful JAX things you want to achieve, this boundary condition is there.

Is this clear?

All I care about is that it does its intended job and doesn't shoot the user. Ie, I'm happy with it now, and I'm happy with whatever you do to it, as long as the user doesn't get NaNs for sensible parameters.

 If you want the type of diag_scale to be a sentinel for the algorithm (do I understand correctly what you want?), you have to make sure it errors out early if you give it diag_scale=0, instead of shooting the user. Also, you must document it clearly.

If you prefer this, feel free to implement it.

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

attila-i-szabo:
Okay, I think I see the point. For 

> ```python
> @jax.jit
> def fun2(pars, sampler_state, shift):
>     _vs = nk.vqs.MCState(sa, ma, variables={"params":pars})
>     _vs.sampler_state = sampler_state
>     S = nk.optimizer.qgt.QGTJacobianDense(_vs, diag_shift=shift, mode="real", rescale_shift=True)
>     return S@pars
> 
> fun2(vs.parameters, vs.sampler_state, 0.01)
> ```

the trick is that if `diag_shift=0` (mathematically) then the rescaling algorithm works even if `diag_scale=0`. We could use `diag_scale=None` as a sentinel for this, so one could go for a dispatch table like this one:

`diag_shift` | `diag_scale` | algorithm | `scale` | `offset`
--- | --- | --- | --- | ---
`None` | `None` | no scale | 0.0 | `None`
X | `None` | no scale | X | `None`
`None` | Y | scale | Y | 0.0
X | Y | scale | Y | X/Y

The only catch is that the last line fails if Y=0, so we must cause it to error.

Re code organisation: it would be better to merge `sanitise_diag_shift` and `to_shift_offset` (I separated it because I thought it would be better to do `sanitise` even before `partial`s but then changed my mind).

PR #1485, opened 2023-05-30
 pass training_kwargs in VariationalStates

### LINKED ISSUE #1484: the training_kwargs in VariationalStates are unused
There's a dictionary parameter for `MCState` called `training_kwargs` that based on my understanding is for setting the variables that are going to be passed when training a model. this parameter is being saved as a frozen dictionary in `MCState` but i don't see it being used anywhere in the code-base.

this parameter is to be deprecated or just the feature was never finished?

vigsterkr:
add support for training_kwargs

### DIFF
--- a/netket/utils/model_frameworks/haiku.py
+++ b/netket/utils/model_frameworks/haiku.py
@@ -52,7 +52,6 @@ class HaikuWrapper:
 
 @framework
 class HaikuFramework(ModuleFramework):
-
     name: str = "Haiku"
 
     @staticmethod
--- a/netket/vqs/full_summ/expect.py
+++ b/netket/vqs/full_summ/expect.py
@@ -17,6 +17,7 @@ from typing import Callable, Tuple
 
 import jax
 from jax import numpy as jnp
+from flax.core.frozen_dict import FrozenDict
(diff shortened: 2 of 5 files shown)


PhilipVinc (commented):
Sorry it took me so long to check out.
That seems correct. 
Does it work fine for you?

Before I can merge, the following things are missing: 
 - We need at least one test case that would break without this feature. A tiny test case that would not work otherwise. You can add it to `test/variational/test_variational.py` for example. I think you could use a simple NN that uses batch norm or such.
 - You should document the description of the argument `training_kwargs` in the init method of `MCState` to describe what this does. 
 - You should add a few lines documenting this change to the CHANGELOG.md under `new features` describing what this change does

vigsterkr:
@PhilipVinc thnx for getting back to me! sure thing i'll add these requested changes with the rebase.
one question remains: i don't see how PRNG being passed to models - like in case of haiku? namely, i'd like to use a [dropout layer](https://dm-haiku.readthedocs.io/en/latest/api.html#haiku.dropout) like:
```
hk.dropout(hk.next_rng_key(), 0.5, x)
```

do you maybe have any ideas regarding this? if there's any patch needed for having this support maybe...?

PR #1511, opened 2023-06-28
[experimental] Add initial support for jax multi-device and multi-process environments (part 2)

inailuig:
Fully relies on the trivial automatic parallelization over the chains using shared global jax.Array's.

Does the following:
- turn the sampler state (just the \sigma) into a global device array using jit with out_shardings. 
- wrap the call to the numba operator so that the xp and mels  are shared arrays just like the samples inputted to it were, or use operators written in jax
- extract the results from the shared arrays before printing/saving
- use jax_threefry_partitionable=True (their new rng impl)

Supports both multiple local devices as well as [jax.distributed](https://jax.readthedocs.io/en/latest/multi_process.html) (the latter is only available on gpu and tpu)
On gpu it Internally uses the [nvidia NCCL](https://developer.nvidia.com/nccl) library for communication, so it should be quite efficient.
On cpu one can  use multiple devices with threads via `--xla_force_host_platform_device_count=XX`, multi-node is not available yet.

My initial benchmarks on gpu showed it's competitive with mpi.

### DIFF
--- a/docs/docs/sharp-bits.md
+++ b/docs/docs/sharp-bits.md
@@ -96,7 +96,7 @@ import netket as nk
 Then it can be conveniently launched with `srun` (on slurm clusters) or `mpirun`.
 For more details and manual setups we refer to the [jax documentation](https://jax.readthedocs.io/en/latest/multi_process.html).
 
-Note that jax internally uses the [grpc library](https://grpc.io) (essentially a rest api on a http server) to infer the cluster topology and the [nvidia nccl library](https://developer.nvidia.com/nccl) for communication between gpus. Thus it is required that `libnccl2` and `libnccl2-dev` are installed in addition to cuda.
+Note that jax internally uses the [grpc library](https://grpc.io) (launching a http server) for setup and book-keeping of the cluster and the [nvidia nccl library](https://developer.nvidia.com/nccl) for communication between gpus. Thus it is required that `libnccl2` and `libnccl2-dev` are installed in addition to cuda.
 Even if launched with mpirun, mpi is not actually used for communication, but the environment variables set by it are instead picked up by `jax.distributed.initialize`.
 
 If you run into communication errors you might want to set the environment variable `NCCL_DEBUG=INFO` for detailed error messages.
--- a/docs/docs/sharp-bits.md
+++ b/docs/docs/sharp-bits.md
@@ -73,8 +73,63 @@ Eventually we would like the selection to be automatic, but this has not yet bee
 
 Please open tickets if you find issues!
 
-(nan)=
+(jax_multi_process)=
+## Running on multi-gpu clusters with jax.distributed
+Historically the main way to run NetKet in parallel has been to use MPI. However, with jax adding shared arrays and collective operations on multiple devices/nodes (see [here](https://jax.readthedocs.io/en/latest/jax_array_migration.html#jax-array-migration) and [here](https://jax.readthedocs.io/en/latest/multi_process.html)) we adapted NetKet to support it.
+This feature is still experimental and can be enabled by setting the flag `NETKET_EXPERIMENTAL_SHARDING=1`.
 
+(jax_multi_process_setup)=
+### Setup
+To launch netket on a multi-node gpu cluster usually all that is required is to add a call to `jax.distributed.initialize()` at the top of the main script, e.g. as follows:
+
+```python
+import jax
+jax.distributed.initialize()
+
+import os
+os.environ['NETKET_EXPERIMENTAL_SHARDING'] = 1
+
+import netket as nk
+# ...
+```
+
+Then it can be conveniently launched with `srun` (on slurm clusters) or `mpirun`.
+For more details and manual setups we refer to the [jax documentation](https://jax.readthedocs.io/en/latest/multi_process.html).
+
+Note that jax internally uses the [grpc library](https://grpc.io) (essentially a rest api on a http server) to infer the cluster topology and the [nvidia nccl library](https://developer.nvidia.com/nccl) for communication between gpus. Thus it is required that `libnccl2` and `libnccl2-dev` are installed in addition to cuda.
+Even if launched with mpirun, mpi is not actually used for communication, but the environment variables set by it are instead picked up by `jax.distributed.initialize`.
+
+If you run into communication errors you might want to set the environment variable `NCCL_DEBUG=INFO` for detailed error messages.
+
+(grpc_proxy)=
+### GRPC incompatibility with http proxy wildcards
+We noticed that communication errors can arise when a http proxy is used on the cluster. Grpc will try to communicate with the other nodes via the proxy, whenever they are only excluded in the `no_proxy` variable via wildcards (e.g. `no_proxy=10.0.0.*`) which we found grpc cannot parse. To avoid this one needs to include all addresses explicitly.
+
+Alternatively, a simple way to work around it is to disable the proxy completely for jax by unsetting the respective environment variables (see [grpc docs](https://grpc.github.io/grpc/cpp/md_doc_environment_variables.html)) e.g. as follows:
+```python
+import os
+del os.environ['http_proxy']
+del os.environ['https_proxy']
+del os.environ['no_proxy']
+
+import jax
+jax.distributed.initialize()
+```
+
+(multi_device)=
+### Multiple devices per process
+In our testing it is best to use 1 process per gpu on the cluster.
+
+Nevertheless, if you want to use multiple gpus per process you can force jax to do so by setting `local_device_ids`, e.g. extracting it from `CUDA_VISIBLE_DEVICES` as follows:
+
+```python
+import os
+import jax
+ldi = list(map(int, os.environ.get('CUDA_VISIBLE_DEVICES').split(',')))
+jax.distributed.initialize(local_device_ids=ldi)
+```
+
+(nan)=
 ## NaNs in training and loss of precision
 
 If you find NaNs while training, especially if you are using your own model, there might be a few reasons:
--- a/docs/docs/sharp-bits.md
+++ b/docs/docs/sharp-bits.md
@@ -74,12 +74,18 @@ Eventually we would like the selection to be automatic, but this has not yet bee
 Please open tickets if you find issues!
 
 (jax_multi_process)=
-## Running on multi-gpu clusters with jax.distributed
+## Running on multi-gpu clusters
 Historically the main way to run NetKet in parallel has been to use MPI. However, with jax adding shared arrays and collective operations on multiple devices/nodes (see [here](https://jax.readthedocs.io/en/latest/jax_array_migration.html#jax-array-migration) and [here](https://jax.readthedocs.io/en/latest/multi_process.html)) we adapted NetKet to support it.
-This feature is still experimental and can be enabled by setting the flag `NETKET_EXPERIMENTAL_SHARDING=1`.
+To run on a single node with multiple gpus all that is necessary is to set the flag `NETKET_EXPERIMENTAL_SHARDING=1`.
+
+:::{warning}
+This feature is still experimental and not everything may work perfectly right out of the box.
+Any feedback, be it positive or negative, would be greatly appreciated.
+:::
+
 
 (jax_multi_process_setup)=
-### Setup
+### Multi-node setup with jax.distributed
 To launch netket on a multi-node gpu cluster usually all that is required is to add a call to `jax.distributed.initialize()` at the top of the main script, e.g. as follows:
 
 ```python
--- a/netket/driver/abstract_variational_driver.py
+++ b/netket/driver/abstract_variational_driver.py
@@ -59,7 +59,7 @@ class AbstractVariationalDriver(abc.ABC):
 
     def __init__(self, variational_state, optimizer, minimized_quantity_name=""):
         self._mynode = mpi.node_number
-        self._is_root = self._mynode == 0
+        self._is_root = self._mynode == 0 and jax.process_index() == 0
         self._mpi_nodes = mpi.n_nodes
         self._loss_stats = None
         self._loss_name = minimized_quantity_name
--- a/netket/jax/_vjp_chunked.py
+++ b/netket/jax/_vjp_chunked.py
@@ -162,6 +162,9 @@ def vjp_chunked(
 ):
     """calculate the vjp in small chunks for a function where the leading dimension of the output only depends on the leading dimension of some of the arguments
 
+    .. note::
+        If experimental sharing is activated, the chunk_argnums are assumed to be sharded (not replicated) among devices.
+
     Args:
         fun: Function to be differentiated. It must accept chunks of size chunk_size of the primals in chunk_argnums.
         primals:  A sequence of primal values at which the Jacobian of ``fun`` should be evaluated.
@@ -180,6 +183,7 @@ def vjp_chunked(
         a function corresponding to the vjp_fun returned by an equivalent ``jax.vjp(fun, *primals)[1]``` call
         which computes the vjp in chunks (recomputing the forward pass every time on subsequent calls).
         If return_forward=True the vjp_fun returned returns a tuple containing the output of the forward pass and the vjp.
+
 
     Example:
         >>> import jax
--- a/netket/jax/_vjp_chunked.py
+++ b/netket/jax/_vjp_chunked.py
@@ -10,6 +10,8 @@ from netket.jax import (
     vjp as nkvjp,
 )
 from netket.utils import HashablePartial
+from netket.utils import config
+from netket.jax.sharding import sharding_decorator
 
 from ._scanmap import _multimap
 from ._chunk_utils import _chunk as _tree_chunk, _unchunk as _tree_unchunk
@@ -227,6 +229,57 @@ def vjp_chunked(
 
     if chunk_argnums == ():
         chunk_size = None
+
+    ############################################################################
+    # sharding
+
+    if config.netket_experimental_sharding and chunk_size is not None:
+        if return_forward:
+            raise NotImplementedError
+
+        # assume the chunk_argnums are also sharded
+        # later we might introduce an extra arg for it
+        sharded_argnums = chunk_argnums
+        sharded_args = tuple(i in sharded_argnums for i in range(len(primals)))
+
+        # for the output we need to consult nondiff_argnums, which are removed
+        non_sharded_argnums = tuple(
+            set(range(len(primals))).difference(sharded_argnums)
+        )
+        out_args = _gen_append_cond_vjp(primals, nondiff_argnums, non_sharded_argnums)
+        red_ops = tuple(jax.lax.psum if c else False for c in out_args)
+
+        # check the chunk_size is not larger than the shard per device
+        chunk_size = sharding_decorator(
+            partial(check_chunk_size, chunk_argnums, chunk_size),
+            sharded_args_tree=sharded_args,
+            reduction_op_tree=True,
+        )(*primals)
+
+        if chunk_size is not None:
+            _vjpc = _vjp_chunked(
+                fun,
+                has_aux=has_aux,
+                chunk_argnums=chunk_argnums,
+                chunk_size=chunk_size,
+                nondiff_argnums=nondiff_argnums,
+                return_forward=return_forward,
+                conjugate=conjugate,
+            )
+            vjp_fun_sh = Partial(
+                sharding_decorator(
+                    _vjpc,
+                    sharded_args_tree=(sharded_args, True),
+                    reduction_op_tree=red_ops,
+                ),
+                primals,
+            )
+            return vjp_fun_sh
+        else:
+            pass  # no chunking, continue below
+
+    ############################################################################
+    # no sharding (or sharded, but not chunking)
 
     # check the chunk_size is not larger than the arrays
     chunk_size = check_chunk_size(chunk_argnums, chunk_size, *primals)
--- a/netket/jax/_vmap_chunked.py
+++ b/netket/jax/_vmap_chunked.py
@@ -11,7 +11,7 @@ from netket.utils import config
 from netket.jax.sharding import sharding_decorator
 
 
-def _fun(vmapped_fun, chunk_size, argnums, *args, **kwargs):
+def _eval_fun_in_chunks(vmapped_fun, chunk_size, argnums, *args, **kwargs):
     n_elements = jax.tree_util.tree_leaves(args[argnums[0]])[0].shape[0]
     n_chunks, n_rest = divmod(n_elements, chunk_size)
 
@@ -45,9 +45,11 @@ def _fun(vmapped_fun, chunk_size, argnums, *args, **kwargs):
     return y
 
 
-def _fun_sharding(vmapped_fun, chunk_size, argnums, *args, **kwargs):
+def _eval_fun_in_chunks_sharding(vmapped_fun, chunk_size, argnums, *args, **kwargs):
+    # Equivalent to `_eval_fun_in_chunks` above but preserves sharding,
+    # by computing the vmapped_fun in chunks on every shard (which sits on a separate device)
     sharded_args_tree = tuple(i in argnums for i, a in enumerate(args))
-    f = HashablePartial(_fun, vmapped_fun, chunk_size, argnums, **kwargs)
+    f = HashablePartial(_eval_fun_in_chunks, vmapped_fun, chunk_size, argnums, **kwargs)
     return sharding_decorator(f, sharded_args_tree)(*args)
 
 
@@ -65,9 +67,11 @@ def _chunk_vmapped_function(
     if isinstance(argnums, int):
         argnums = (argnums,)
     if axis_0_is_sharded:
-        return HashablePartial(_fun_sharding, vmapped_fun, chunk_size, argnums)
+        return HashablePartial(
+            _eval_fun_in_chunks_sharding, vmapped_fun, chunk_size, argnums
+        )
     else:
-        return HashablePartial(_fun, vmapped_fun, chunk_size, argnums)
+        return HashablePartial(_eval_fun_in_chunks, vmapped_fun, chunk_size, argnums)
 
 
 def _parse_in_axes(in_axes):
@@ -109,6 +113,10 @@ def apply_chunked(
 
         f = jax.vmap(f_orig)
 
+    .. note::
+        If netket_experimental_sharding is enabled, this function assumes that chunked in_axes are sharded by default.
+        This can be overridden by specifying axis_0_is_sharded=False.
+
     Args:
         f: A function that satisfies the condition above
         in_axes: The axes that should be scanned along. Only supports `0` or `None`
@@ -140,6 +148,10 @@ def vmap_chunked(
         nk.jax.apply_chunked(jax.vmap(f, in_axes), in_axes, chunk_size)
 
     Some limitations to `in_axes` apply.
+
+    .. note::
+        If netket_experimental_sharding is enabled, this function assumes that chunked in_axes are sharded by default.
+        This can be overridden by specifying axis_0_is_sharded=False.
 
     Args:
         f: The function to be vectorised.
--- a/netket/jax/_vmap_chunked.py
+++ b/netket/jax/_vmap_chunked.py
@@ -7,6 +7,8 @@ from ._chunk_utils import _chunk, _unchunk
 from ._scanmap import scanmap, scan_append
 
 from netket.utils import HashablePartial
+from netket.utils import config
+from netket.jax.sharding import sharding_decorator
 
 
 def _fun(vmapped_fun, chunk_size, argnums, *args, **kwargs):
@@ -43,10 +45,17 @@ def _fun(vmapped_fun, chunk_size, argnums, *args, **kwargs):
     return y
 
 
+def _fun_sharding(vmapped_fun, chunk_size, argnums, *args, **kwargs):
+    sharded_args_tree = tuple(i in argnums for i, a in enumerate(args))
+    f = HashablePartial(_fun, vmapped_fun, chunk_size, argnums, **kwargs)
+    return sharding_decorator(f, sharded_args_tree)(*args)
+
+
 def _chunk_vmapped_function(
     vmapped_fun: Callable,
     chunk_size: Optional[int],
     argnums=0,
+    axis_0_is_sharded=False,
 ) -> Callable:
     """takes a vmapped function and computes it in chunks"""
 
@@ -55,8 +64,10 @@ def _chunk_vmapped_function(
 
     if isinstance(argnums, int):
         argnums = (argnums,)
-
-    return HashablePartial(_fun, vmapped_fun, chunk_size, argnums)
+    if axis_0_is_sharded:
+        return HashablePartial(_fun_sharding, vmapped_fun, chunk_size, argnums)
+    else:
+        return HashablePartial(_fun, vmapped_fun, chunk_size, argnums)
 
 
 def _parse_in_axes(in_axes):
@@ -77,6 +88,7 @@ def apply_chunked(
     in_axes=0,
     *,
     chunk_size: Optional[int],
+    axis_0_is_sharded=config.netket_experimental_sharding,
 ) -> Callable:
     """
     Takes an implicitly vmapped function over the axis 0 and uses scan to
@@ -102,10 +114,13 @@ def apply_chunked(
         in_axes: The axes that should be scanned along. Only supports `0` or `None`
         chunk_size: The maximum size of the chunks to be used. If it is `None`, chunking
             is disabled
+        axis_0_is_sharded: specifies if axis 0 of the arrays scanned is sharded among multiple devices,
+            The function is then computed in chunks of size chunk_size on every device.
+            Defaults True if config.netket_experimental_sharding, oterhwise defaults to False.
 
     """
     _, argnums = _parse_in_axes(in_axes)
-    return _chunk_vmapped_function(f, chunk_size, argnums)
+    return _chunk_vmapped_function(f, chunk_size, argnums, axis_0_is_sharded)
 
 
 def vmap_chunked(
@@ -113,6 +128,7 @@ def vmap_chunked(
     in_axes=0,
     *,
     chunk_size: Optional[int],
+    axis_0_is_sharded=config.netket_experimental_sharding,
 ) -> Callable:
     """
     Behaves like jax.vmap but uses scan to chunk the computations in smaller chunks.
@@ -130,10 +146,13 @@ def vmap_chunked(
         in_axes: The axes that should be scanned along. Only supports `0` or `None`
         chunk_size: The maximum size of the chunks to be used. If it is `None`, chunking
             is disabled
+        axis_0_is_sharded: specifies if axis 0 of the arrays scanned is sharded among multiple devices,
+            The function is then computed in chunks of size chunk_size on every device.
+            Defaults True if config.netket_experimental_sharding, oterhwise defaults to False.
 
     Returns:
         A vectorised and chunked function
     """
     in_axes, argnums = _parse_in_axes(in_axes)
     vmapped_fun = jax.vmap(f, in_axes=in_axes)
-    return _chunk_vmapped_function(vmapped_fun, chunk_size, argnums)
+    return _chunk_vmapped_function(vmapped_fun, chunk_size, argnums, axis_0_is_sharded)
--- a/netket/jax/sharding.py
+++ b/netket/jax/sharding.py
@@ -29,7 +29,7 @@ from jax.tree_util import Partial
 from jax.sharding import Mesh, PartitionSpec as P, PositionalSharding
 from jax.experimental.shard_map import shard_map
 
-from netket.utils import config
+from netket.utils import config, mpi
 from netket.errors import concrete_or_error, NumbaOperatorGetConnDuringTracingError
 
 
@@ -158,7 +158,10 @@ def distribute_to_devices_along_axis(
     inp_data, axis=0, pad=False, pad_value=None, devices=jax.devices()
 ):
     """
-    Distribute a local array equally along an axis to multiple jax devices devices.
+    Distribute a local array equally along an axis to multiple jax devices devices
+
+     .. note:
+        Does nothing if netket.config.netket_experimental_sharding=False.
 
      .. note:
         Each jax process needs to have the whole array (parts not belonging to it can be filled with garbage).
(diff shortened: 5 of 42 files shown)


netket/utils/mpi/mpi.py
```diff
@@ -25,9 +25,9 @@
 
 try:
     if not config.netket_mpi:  # pragma: no cover
-        # if mpi is disabled, import a package that does not exist
-        # to trigger import error and follow the no-mpi code path
-        import this_package_does_not_exist_zuzzurellone  # noqa: F401
+        # if mpi is disabled trigger import error
+        # and follow the no-mpi code path
+        raise ImportError
```
PhilipVinc: I'm very saddened that this code disappears and that you define it <embarassing>.

netket/jax/_distributed.py
```diff
@@ -0,0 +1,63 @@
...
+from functools import partial, wraps
+import numpy as np
+
+
+from jax.sharding import Mesh, PartitionSpec as P
+from jax.experimental.shard_map import shard_map
+
+
+@partial(jax.jit, static_argnums=0)
+def _replicate_shmap_callback(f, x):
```
PhilipVinc, line 11: doc?

netket/jax/_distributed.py
```diff
@@ -0,0 +1,63 @@
...
+            x.shape[:-1] + dummy_mels.shape[x.ndim - 1 :], dummy_mels.dtype
+        )
+        result_shape = (xp_shape, mels_shape)
+        return jax.pure_callback(f, result_shape, x, vectorized=True)
+
+    return _f(x)
+
+
+def replicate_sharding_shmap(f):
+    return partial(_replicate_shmap_callback, f)
```
PhilipVinc, line 34: doctoring or comment

netket/jax/utils.py
```diff
@@ -304,7 +305,10 @@ def PRNGKey(
     The same seed will be distributed to all processes.
     """
     if seed is None:
-        key = jax.random.PRNGKey(random_seed())
+        seed = random_seed()
+        if config.netket_experimental_pjit and jax.process_count() > 1:
+            seed = jax.experimental.multihost_utils.broadcast_one_to_all(seed)
```
PhilipVinc, line 310: shouldn't this go below, at the same level of the MPI call?
otherwise one might end up with different seeds on different nodes?

netket/stats/mc_stats.py
```diff
@@ -20,6 +20,7 @@
 from jax import numpy as jnp
 
 from netket.utils import config, mpi, struct
+from netket.jax._distributed import extract_replicated  # TODO; avoid cirular import
```
PhilipVinc, line 23: reminder

netket/jax/_distributed.py
```diff
@@ -0,0 +1,63 @@
...
+    # each process needs to have the whole array; parts not belonging to it can be filled with garbage
+    shape = [
+        1,
+    ] * inp_data.ndim
+    shape[axis] = -1
+    sharding = jax.sharding.PositionalSharding(jax.devices()).reshape(shape)
+    return jax.jit(_identity, out_shardings=sharding)(inp_data)
+
+
+def extract_replicated(t):
```
PhilipVinc, line 54: docstring

netket/utils/config_flags.py
```diff
@@ -229,3 +229,15 @@ def update(self, name, value):
...
     runtime=True,
 )
+
+config.define(
+    "NETKET_EXPERIMENTAL_PJIT",
+    bool,
+    default=False,
+    help=dedent(
+        """
+        TODO
```
PhilipVinc, line 239: reminder

netket/vqs/mc/mc_state/expect.py
```diff
@@ -45,7 +46,7 @@ def get_local_kernel_arguments(vstate: MCState, Ô: Squared):  # noqa: F811
     check_hilbert(vstate.hilbert, Ô.hilbert)
 
     σ = vstate.samples
-    σp, mels = Ô.parent.get_conn_padded(σ)
+    σp, mels = replicate_sharding(Ô.parent.get_conn_padded)(σ)
```
PhilipVinc, line 49: what does this do exactly if you have 2 nodes with M devices each?

inailuig:
> ~Soldato~ Clemens, what you think is required before we can merge this?

really only docs.
And, if you insist on it, some warnings  and errors telling people it is experimental and a lot of stuff does not work yet with it.

PhilipVinc:
Ok. Can you make a list of what works and what does not?

netket/vqs/mc/mc_state/expect.py
inailuig: it will callback to the  numba operator M times on each of the 2 nodes on n_samples / (2 M) size inputs.

netket/jax/utils.py
inailuig: The only think I wanted to make sure is that all the processes have the same seed if we use the default one which comes from numba. It's not strictly necessary, but this way we have the same random state on all of them, so it's easier to reproduce results
In any other case we require the user to run the same program on every process, and give us the same seed/inital key.
Jax then knows how many devices and processes there are and generates a different random number depending on the process and device, just like it would be if it was a single one, given the same key.

Also note that we will [not be able to distribute keys in the future anyway](https://github.com/google/jax/pull/16511#issuecomment-1603923832).

inailuig:
> Ok. Can you make a list of what works and what does not?

_works_:
- VMC with SR, both numba and jax operators

_does not work_:
- exactstate
- any kind of chunking
- saving samples etc to disk

_unknown_:
- dynamics, mixed states, ...

netket/jax/_distributed.py
inailuig: documented the one calling it below

netket/jax/utils.py
PhilipVinc: food point. thanks Clemens.

netket/sampler/metropolis.py
```diff
@@ -249,13 +253,17 @@ def _init_state(sampler, machine, params, key):
             (sampler.n_chains_per_rank, sampler.hilbert.size), dtype=sampler.dtype
         )
 
+        if config.netket_experimental_pjit and jax.device_count() > 1:
+            σ = put_global(σ)
```
inailuig, line 257: If we end up rewriting the hilbert spaces in jax  then we can avoid this and instead jit the random state with the correct out_shardings param

netket/driver/abstract_variational_driver.py
```diff
@@ -232,7 +232,7 @@ def run(
             )
 
         # Log only non-root nodes
-        if self._mynode == 0:
+        if self._mynode == 0 and jax.process_index() == 0:
```
gcarleo, line 235: can we just define something like self._is_root = self._mynode == 0 and jax.process_index() == 0 ?

netket/stats/mc_stats.py
```diff
@@ -108,14 +109,15 @@ def to_compound(self):
         return "Mean", self.to_dict()
 
     def __repr__(self):
-        mean, err, var = _format_decimal(self.mean, self.error_of_mean, self.variance)
-        if not math.isnan(self.R_hat):
-            ext = f", R̂={self.R_hat:.4f}"
+        s = extract_replicated(self)
```
PhilipVinc, line 112:
```suggestion
        # Un-replicate replicated arrays (which are fully replicated among processes)
        self = extract_replicated(self)
```

and

netket/jax/_jacobian/logic.py
```diff
@@ -304,8 +304,8 @@ def jacobian(
     # - (n_samples, 2, ...) if mode complex, holding the real and imaginary jacobian
     # - (n_samples, ...) if mode real/holomorphic
     jacobians = vmap_chunked(
-        jacobian_fun, in_axes=(None, None, 0), chunk_size=chunk_size
-    )(f, params, samples)
+        partial(jacobian_fun, f), in_axes=(None, 0), chunk_size=chunk_size
+    )(params, samples)
```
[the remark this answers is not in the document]
inailuig: shard map does not support non-array arguments. As discussed I opened a feature request on the jax repo

netket/driver/abstract_variational_driver.py
inailuig: done

netket/jax/distributed.py
```diff
@@ -0,0 +1,119 @@
...
+        )
+        result_shape = (xp_shape, mels_shape)
+        return jax.pure_callback(f, result_shape, x, vectorized=True)
+
+    return _f(x)
+
+
+# TODO eventually make this a bit more univesal
+# supporting more fancy sharding and multiple input/output args
+def replicate_sharding_shmap(f):
```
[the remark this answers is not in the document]
inailuig: done

netket/vqs/full_summ/expect.py
```diff
@@ -41,27 +44,41 @@ def _check_hilbert(A, B):
...
     # TODO: This performs the full computation on all MPI ranks.
     # It would be great if we could split the computation among ranks.
 
-    OΨ = O @ Ψ
-    expval_O = (Ψ.conj() * OΨ).sum()
+    if config.netket_experimental_pjit:
+        # TODO store blocks of rows on each device
+        # once jax supports shared sparse arrays
+        O = sparsify(Ô, _pad_size=Ψ.shape[0])
+        OΨ = jax.jit(operator.matmul, O, Ψ)
```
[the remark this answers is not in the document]
inailuig: done

netket/stats/mc_stats.py
inailuig: done

netket/optimizer/qgt/qgt_onthefly_logic.py
```diff
@@ -93,66 +98,101 @@ def fun(W):
...
+    @partial(shard_map, mesh=mesh, in_specs=in_specs, out_specs=out_specs)
+    def _jvp(params, samples, v):
+        samples, ucf = chunk(samples, chunk_size)
+        res = __O_jvp(forward_fn, params, samples, v)
+        return ucf(res)
+
+    return _jvp(params, samples, v)
+
+
+def _O_vjp(forward_fn, params, samples, w, chunk_size):
```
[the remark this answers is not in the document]
inailuig: did it

netket/jax/distributed.py
```diff
@@ -0,0 +1,159 @@
...
+            return x
+
+    return jax.tree_map(_extract_replicated, t)
+
+
+def gather(x):
+    return jax.jit(_identity, out_shardings=x.sharding.replicate())(x)
+
+
+def sharding_decorator(f, sharded_argnums, reduction_op=None):
```
PhilipVinc, line 120: please a docstring here.

PhilipVinc (commented):
So, thank you great Clemens-san.

My comments, as you can hear from me are:
 - I now like (or rather, not dislike) the QGT implementation. I deem it human.
 - discussions face toffee

netket/sampler/exact.py
```diff
@@ -83,6 +86,9 @@ def _reset(sampler, machine, parameters, state):
         )
         pdf = pdf / pdf.sum()
 
+        if config.netket_experimental_sharding:
+            pdf = gather(pdf)[: sampler.hilbert.n_states]
```
PhilipVinc, line 90: can't this be handled by the argument `allgather=True`of `nn.to_array()`? it's already True by default and should guarantee this. I think you could copy-paste those lines in there.

PhilipVinc: By the way, this way there is no need to add the code to handle sharding in the operators to sparse representation.
Don't trash it! But remove those changes from this PR.

netket/jax/_vmap_chunked.py
```diff
@@ -70,7 +94,13 @@ def _parse_in_axes(in_axes):
...
 
 
-def apply_chunked(f: Callable, in_axes=0, *, chunk_size: Optional[int]) -> Callable:
+def apply_chunked(
+    f: Callable,
+    in_axes=0,
+    *,
+    chunk_size: Optional[int],
+    axis_0_is_sharded=config.netket_experimental_pjit,
+) -> Callable:
```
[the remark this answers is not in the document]
inailuig: added it

netket/jax/_vmap_chunked.py
```diff
@@ -98,10 +128,16 @@ def apply_chunked(f: Callable, in_axes=0, *, chunk_size: Optional[int]) -> Calla
...
-def vmap_chunked(f: Callable, in_axes=0, *, chunk_size: Optional[int]) -> Callable:
+def vmap_chunked(
+    f: Callable,
+    in_axes=0,
+    *,
+    chunk_size: Optional[int],
+    axis_0_is_sharded=config.netket_experimental_pjit,
+) -> Callable:
     """
     Behaves like jax.vmap but uses scan to chunk the computations in smaller chunks.
```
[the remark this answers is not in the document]
inailuig: added it

netket/jax/distributed.py
```diff
@@ -176,6 +177,7 @@ def sharding_decorator(f, sharded_args_tree, reduction_op_tree=False):
         reduction_op_tree: a tuple/pyrtree of reduction_op, where for each output:
             reduction_op is e.g. jax.lax.psum if it is to be reduced, then f_wrapped returns a replicated array
             reduction op is False if it is not to be reduced, then f_wrapped returns a sharded array
+            reduction op is True if it is not an array/pytree, then it is returned as python object
```
inailuig: proposals for a better interface? maybe could use strings ('static', 'None', etc) ?

(cannot use None as it's ignored by the tree_map)

PhilipVinc:
@inailuig Do you think it's just an accuracy problem? It's a bit surprising that those tests, which never failed before, are starting to fail now, even if they are close to the target accuracy...

inailuig:
> @inailuig Do you think it's just an accuracy problem? It's a bit surprising that those tests, which never failed before, are starting to fail now, even if they are close to the target accuracy...

looks like it. After all, all I changed was to double the number of samples. I shall try to increase the tolerance a tiny bit

Also the test with 2 cpus is not yet working (env flags seem to be ignored). Locally I am still getting a few errors there.

gcarleo:
Do we understand why we need to increase the tolerance of the test though? I wouldn't recommend reducing the accuracy of tests unless there is some specific reason!

PhilipVinc:
@inailuig can you add a file `conf/run_multidevice_tests.sh` with the correct incantantation/env variables to run the multi device tests?

test/optimizer/test_qgt_logic.py
```diff
@@ -259,7 +261,7 @@ def test_reassemble_complex(e):
 
 
 @common.named_parametrize("holomorphic", [True, False])
-@common.named_parametrize("n_samp", [24, 1024])
+@common.named_parametrize("n_samp", [48, 1024])
```
inailuig, line 264:
```suggestion
@common.named_parametrize("n_samp", [24*jax.device_count(), 1024])
```

test/optimizer/test_qgt_logic.py
```diff
@@ -288,20 +292,22 @@ def f(params_model_state, x):
...
```
inailuig, line 295:
```suggestion
@common.named_parametrize("n_samp", [24*jax.device_count(), 1024])
```

test/optimizer/test_qgt_logic.py
```diff
@@ -332,7 +338,7 @@ def mvt(v, w):
 
 # TODO separate test for prepare_centered_oks
 @common.named_parametrize("holomorphic", [True])
-@common.named_parametrize("n_samp", [25, 1024])
+@common.named_parametrize("n_samp", [50, 1024])
```
inailuig, line 341:
```suggestion
@common.named_parametrize("n_samp", [25*jax.device_count(), 1024])
```

test/optimizer/test_qgt_logic.py
```diff
@@ -369,7 +376,7 @@ def test_matvec_treemv(e, jit, holomorphic, pardtype, outdtype, chunk_size):
 # TODO separate test for prepare_centered_oks
 # TODO test C->R ?
 @common.named_parametrize("holomorphic", [True, False])
-@common.named_parametrize("n_samp", [25, 1024])
+@common.named_parametrize("n_samp", [50, 1024])
```
inailuig, line 379:
```suggestion
@common.named_parametrize("n_samp", [25*jax.device_count(), 1024])
```

test/optimizer/test_qgt_logic.py
```diff
@@ -425,7 +432,7 @@ def e_offset(n_samp, outdtype, pardtype, holomorphic, offset, seed=123):
 
 
 @pytest.mark.parametrize("holomorphic", [True])
-@pytest.mark.parametrize("n_samp", [25, 1024])
+@pytest.mark.parametrize("n_samp", [50, 1024])
```
inailuig, line 435:
```suggestion
@pytest.mark.parametrize("n_samp", [25*jax.device_count(), 1024])
```

netket/vqs/full_summ/expect.py
```diff
@@ -79,7 +83,7 @@ def expect_and_forces(
 
     O = sparsify(Ô)
     Ψ = vstate.to_array()
-    OΨ = O @ Ψ
+    OΨ = jax.jit(lambda x, y: x @ y)(O, Ψ)  # jit for gda
```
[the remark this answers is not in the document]
inailuig: with gda i refer to global device arrays (which have been merged into jax.Array, [see docs](https://jax.readthedocs.io/en/latest/jax_array_migration.html#new-public-type-named-jax-array))

it's to avoid this error:
```RuntimeError: Running operations on `Array`s that are not fully addressable by this process (i.e. `Array`s with data sharded across multiple devices and processes.) is dangerous. It’s very important that all processes run the same cross-process computations in the same order otherwise it can lead to hangs. If you’re not already familiar with JAX’s multi-process programming model, please read https://jax.readthedocs.io/en/latest/multi_process.html. To fix this error, run your `jitted` computation inside `with jax.spmd_mode('allow_all'):` context manager.```

I guess I could have used the [context manager](https://github.com/google/jax/blob/57e33dc3b553fc49095090d0389b7cbf0e694728/jax/_src/config.py#L847-L857) here, so far I have just added jit around everything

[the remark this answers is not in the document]
Comment (PhilipVinc) [reply netket/vqs/full_summ/expect.py:86]: @inailuig didn't we agree that O is just a fully replicated matrix for now, as should be psi, in fullsumstate, for now?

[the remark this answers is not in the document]
Comment (inailuig) [reply netket/vqs/full_summ/expect.py]: I changed it so that I just extract one local shard from the (replicated) global array, now I just do everything like in the scalar case.

File: netket/jax/_scanmap.py [hunk 27]
```diff
@@ -82,10 +82,27 @@ def f_(carry, x):
...
+            y_reduce = jax.tree_map(
+                partial(jax.lax.select, is_first), y_op, op(y_carry, y_op)
+            )
+        else:  # workaround
+            if op is _tree_add:
+                # the zeros in carry_init above are the 0 element of _tree_add
+                # so we don't need the select
+                y_reduce = op(y_carry, y_op)
+            else:
+                raise NotimplementedError(
```
inailuig, line 103:
```suggestion
                raise NotImplementedError(
```

PhilipVinc:
@inailuig The error is an OOM? Do you have any idea?

inailuig:
> @inailuig The error is an OOM? Do you have any idea?

I believe so. It's difficult to debug as it does not happen consistently, sometimes the runner is kiled, sometimes not

inailuig:
@PhilipVinc can you have another look, I think this is more or less ready to merge. I'll add a commit with a few lines of documentation on how to run it soon.

test_sharding/run_standard_tests_with_sharding.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
```
PhilipVinc, line 1: can you add a readmit ile to this folder explaining briefly what the scripts do?

PhilipVinc (approved):
Thanks @inailuig .
 
I have a few remarks asking for more documentation to ease maintainability and usage by other people in the future, otherwise there are only two technical remarks hidden in there.

Other than that, I think this is mergeable once the doc points above are addressed.

netket/sampler/autoreg.py
```diff
@@ -137,6 +138,11 @@ def scan_fun(carry, index):
             dtype=sampler.dtype,
         )
 
+        if config.netket_experimental_sharding:
+            σ = jax.lax.with_sharding_constraint(
```
gcarleo, line 142: what is this doing? a small comment maybe?

PhilipVinc: This is one of the few clear things 😂

netket/sampler/base.py
```diff
@@ -69,50 +69,6 @@ class Sampler(abc.ABC):
...
     """The dtype of the states sampled."""
 
-    def __pre_init__(
-        self, hilbert: AbstractHilbert, n_chains: Optional[int] = None, **kwargs
-    ):
-        """
-        Construct a Monte Carlo sampler.
-
-        Args:
-            hilbert: The Hilbert space to sample.
```
gcarleo, line 79: do we lose these docstring?

netket/optimizer/qgt/qgt_onthefly_logic.py
```diff
@@ -92,6 +93,7 @@ def fun(W):
 # Methods below are needed for the chunked version of QGTOnTheFly
 
 
+@partial(sharding_decorator, sharded_args_tree=(False, False, True, False, False))
```
gcarleo, line 96: having condensed all in this decorator is excellent, thanks!

gcarleo (commented):
sharding_decorator makes things much more readable in places where sharding details are not needed, thanks! the part I personally like the least are the changes to metropolis.py. Why do we need `serialize_MetropolisSamplerState_sharding` and similar? Is there a way to somehow hide them from the sampler?

netket/sampler/metropolis.py
inailuig: done

docs/docs/sharp-bits.md
```diff
@@ -73,8 +73,63 @@ Eventually we would like the selection to be automatic, but this has not yet bee
 
 Please open tickets if you find issues!
 
-(nan)=
+(jax_multi_process)=
+## Running on multi-gpu clusters with jax.distributed
+Historically the main way to run NetKet in parallel has been to use MPI. However, with jax adding shared arrays and collective operations on multiple devices/nodes (see [here](https://jax.readthedocs.io/en/latest/jax_array_migration.html#jax-array-migration) and [here](https://jax.readthedocs.io/en/latest/multi_process.html)) we adapted NetKet to support it.
+This feature is still experimental and can be enabled by setting the flag `NETKET_EXPERIMENTAL_SHARDING=1`.
```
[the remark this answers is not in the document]
inailuig: done

netket/jax/_vjp_chunked.py
```diff
@@ -228,6 +230,57 @@ def vjp_chunked(
...
 
+    ############################################################################
+    # sharding
+
+    if config.netket_experimental_sharding and chunk_size is not None:
+        if return_forward:
+            raise NotImplementedError
+
+        # assume the chunk_argnums are also sharded
+        # later we might introduce an extra arg for it
```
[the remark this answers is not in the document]
inailuig: done

netket/jax/_vjp_chunked.py
```diff
@@ -228,6 +230,57 @@ def vjp_chunked(
...
+        red_ops = tuple(jax.lax.psum if c else False for c in out_args)
+
+        # check the chunk_size is not larger than the shard per device
+        chunk_size = sharding_decorator(
+            partial(check_chunk_size, chunk_argnums, chunk_size),
+            sharded_args_tree=sharded_args,
+            reduction_op_tree=True,
+        )(*primals)
+
+        if chunk_size is not None:
```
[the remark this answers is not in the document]
inailuig: It is taken per device now, however this is not documented anywhere explicitly

netket/jax/_vmap_chunked.py
```diff
@@ -43,10 +45,17 @@ def _get_rest(x):
     return y
 
 
+def _fun_sharding(vmapped_fun, chunk_size, argnums, *args, **kwargs):
```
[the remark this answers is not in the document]
inailuig: done

netket/jax/_vmap_chunked.py
```diff
@@ -7,6 +7,8 @@
 from ._scanmap import scanmap, scan_append
 
 from netket.utils import HashablePartial
+from netket.utils import config
+from netket.jax.sharding import sharding_decorator
 
 
 def _fun(vmapped_fun, chunk_size, argnums, *args, **kwargs):
```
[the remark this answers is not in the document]
inailuig: done

netket/jax/_vmap_chunked.py
```diff
@@ -102,17 +114,21 @@ def apply_chunked(
         in_axes: The axes that should be scanned along. Only supports `0` or `None`
         chunk_size: The maximum size of the chunks to be used. If it is `None`, chunking
             is disabled
+        axis_0_is_sharded: specifies if axis 0 of the arrays scanned is sharded among multiple devices,
```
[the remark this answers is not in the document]
inailuig: done

netket/jax/_vmap_chunked.py
```diff
@@ -130,10 +146,13 @@ def vmap_chunked(
...
```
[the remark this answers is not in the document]
inailuig: done

netket/jax/_vjp_chunked.py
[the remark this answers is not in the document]
PhilipVinc: great, thanks. That's ok.

netket/jax/sharding.py
```diff
@@ -0,0 +1,282 @@
...
+from jax.sharding import Mesh, PartitionSpec as P
+from jax.experimental.shard_map import shard_map
+
+from netket.utils import config
+from netket.errors import concrete_or_error, NumbaOperatorGetConnDuringTracingError
+
+
+def replicate_sharding(f):
+    """
+    Wrapper for python get_conn_padded to make it work with shared/global device arrays.
```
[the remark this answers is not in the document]
inailuig: No this is specifically hardcoded for get_conn_padded. I changed it's name to make it clear. Also I added a warning that people should use jax operators instead if possible . I have seen it deadlock on multi-gpu a few times (on cpu was fine). In any case this function is a huge mess, but for now It does what we need (mainly I wanted to run the tests without having to change everything)

netket/jax/sharding.py
```diff
@@ -0,0 +1,282 @@
...
+    Args:
+        t: a jax Array (or a pytree of jax Arrays)
+
+    Returns:
+        A locally adressable representation of t
+    """
+
+    def _extract_replicated(x):
+        if isinstance(x, jax.Array) and not x.is_fully_addressable:
+            assert x.is_fully_replicated
```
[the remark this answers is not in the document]
inailuig: done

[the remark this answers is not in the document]
inailuig: done

netket/jax/sharding.py
```diff
@@ -0,0 +1,282 @@
...
+            return x.addressable_data(0)
+        else:
+            return x
+
+    return jax.tree_map(_extract_replicated, t)
+
+
+def gather(x):
+    """
+    make a sharded array fully replicated by gathering all parts on every device
```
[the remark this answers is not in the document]
inailuig: done

netket/jax/sharding.py
```diff
@@ -0,0 +1,282 @@
...
+    """
+    return jax.jit(_identity, out_shardings=x.sharding.replicate())(x)
+
+
+def broadcast(x):
+    """
+    broadcast an array to all devices. Input on different processes is assumed to be the same
```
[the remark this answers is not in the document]
inailuig: done

netket/jax/sharding.py
```diff
@@ -0,0 +1,282 @@
...
+
+
+def sharding_decorator(f, sharded_args_tree, reduction_op_tree=False):
+    """
+    A decorator which wraps a function so that it is evaluated on every shard of the distributed arguments,
+    and the output is either returned sharded, or reduced with a collective operation.
+
+    Does nothing unless config.netket_experimental_sharding=True.
+    Intended for netket internal use only, interface might change in the future depending on requirements.
+
```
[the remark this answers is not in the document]
inailuig: done

netket/jax/sharding.py
```diff
@@ -0,0 +1,282 @@
+import math
+from functools import partial, wraps
+
```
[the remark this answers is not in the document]
inailuig: done

netket/jax/sharding.py
```diff
@@ -0,0 +1,282 @@
+import math
```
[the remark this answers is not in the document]
inailuig: done

netket/jax/sharding.py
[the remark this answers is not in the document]
PhilipVinc: it's ok, thanks. Documenting it clearly is what we need.

netket/jax/sharding.py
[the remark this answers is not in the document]
PhilipVinc: doubly done!

netket/logging/json_log.py
```diff
@@ -173,7 +175,9 @@ def _flush_params(self, variational_state):
 
         _time = time.time()
 
-        binary_data = serialization.to_bytes(variational_state.variables)
+        binary_data = serialization.to_bytes(
+            extract_replicated(variational_state.variables)
+        )
```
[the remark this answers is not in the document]
inailuig: the logger can run on any process, and will output the same (if you have shared filesystem you should obviously only run it on one of them) . If you log the parameters (or the sampler state for which I am using a custom flax serialization routine) they are all gathered to every process and then serialized. Other quantities like the stats are already replicated, so I just turn them into a non sharded array (even if its fully replicated you cannot access it from python if its on more than one process) and serialize that

netket/nn/utils.py
```diff
@@ -74,13 +79,23 @@ def to_array(
         chunk_size: Optional integer to specify the largest chunks of samples that
             the model will be evaluated upon. By default it is `None`, and when specified
             samples are split into chunks of at most `chunk_size`.
+
+    When running with netket_experimental_sharding=True:
+      If allgather=True, the final wave function is fully replicated array
+      If allgather=False, the final wave function is a sharded array,
+                          padded with zeros to the next multiple of the number of devices
```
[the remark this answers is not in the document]
inailuig: moved it into args

netket/sampler/metropolis.py
```diff
@@ -96,6 +102,34 @@ def __repr__(self):
...
+def deserialize_MetropolisSamplerState_sharding(sampler_state, state_dict):
+    for prop in ["σ", "n_accepted_proc"]:
+        x = state_dict[prop]
+        if x is not None:
+            state_dict[prop] = put_global(x)
+    return sampler_state.replace(**state_dict)
+
+
+if config.netket_experimental_sharding:
+    serialization.register_serialization_state(
```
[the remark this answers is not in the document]
inailuig: done

netket/sampler/metropolis.py
```diff
@@ -176,6 +241,7 @@ def __pre_init__(self, hilbert: AbstractHilbert, rule: MetropolisRule, **kwargs)
                 `n_chains/mpi.n_nodes` chains. In general, we recommend specifying `n_chains_per_rank`
                 as it is more portable.
             n_chains_per_rank: Number of independent chains on every MPI rank (default = 16).
+            n_chains_per_device: Number of independent chains on every jax device (default = 16).
```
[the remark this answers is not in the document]
inailuig: I decided that for now the only simple solution is to separate the idea of devices and ranks. If we don't use MPI I need to have exactly 1 rank, and n_chains_per_rank needs to be the total number of chains (otherwise we have to change it everywhere, which I wouldn't mind we do later on)

If you want I can change the argument of this the function, but we have to live with the inconsistency that in the object itself the n_chains_per_rank property will then be n_chains_per_rank*n_devices, different from what you set

test_sharding/run_standard_tests_with_sharding.sh
inailuig: done

netket/sampler/metropolis.py
```diff
@@ -153,7 +217,8 @@ class MetropolisSampler(Sampler):
 
     The dtype of the sampled states can be chosen.
     """
-
+    n_chains_per_rank: int = struct.field(pytree_node=False, default=None, repr=False)
```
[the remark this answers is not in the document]
inailuig: I moved it, because I could not see a reason for it to be in the base class anymore, and it simplified things as the number of chains computation was split between the two . See the commit 06a18733329a67c0954d87c69ad0147fdfdce6ff

netket/sampler/base.py
inailuig: the dataclass docstring above contains essentially all of the same information already. Also this one was in my understanding never seen by the users, as it was overridden in the child classes

netket/sampler/exact.py
```diff
@@ -126,4 +127,11 @@ def _sample_chain(
             sampler.n_chains_per_rank, chain_length, sampler.hilbert.size
         )
 
+        # TODO run the part above in parallel
+        if config.netket_experimental_sharding:
+            samples = jax.lax.with_sharding_constraint(
+                samples,
+                jax.sharding.PositionalSharding(jax.devices()).reshape(1, -1, 1),
```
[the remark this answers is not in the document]
inailuig: with mpi and exactsampler we have 1 fake chain per rank. Here I don't need to to this and instead I have 1 global chain (axis 0 of size 1), whose samples are sharded (axis 1 of size Ns/n_devices on each device) . Since we flatten the samples later, this gives the same as we would have had with a separate fake chain per device

netket/sampler/autoreg.py
```diff
@@ -63,7 +64,7 @@ def __pre_init__(self, *args, **kwargs):
             warn_deprecation(
                 "Specifying `n_chains` or `n_chains_per_rank` when constructing exact samplers is deprecated."
             )
-
+        kwargs["n_chains_per_rank"] = 1
```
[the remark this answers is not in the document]
inailuig: see the other comments

netket/sampler/autoreg.py
```diff
@@ -141,6 +142,11 @@ def scan_fun(carry, index):
             dtype=sampler.dtype,
         )
 
+        if config.netket_experimental_sharding:
+            σ = jax.lax.with_sharding_constraint(
+                σ, jax.sharding.PositionalSharding(jax.devices()).reshape(-1, 1)
+            )
```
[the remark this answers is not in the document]
inailuig: for now without mpi `n_chains===n_chains_per_rank` as for jax there is only 1 rank

inailuig:
> sharding_decorator makes things much more readable in places where sharding details are not needed, thanks! the part I personally like the least are the changes to metropolis.py. Why do we need `serialize_MetropolisSamplerState_sharding` and similar? Is there a way to somehow hide them from the sampler?

If you want to be able to serialize the sampler state ( > 1 jax process, using flax.serialization) then this is necessary. I added a comment to the code.
Otherwise we could hardcode it in the loggers, but I wanted to aovid that, or move these functions to a different file (where nobody might find it, here seemed like the natural place to put it)

PhilipVinc:
> If you want to be able to serialize the sampler state ( > 1 jax process, using flax.serialization) then this is necessary. I added a comment to the code.
Otherwise we could hardcode it in the loggers, but I wanted to aovid that, or move these functions to a different file (where nobody might find it, here seemed like the natural place to put it)

Yeah, I agree with you.
Actually we should just move over to `orbax` which can serialise sharded arrays but... 
work.

One day,

netket/sampler/base.py
PhilipVinc: it was shown here https://netket.readthedocs.io/en/latest/api/_generated/samplers/netket.sampler.Sampler.html#netket.sampler.Sampler but now you removed (see https://netket--1511.org.readthedocs.build/en/1511/api/_generated/samplers/netket.sampler.Sampler.html#netket.sampler.Sampler )

Can you just leave the method defined and simply return `(Hilbert, n_chains), kwargs`, in order to have the docstring?

inailuig: I reverted the constructor in the base to the previous behaviour in case anybody outside was using it, wdyt?

netket/sampler/metropolis.py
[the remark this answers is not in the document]
gcarleo: but then you should be careful with defaults, does it make sense to have both default to 16?

test/jax/test_math.py
```diff
@@ -36,7 +36,7 @@ def test_logdet():
     assert ld.dtype == jnp.complex64
     ldc = logdet_cmplx(A.astype(jnp.complex64))
     assert ldc.dtype == jnp.complex64
-    np.testing.assert_allclose(ld, ldc, rtol=1e-6)
+    np.testing.assert_allclose(ld, ldc, rtol=1.5e-6)
```
gcarleo, line 39: Convenient adjustment! why?

inailuig: we change the rng implementation to the the partitionable one, 
and the tolerance in this test is so tight that it fails even if you just change the random seed:


```python
import netket as nk
import jax
import jax.numpy as jnp
import numpy as np    
from netket.jax import logsumexp_cplx, logdet_cmplx

#jax.config.update('jax_threefry_partitionable', True)
#k = jax.random.PRNGKey(1)

k = jax.random.PRNGKey(4)


A = jax.random.normal(k, (3, 4, 5, 5), dtype=jnp.float32)
ld = logdet_cmplx(A)
assert ld.shape == (3, 4)
assert ld.dtype == jnp.complex64
ldc = logdet_cmplx(A.astype(jnp.complex64))
assert ldc.dtype == jnp.complex64
np.testing.assert_allclose(ld, ldc, rtol=1e-6)
```
This runs in single precision and the absolute error when it fails is around 1e-5

netket/sampler/metropolis.py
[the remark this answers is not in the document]
PhilipVinc: I'm ok with moving that logic to Metropolis. In the end, that is the only sampler that uses multiple chains..

netket/sampler/exact.py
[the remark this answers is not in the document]
PhilipVinc: can you add this as a comment?

netket/sampler/base.py
PhilipVinc: thanks

PhilipVinc:
@inailuig are we good to merge?

inailuig:
I just re-rand the distributed tests on 2 gpus, and they pass except for the numba wrapper, which seems to go into some sort of deadlock (I have seen this all along also with the shard map based wrapper I was using previously). We give a big warning if the wrapper is used, so I suppose this is ok for now.

PR #1515, opened 2023-06-29
Conversion of LocalOperator to PauliStrings

jwnys:
Implemented "LocalOperator.to_pauli_strings()" in the way I suggested in #1335


PhilipVinc (approved):
Thanks! Looks good. I just added a few more comments as it took me a bit to figure out what was going on.

PR #1559, opened 2023-08-05
Fix docstring to match ordering introduced in v3.9 (see #1502)

Z-Denis:
Everything is in the title.


PhilipVinc (approved):
Nice, thank you.

One day we should also add a new guide also on how to define a jax custom operator ...

PR #1567, opened 2023-08-11
Fix toml

gpescia:
For me performing netket tests as indicated in the documentation (`pytest -n auto test/hilbert`) causes the following error:
ERROR: usage: pytest [options] [file_or_dir] [file_or_dir] [...]
pytest: error: unrecognized arguments: -n -n auto test/operator

When leaving away the options -n auto I get:
ERROR: usage: pytest [options] [file_or_dir] [file_or_dir] [...]
pytest: error: unrecognized arguments: -n test/operator

Removing the options -n auto from the .toml file allows me to do the test using:
`pytest test`

### DIFF
--- a/netket/operator/_kinetic.py
+++ b/netket/operator/_kinetic.py
@@ -83,7 +83,7 @@ class KineticEnergy(ContinuousOperator):
         return self._is_hermitian
 
     def _expect_kernel_single(
-        self, logpsi: Callable, params: PyTree, x: Array, inverse_mass: Optional[PyTree]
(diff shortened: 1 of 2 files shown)


gpescia:
But here it is failing so maybe there is a problem with my local configuration?

inailuig:
@gpescia do you have  the `pytest-xdist` package installed in your environment?

I believe this was working as intended, and this PR should be closed.

gpescia:
You are right it is the missing pytest-xdist package. Strange it was not automatically installed. Thanks!

PR #1568, opened 2023-08-13
Require python 3.9

gcarleo:
Since Jax requires python 3.9, we should require it too. It also addresses [PEP 585](https://peps.python.org/pep-0585/ ).

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -33,13 +33,13 @@ jobs:
             oldest_deps: false
 
           - os: macos-latest
-            python-version: "3.8"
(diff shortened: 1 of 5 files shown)


PhilipVinc (approved):

PhilipVinc:
(I stopped testing netket on 3.8 because we had too many testers, but otherwise it should still work)

PhilipVinc:
(if we drop 3.8 we should add it to the changelog, like I did in the last version with 3.7)

gcarleo:
I already added it to the changelog

PR #1591, opened 2023-09-26
Rényi2 entanglement entropy

alleSini99:
Operator computing the Rényi2 entanglement entropy for `MCState` (using the SWAP estimator of Hastings et al., PRL 104, 157201) and for `FullSumState` on Hilbert spaces with discrete dofs.


fabienalet:
Hi,
thanks for the PR. Before having a look in the code, I have a quick question : would this work when ones enforces that the state has a fixed number of particles / magnetization ?
This is a typical situation where swap-based entanglement entropies give incorrect results if one is not careful to match the subsystem symmetry sectors.

gcarleo:
@fabienalet my understanding is that in that case it is correct to set wave function amplitudes to zero if the total number of particles after swap is not conserved? that should be correct

gcarleo:
I mean if I have  Psi(x1_a,x1_b) and after swap I have Psi(x1_a,x2_b) but N1_a + N2_b \ neq N, then you'd just use Psi(x1_a,x2_b)=0 in the estimator

PhilipVinc:
I think what Fabien is referring to is the case where you have two sublattices with different total number of conserved particles. If you swap configurations between the two you might not conserve the local number.

Indeed this code won't work there, as we do nothing for this case. It will only work fine for global conservation.

gcarleo:
ah yes, this would work only for global conservation indeed... for local conservation it's trickier, to be implemented if necessary

netket/experimental/observable/renyi2/expect.py
```diff
@@ -0,0 +1,90 @@
...
+    def kernel_fun(params, model_state, σ_ηp, σp_η, σ_η, σp_ηp):
+        W = {"params": params, **model_state}
+
+        return jnp.exp(afun(W, σ_ηp) + afun(W, σp_η) - afun(W, σ_η) - afun(W, σp_ηp))
+
+    kernel_values = kernel_fun(params, model_state, σ_ηp, σp_η, σ_η, σp_ηp)
+
+    Renyi2_stats = mpi_statistics(kernel_values.reshape((n_chains, -1)).T)
+
+    Renyi2_stats = Renyi2_stats.replace(
```
gcarleo, line 78: can you add comments on what this stuff is doing? is it propagating errors?

netket/vqs/base.py
```diff
@@ -304,10 +305,10 @@ def to_qobj(self):  # -> "qutip.Qobj"
         Returns:
             A :class:`qutip.Qobj` object.
         """
-        from qutip import Qobj
+        qutip = import_optional_dependency("qutip", descr="to_qobj")
```
gcarleo, line 308: looks like there is several unrelated commits

PhilipVinc:
To set the wavefunction to zero we would need a simple/cheap way to test if a basis element is in the correct symmetry subsector / computational basis.

We have this functionality (hidden away) for standard Hilbert spaces but I don't think we expose it for tensor Hilbert.
Wouldn't be hard to implement but this has to be implemented first.

In the meantime, is there an easy way to error if we have a tensorhilbert of constrained spaces?

netket/vqs/base.py
PhilipVinc: That's me adding it. The full summation code uses qutip (otherwise we'd have to implement a partial trace by hand) and I wanted a clean error message. I can get it this commit in a separate PR if you prefer.

netket/experimental/observable/renyi2/expect.py
alleSini99: It is propagating the error on a random variable x to logx.

gcarleo: yes please add a comment in the code

netket/vqs/base.py
gcarleo: it's fine

gcarleo:
Also, I wonder if this code is already compatible with continuos space?

fabienalet:
> I think what Fabien is referring to is the case where you have two sublattices with different total number of conserved particles. If you swap configurations between the two you might not conserve the local number.
> 
> Indeed this code won't work there, as we do nothing for this case. It will only work fine for global conservation.

I was referring to the two cases in fact. 
Yes for global conservation putting by hand psi to 0 is enough (or implement that hard in the entropy estimator, with a simple test).
For local conservation, this can become very tricky and writing up something general might become cumbersome.
I guess for most users, ensuring that the swap estimator takes care of the global conservation rule should be enough

PhilipVinc:
But isn't the global conservation automatically enforced? If the samples are valid (which they must be, as we assume it throughout netket) then swapping two sites won't break the global quantity?

gcarleo:
No, you might still compute the wave function on a state with wrong total
number of particles / spin, since you swap subsets of the lattice

gcarleo:
The easiest thing would be to add a check that the configuration after swap
is in the Hilbert space I think

PhilipVinc:
I'm talking about global population constraints (eg total magnetisation / n bosons). If you swap configurations between two partitions the global constraint is still respected. No?

PhilipVinc:
Ah, no, sorry. You're right. It's two different chains.

gcarleo:
No, because you swap subsets of two independent configurations

netket/experimental/observable/renyi2/expect.py
```diff
@@ -0,0 +1,91 @@
...
+    params,
+    model_state,
+    samples,
+    subsystem,
+):
+    N = samples.shape[-1]
+    n_chains = int(samples.shape[0] / 2)
+
+    σ_η = samples[:n_chains]
+    σp_ηp = samples[n_chains:]
```
PhilipVinc, line 53: This won't work well. Try for example to use this with `ExactSampler` and it will crash. @alleSini99 you need to reshape the samples to a 2D matrix and then just partition those

fabienalet:
What I did by hand long time ago was very simple : compute the (number of particles / magnetization) in subsystem A in configuration 1 and configuration 2 (from thw two independent chains), if not the same then don't bother doing the swap. This assumes you know which symmetry you want to enforce of course

gcarleo:
@PhilipVinc it seems we don't have a function in Hilbert to check if a configuration is valid, maybe we should add it?

PhilipVinc:
Absolutely

fabienalet:
Since we're on the topic on symmetries and Renyi entanglement entropies, another thing which can be tricky is the following : most of the time the entanglement cut (which degrees of freedom are in A or B) (1) does not preserve the lattice symmetries, but (2) sometimes some symmetries are preserved. I don't think (1) causes problems in evaluating Renyi entropy for symmetrized NQS (e.g. RBMSymm or GCNN) as implemented in NetKet, but maybe (2) could be leveraged to improve computations.

netket/experimental/observable/renyi2/expect.py
alleSini99: Ok, I have also added a test for it

PhilipVinc (approved):
Thank you @alleSini99 for bearing with me.
I added a check that there are no constraints, in order to error informatively if there are.

If at some point someone contributes the logic to Hilbert to check validity of configurations then we can implement the logic for constrained spaces as well in here.

My only question is the following: should we rename `subsystem` to `partition`? that way we could use the same argument for a geometry object to work in continuous space in the future (@gcarleo )

gcarleo:
Yes and the partition should be specified by a Graph object already in this
PR, os that it's general enough / we can use also a Geometry object to
specify the partition

PhilipVinc:
how do you define a partition with a graph object? A partition is a list of vertices, not a graph

gcarleo:
A list of vertices is a graph

PhilipVinc:
@alleSini99 can you puoi change the name of `subsystem ` to   partition?

gcarleo:
Another argument in favor of using a Graph is that if you want to specify symmetries for the partition you cannot have only vertices, you need more info

PhilipVinc:
I understand the spirit. But I still don't think the object you are talking about is a graph.

Case in point, the symmetries you are talking about are the symmetries induced by th underlying graph on th partitioning.

You need the underlying graph for the symmetries and a list of vertices to define the partitioning, and then you can augment that by turning it into a set of partitionsthat should be equivalent due to simmetries.

--
Filippo Vicentini
Assistant Professor (Ecole Polytechnique, Paris)
Scientist (EPFL, Lausanne)
Google Scholar, GitHub

+41 (0)77 238 85 68
+33 (0)7 83 93 08 23

Il 27 set 2023, 8:39 PM +0200, Giuseppe Carleo ***@***.***>, ha scritto:

PhilipVinc:
I guess you want a colored graph, where the two colors represent the  two parti ions. But then we want to compute the different possible colorings that respect some underlying simmetries , which is not something that we have already implemented either
Il 27 set 2023, 8:39 PM +0200, Giuseppe Carleo ***@***.***>, ha scritto:

gcarleo:
ok maybe it's too complex, since this is in experimental I am ok to merge it as it is, in any case the continous space version will have different requirements, it's best to merge this and then think later on if we can merge interfaces

fabienalet:
> Another argument in favor of using a Graph is that if you want to specify symmetries for the partition you cannot have only vertices, you need more info

I think this is the correct interpretation. Indeed a partition is only a collection of sites, but if the reduced density matrix is symmetric under operations on this set of sites, then this symmetry is better interpreted as a lattice symmetry on a graph formed by these sites. In some cases this graph is just the sub-graph obtained by removing all the bonds / interactions connecting these sites to the ones in the other partition (e.g. a graph : horizontal ladder --> subgraph : a chain if you consider all the sites in a leg as your partition). In some other cases, there is no symmetry and there is no subgraph (e.g.  graph = horizontal  ladder --> subgraph : a set of independent "dimers"  if you consider the partition with all sites on the left rungs of the ladder). It is indeed hard to make generic statements and codes. The best way to keep options open is indeed to have an optional parameter which gives the partition graph besides the list of sites.
But I also agree that this is not essential, and that conservation of magnetization/number of particles is much more important to take into account (as this can easily give wrong results)

PR #1602, opened 2023-10-03
create fermionic operator from pyscf molecule

inailuig:
requires the [sparse](https://github.com/pydata/sparse) package

example:
```python
from pyscf import gto, scf, fci
from netket.experimental.operator.pyscf import FermiOperator2nd_from_pyscf_molecule
import netket as nk
import numpy as np

bond_length = 1.5109
basis='STO-3G'
geometry = [('Li', (0., 0., -bond_length/2)), ('H', (0., 0., bond_length/2))]
mol = gto.M(atom=geometry, basis=basis)

mf = scf.RHF(mol).run()
E_hf = sum(mf.scf_summary.values())

E_fci = fci.FCI(mf).kernel()[0]
print('fci', E_fci)

ha = FermiOperator2nd_from_pyscf_molecule(mol)
E0 = float(nk.exact.lanczos_ed(ha))
assert np.allclose(E0, E_fci)
print(E0)
```

### DIFF
--- a/netket/experimental/operator/pyscf.py
+++ b/netket/experimental/operator/pyscf.py
@@ -1,6 +1,19 @@
+# Copyright 2023 The NetKet Authors - All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 from typing import Optional
 
-import sparse
 import numpy as np
 
 import jax
@@ -25,6 +38,8 @@ def compute_pyscf_integrals(mol, mo_coeff):
 
 
 def spinorb_from_spatial_sparse_coo2(tij_sparse, interleave=False, spin_values=[0, 1]):
+    sparse = import_optional_dependency("sparse", descr="TV_from_pyscf_molecule")
+
     # Σ_ijσ t_ij c†_iσ c_jσ
     # for σ ∈ spin_values
     # interleave=True -> 2i+spin
@@ -66,6 +81,8 @@ def spinorb_from_spatial_sparse_coo2(tij_sparse, interleave=False, spin_values=[
 def spinorb_from_spatial_sparse_coo4(
     vijkl_sparse, interleave=False, _order_preserving=False
 ):
+    sparse = import_optional_dependency("sparse", descr="TV_from_pyscf_molecule")
+
     # Σ_ijklμσ v_ijkl c†_iμ c†_jσ c_kμ c_lσ
     # interleave=True -> 2i+spin
     # interleave=False -> n*spin+i
@@ -103,6 +120,8 @@ def spinorb_from_spatial_sparse_coo4(
 
 
 def to_desc_order_sparse(vijkl_sparse, cutoff, set_zero_same=True):
+    sparse = import_optional_dependency("sparse", descr="TV_from_pyscf_molecule")
+
     # !! use this only after adding spin, spinorb_from_spatial_sparse will be wrong
     # because the (implicitly assumed) symmetries of the tensor are not conserved
 
@@ -235,6 +254,8 @@ def TV_from_pyscf_molecule(
         ...
 
     """
+    sparse = import_optional_dependency("sparse", descr="TV_from_pyscf_molecule")
+
     # Compute the T and V matrices
     E_nuc, tij, vijkl = compute_pyscf_integrals(molecule, mo_coeff)
     tij_sparse = sparse.COO.from_numpy(tij * (np.abs(tij) >= cutoff))
--- a/netket/experimental/operator/pyscf.py
+++ b/netket/experimental/operator/pyscf.py
@@ -135,9 +135,9 @@ def spinorb_from_spatial_sparse(tij_sparse, vijkl_sparse, interleave=False):
 
 def pyscf_molecule_to_arrays(mol, mo_coeff=None):
     if mo_coeff is None:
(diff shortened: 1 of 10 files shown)


PhilipVinc:
(This would require a few extra comments, because the code is really complex, and some tests. 

- Also, can we agree on the signature of the function to present to the user?

- Ah, why is it accepting a class argument? is it because you want to be able to use it with your custom classes? I'm not against it, just to try to understand...)

I am not against adding `sparse`. It depends only on numba+numpy+scipy which we already carry as dependencies, it's a pure python package and it's relatively lightweight, meaning that it should not create many additional problems down the line.

But the question stands.. why do you need it?

PhilipVinc:
By the way if we un-transpose FermionOperators later do we break this code?

PhilipVinc:
Ok. I'm good with having the extra dependency. We can add it as an extra dependency `[pyscf]` with `sparse`and `pyscf`.

```python
def FermiOperator2nd_from_pyscf_molecule(mol, 
mo_coeff=None, 
cutoff=1e-11, 
cls=FermionOperator2nd
)
```

What is `mo_coeff`?

My main argument is whether we want this called as it is, or have `FermiOperator2nd.from_pyscf_molecule(...)`as a class method (which incidentally would solve the class argument as well, in a sense.

I don't like class methods, because I think they are hard to 'discover', so I would be more in line for having it as a second function. 
However, we have `FermiOperator2nd.from_openfermion` as well, so we need some sort of consistency.

Mainly, if we go for it being a free function, I dislike the name being so long. Can't we make it simply `nk(x).operator.from_pyscf_molecule(...)`?

gcarleo:
yes ok... so, indeed it shoudl be a method of `FermionOperator2nd` for consistency with all the other cases (also not sure why you are passing `cls=FermionOperator2nd` ? are there other possible classes you can convert to?)

inailuig:
> Can't we leave these details to pyscf only? Or is there a reason why pyscf cannot give the Mol object with the Boys, for example?

afaik the mol object itself only knows about atomic orbitals, so there is no way to avoid passing it separately

PhilipVinc:
>afaik the mol object itself only knows about atomic orbitals, so there is no way to avoid passing it separately

well then, we have no choice.

> Default hartree fock.

@gcarleo is this a good default? I think here we really want users to DIY. I think we should just document it with an example probperly.

> I had thought about that, but I think the counter argumentis that FermionOperator2nd (and also the openfermion one) are really general fermionic operators, whereas the 2nd quantized fermionic hamiltonian has a very specific structure for which you can write a dedicated more efficient implementation towards which we should push users.

So it would make sense to have this thing return the best format by default and one can override... indeed.
Well, this could go in experimental (and stay for the time being) in `nkx.operator.from_pyscf(molecule, orbitals, implementation=..)`

PhilipVinc:
>afaik the mol object itself only knows about atomic orbitals, so there is no way to avoid passing it separately

well then, we have no choice.

> Default hartree fock.

@gcarleo is this a good default? I think here we really want users to DIY. I think we should just document it with an example probperly.

> I had thought about that, but I think the counter argumentis that FermionOperator2nd (and also the openfermion one) are really general fermionic operators, whereas the 2nd quantized fermionic hamiltonian has a very specific structure for which you can write a dedicated more efficient implementation towards which we should push users.

So it would make sense to have this thing return the best format by default and one can override... indeed.
Well, this could go in experimental (and stay for the time being) in `nkx.operator.from_pyscf(molecule, orbitals, implementation=..)`

gcarleo:
I would have a minimal general interface where you pass the one body and
the two body tensor and that's it. That should be the general
ChemistryHamiltonian or something ElectronicHamiltonian, then for pyscf I
am still not sure we needed to pass more stuff than mol, but OK... It
doesn't hurt if by default it's None

PhilipVinc:
Those are two separate functions. I want the `from_pyscf` function because that's a common use case. Internally we should have the function working on V and T. we can discuss whether to expose them.

gcarleo:
OK agreed, but I suspect that the function converting from pyscf already converts to a standard T+V hamiltonian in second quantization? if it is the case, I would also expose that already with a suitable name, and transform the pyscf converter into something that simply generates the T and V tensors that make sense for our internal implementation. Our main hamiltonian should be something like `ElectronicHamiltonian` or `2BodyFermionic` or similar (I am just opposed maybe to calling it `MolecularHamiltonian` because it's too restrictive) that takes the T and V tensors, regardless of how they are generated / from what interactions they come from

netket/experimental/operator/pyscf.py
```diff
@@ -0,0 +1,226 @@
+import sparse
+import numpy as np
+
+import jax
+import jax.numpy as jnp
+
```
inailuig, line 6:
```suggestion

from netket.utils.optional_deps import import_optional_dependency
```

netket/experimental/operator/pyscf.py
```diff
@@ -0,0 +1,226 @@
...
+def spinorb_from_spatial_sparse(tij_sparse, vijkl_sparse, interleave=False):
+    return spinorb_from_spatial_sparse_coo2(
+        tij_sparse, interleave
+    ), spinorb_from_spatial_sparse_coo4(vijkl_sparse, interleave)
+
+
+def pyscf_molecule_to_arrays(mol, mo_coeff=None):
+    if mo_coeff is None:
+        from pyscf import scf
```
inailuig, line 137:
```suggestion
        pyscf = import_optional_dependency("pyscf", descr="pyscf_molecule_to_arrays")
```

netket/experimental/operator/pyscf.py
```diff
@@ -0,0 +1,226 @@
...
+
+        mf = scf.HF(mol).run()
```
inailuig, line 139:
```suggestion
        mf = pyscf.scf.HF(mol).run()
```

PhilipVinc:
I cleaned this up a bit... Some more documentation is missing, still, and tests.
I would love this to be merged for 3.10

PhilipVinc (approved):
@inailuig unless you have something against it, I'm merging this for now. I added a few tests (more would be welcome) and cleaned up the documentation. Regardless, it's experimental so we can change things if we need in the future.

PR #1610, opened 2023-10-12
nkvjp: return a Partial

inailuig:
make it consistent with jax.vjp
useful if you want to return the vjp_fun from a jitted function (or from shard_map which is what I was trying to do)

### DIFF
--- a/netket/jax/_vjp.py
+++ b/netket/jax/_vjp.py
@@ -18,7 +18,7 @@ from functools import partial
 import jax
 
 from jax import numpy as jnp
-from jax.tree_util import tree_map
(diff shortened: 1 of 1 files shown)


netket/jax/_vjp.py
```diff
@@ -148,6 +152,9 @@ def vjp_fun(ȳ):
 
         return out
 
+    # TODO pass vals_r and vals_j dtype via HashablePartial?
+    vjp_fun = Partial(vjp_fun_rc, vjp_r_fun, vjp_j_fun)
```
inailuig, line 156: should we?

PR #1683, opened 2024-01-05
Add metropolis exchange rule for fermions

jwnys:
This sampler avoids proposing the same sample for fermions.
The previous approach was to use ExchangeSampler, which often applies an exchange between two occupied or two unoccupied fermionic modes.
Now we only consider clusters of one occupied and one unoccupied mode.
This should yield better acceptances.
Relevant for @inailuig and @imi-hub , and most specifically for the few-fermion case

### DIFF
--- a/Examples/Fermions/fermi_hubbard.py
+++ b/Examples/Fermions/fermi_hubbard.py
@@ -54,7 +54,7 @@ print("Hamiltonian =", ham.operator_string())
 
 # g.n_nodes == L*L --> disj_graph == 2*L*L
 disj_graph = nk.graph.disjoint_union(g, g)
-sa = nk.sampler.MetropolisExchange(hi, graph=disj_graph, n_chains=16)
+sa = nk.sampler.MetropolisFermionExchange(hi, graph=disj_graph, n_chains=16)
 
 # since the hilbert basis is a set of occupation numbers, we can take a general RBM
 # we take complex parameters, since it learns sign structures more easily, and for even fermion number, the wave function might be complex
--- a/Examples/Fermions/fermi_hubbard.py
+++ b/Examples/Fermions/fermi_hubbard.py
@@ -52,9 +52,12 @@ print("Hamiltonian =", ham.operator_string())
 # move the fermions around independently for both spins
 # and therefore conserve the number of fermions with up and down spin
 
+# we can do this explicitly
 # g.n_nodes == L*L --> disj_graph == 2*L*L
 disj_graph = nk.graph.disjoint_union(g, g)
-sa = nkx.sampler.MetropolisFermionExchange(hi, graph=disj_graph, n_chains=16)
+sa = nkx.sampler.MetropolisFermionExchange(hi, graph=g, n_chains=16)
+# or let netket copy the graph per spin sector
+sa = nkx.sampler.MetropolisFermionExchange(hi, graph=g, n_chains=16, copy_per_spin=True)
 
 # since the hilbert basis is a set of occupation numbers, we can take a general RBM
 # we take complex parameters, since it learns sign structures more easily, and for even fermion number, the wave function might be complex
--- a/Examples/Fermions/fermi_hubbard.py
+++ b/Examples/Fermions/fermi_hubbard.py
@@ -58,7 +58,7 @@ disj_graph = nk.graph.disjoint_union(g, g)
 sa = nkx.sampler.MetropolisParticleExchange(hi, graph=g, n_chains=16, sweep_size=64)
 # or let netket copy the graph per spin sector
 sa = nkx.sampler.MetropolisParticleExchange(
-    hi, graph=g, n_chains=16, sweep_size=64, exhange_spins=False
+    hi, graph=g, n_chains=16, sweep_size=64, exchange_spins=False
 )
 
 # since the hilbert basis is a set of occupation numbers, we can take a general RBM
--- a/Examples/Fermions/fermi_hubbard.py
+++ b/Examples/Fermions/fermi_hubbard.py
@@ -15,7 +15,7 @@ g = nk.graph.Hypercube(length=L, n_dim=D, pbc=True)
 n_sites = g.n_nodes
 
 # create a hilbert space with 2 up and 2 down spins
-hi = nkx.hilbert.SpinOrbitalFermions(n_sites, s=1 / 2, n_fermions=(2, 2))
+hi = nkx.hilbert.SpinOrbitalFermions(n_sites, s=1 / 2, n_fermions_per_spin=(2, 2))
 
 
 # create an operator representing fermi hubbard interactions
@@ -54,7 +54,7 @@ print("Hamiltonian =", ham.operator_string())
 
 # g.n_nodes == L*L --> disj_graph == 2*L*L
 disj_graph = nk.graph.disjoint_union(g, g)
-sa = nk.sampler.MetropolisFermionExchange(hi, graph=disj_graph, n_chains=16)
+sa = nkx.sampler.MetropolisFermionExchange(hi, graph=disj_graph, n_chains=16)
 
 # since the hilbert basis is a set of occupation numbers, we can take a general RBM
 # we take complex parameters, since it learns sign structures more easily, and for even fermion number, the wave function might be complex
@@ -63,7 +63,7 @@ vs = nk.vqs.MCState(sa, ma, n_discard_per_chain=100, n_samples=512)
 
 # we will use sgd with Stochastic Reconfiguration
 opt = nk.optimizer.Sgd(learning_rate=0.01)
-sr = nk.optimizer.SR(diag_shift=0.1)
+sr = nk.optimizer.SR(diag_shift=0.1, holomorphic=True)
 
 gs = nk.driver.VMC(ham, opt, variational_state=vs, preconditioner=sr)
 
--- a/Examples/Fermions/fermi_hubbard.py
+++ b/Examples/Fermions/fermi_hubbard.py
@@ -55,12 +55,10 @@ print("Hamiltonian =", ham.operator_string())
 # we can do this explicitly
 # g.n_nodes == L*L --> disj_graph == 2*L*L
 disj_graph = nk.graph.disjoint_union(g, g)
-sa = nkx.sampler.MetropolisParticleExchange(
-    hi, graph=g, n_chains=16, sweep_size=64, copy_per_spin=False
-)
+sa = nkx.sampler.MetropolisParticleExchange(hi, graph=g, n_chains=16, sweep_size=64)
 # or let netket copy the graph per spin sector
 sa = nkx.sampler.MetropolisParticleExchange(
-    hi, graph=g, n_chains=16, sweep_size=64, copy_per_spin=True
+    hi, graph=g, n_chains=16, sweep_size=64, exhange_spins=False
 )
 
 # since the hilbert basis is a set of occupation numbers, we can take a general RBM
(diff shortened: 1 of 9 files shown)


Examples/Fermions/fermi_hubbard.py
```diff
@@ -54,7 +54,7 @@ def nc(site, sz):
 
 # g.n_nodes == L*L --> disj_graph == 2*L*L
 disj_graph = nk.graph.disjoint_union(g, g)
-sa = nk.sampler.MetropolisExchange(hi, graph=disj_graph, n_chains=16)
+sa = nkx.sampler.MetropolisFermionExchange(hi, graph=disj_graph, n_chains=16)
```
gcarleo, line 57: since we are doing a specific sampler for fermions (which is very much needed, thanks!) why don't we make it such we don't need to do anymore this hack of the disjoint union?

netket/experimental/sampler/rules/fermion_2nd.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+class FermionExchangeRule(ExchangeRule):
+    """Exchange rule for fermions.
+
+    Works similarly to exchange rule, but:
+    takes into account that only occupied orbitals
+    can be exchanged with unoccupied ones.
+    """
+
```
PhilipVinc, line 13: I would include this in here, as it will be the standard use-case.

We can also ask the Hilbert space in here in order to make sure we get the right number of spin values... I dont know if this makes sense, but it seems to me it might?

```suggestion

    def __init__(
        self,
        hilbert, 
        *,
        clusters: Optional[list[tuple[int, int]]] = None,
        graph: Optional[AbstractGraph] = None,
        d_max: int = 1,
        copy_per_spin: bool = True,
    ):
        r"""
        Constructs the Exchange Rule.

        You can pass either a list of clusters or a netket graph object to
        determine the clusters to exchange.

        Args:
            clusters: The list of clusters that can be exchanged. This should be
                a list of 2-tuples containing two integers. Every tuple is an edge,
                or cluster of sites to be exchanged.
            graph: A graph, from which the edges determine the clusters
                that can be exchanged.
            d_max: Only valid if a graph is passed in. The maximum distance
                between two sites
            copy_per_spin: (default True) If True, the graph must encode the
                connectivity  between the first N physical sites having same spin, and
                it is replicated using :ref:`netket.graph.disjoint_union` other every
                spin subsector. This option conserves the number of fermions per
                spin subsector. 
        """
        if copy_per_spin and hilbert.n_spin_subsectors > 1:
            if graph is not None and graph.n_nodes == hilbert.n_orbitals:
                 graph = disjoint_union(*[graph] * hilbert.n_spin_subsectors)
            if clusters is not None and np.max(clusters) < hilbert.n_orbitals:
                clusters = np.concatenate(
                    [
                        clusters + i * hilbert.n_orbitals
                        for i in range(hilbert.n_spin_subsectors)
                    ]
                )
        super().__init__(clusters=clusters, graph=graph, d_max=d_max) 

```

PhilipVinc (commented):
My only doubt is that maybe it would be nice to keep the spin flag in the rule itself. 
The samplers object are usually just wrappers for writing more Terse code...

what do you think?

netket/experimental/sampler/metropolis.py
```diff
@@ -0,0 +1,43 @@
+from netket.sampler import MetropolisSampler
+from netket.experimental.hilbert import SpinOrbitalFermions
+from netket.graph import disjoint_union
+import numpy as np
+
+
+def MetropolisFermionExchange(
+    hilbert, *, clusters=None, graph=None, d_max=1, copy_per_spin=True, **kwargs
```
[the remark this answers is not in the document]
jwnys: I'm open to suggestions for the naming.

jwnys:
Two main changes:
- change default dtype to integers (exchange gives float64).
- I renamed the sampler to a more general "ParticleExchange". After talking to @Z-Denis , the same sampler can be used for bosons as well (after generalizing the way hoppable_clusters are identified). I didn't allow for bosons yet, because it seems that there is no such thing as SpinOrbitalBosons, or similar. I'll leave this up to @Z-Denis  for bosons. Nothing needs to be changed, it's just not clear to me what to do with the spin checks for that. What do you think @PhilipVinc @gcarleo ?

PhilipVinc (approved):
> change default dtype to integers (exchange gives float64).

I'm all in favour for this change. I actually think we should use this everywhere...

I'd just like to point out that netket's operators (and everything else) is not throughly tested for integer sample dtypes, so be on the lookout for problems that might appear elsewhere, like QGT.

> I renamed the sampler to a more general "ParticleExchange". 

Good idea! I have the code to do this for bosons somewhere... It applies to Spin->=1 spaces as well.

> it's just not clear to me what to do with the spin checks for that

can be done in a follow up PR by someone else.

--

So the only remaining thing for me is whether we keep `copy_per_spin` or rename it. Otherwise this can be merged.

netket/experimental/sampler/rules/fermion_2nd.py
```diff
@@ -0,0 +1,125 @@
...
+    def __init__(
+        self,
+        hilbert,
+        *,
+        clusters: Optional[list[tuple[int, int]]] = None,
+        graph: Optional[AbstractGraph] = None,
+        d_max: int = 1,
+        copy_per_spin: bool = True,
```
PhilipVinc, line 35: Actually @jwnys , would it make sense to default this to `None` and have it default to
```python
if copy_per_spin is None:
    copy_per_spin = all(n is not None for n in hilbert.n_fermions_per_spin)
```
?

gcarleo:
Yes I also think copy is not the right word here, can we have something
like swap_spins or similar?

PhilipVinc:
Same spin, rather...

jwnys:
Hm, then I'd rather like to keep it more explicit that we are copying the interaction graph per spin sector..

netket/experimental/sampler/rules/fermion_2nd.py
```diff
@@ -1,15 +1,79 @@
...
     can be exchanged with unoccupied ones.
+    This sampler conserves the number of particles.
+    """
+
+    copy_per_spin: bool
     """
+    Whether to copy the graph for each spin sector.
+    This only occurs if cluster indices or number of graph nodes
+    are lower than the number of orbitals per spin sector.
+    """
```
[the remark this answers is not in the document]
jwnys: You're right, I'll remove it

jwnys:
Maybe "spin_symmetric" then?

PhilipVinc:
But is this simmetry? isn't it more invariant? Maybe `replicate_spin_subsectors` or `identical_spin_subsectors` ?

jwnys:
let's maybe just keep the copy for now then?

gcarleo:
What does the flag do exactly? It disallows exchanging different spins so that the spin populations are constant ? If so, I'd not use copy it's really not clear what it means. Just call it exhange_spins=False or something?

jwnys:
> exhange_spins

Yeah, it's just that I rather copy the graph per spin, so I'm not removing any spin exchanges if you were to pass a graph that would do that.

gcarleo:
I see... Then maybe can we impose the graph size is the same as the number of sites? (not twice as much, which is a hack and should not be used)

jwnys:
> I see... Then maybe can we impose the graph size is the same as the number of sites? (not twice as much, which is a hack and should not be used)

Alternatively, I just mention explicitly in the docs that i this flag only has an impact if the number of sites = number of orbitals..

PhilipVinc:
> Yeah, it's just that I rather copy the graph per spin, so I'm not removing any spin exchanges if you were to pass a graph that would do that

Though right now it would crash, because you'd end up with a graph with more connections.
Crashing is kind-of like saying it does not work in other cases.

> exhange_spins

I like this. It could be that it can be used only if you have conserved number per spin, and you cannot specify it otherwise...

netket/experimental/sampler/metropolis.py
```diff
@@ -0,0 +1,40 @@
...
+    **kwargs,
+) -> MetropolisSampler:
+    r"""
+    This sampler moves (or hops) a random particle to a different but random empty mode.
+    It works similar to MetropolisExchange, but only allows exchanges between occupied and unoccupied modes.
+
+    Args:
+        hilbert: The Hilbert space to sample.
+        d_max: The maximum graph distance allowed for exchanges.
+        exhange_spins: If False, exchanges are only allowed between modes with the same spin projection.
```
PhilipVinc, line 22:
```suggestion
        exhange_spins: (default False) If False, exchanges are only allowed between modes with the same spin projection.
```

netket/experimental/sampler/rules/fermion_2nd.py
```diff
@@ -0,0 +1,118 @@
...
+from netket.sampler.rules import ExchangeRule
+from netket.graph import AbstractGraph
+from netket.graph import disjoint_union
+from netket.experimental.hilbert import SpinOrbitalFermions
+
+
+class ParticleExchangeRule(ExchangeRule):
+    """Exchange rule for particles on a lattice.
+
+    Works similarly to ExchangeRule, but:
```
PhilipVinc, line 15:
```suggestion
    Works similarly to :class:`netket.sampler.rules.ExchangeRule`, but:
```

netket/experimental/sampler/rules/fermion_2nd.py
```diff
@@ -0,0 +1,118 @@
...
+    takes into account that only occupied orbitals
+    can be exchanged with unoccupied ones.
+    This sampler conserves the number of particles.
```
PhilipVinc, line 18:
```suggestion

    This sampler conserves the number of particles.
```

docs/api/sampler.md
```diff
@@ -83,6 +83,7 @@ This is a list of shorthands that allow to construct a {class}`~netket.sampler.M
 
    MetropolisLocal
    MetropolisExchange
+   MetropolisParticleExchange
```
PhilipVinc, line 86: those go in `api/experimental.md`, not here. this will break the docs build

docs/api/sampler.md
```diff
@@ -117,6 +118,7 @@ Sampler. Rules with `Numpy` in their name can only be used with
   rules.LocalRule
   rules.CustomRuleNumpy
   rules.ExchangeRule
+  rules.ParticleExchangeRule
```
PhilipVinc, line 121: You could also add below a block like the following, which duplicates a bit the docs but maybe would help searchability

```
## Experimental models

The following samplers are experimental, meaning that we could change them at some point, and we actively seeking for feedback and opinions on their usage and APIs.


### Particle-specific samplers

The following models are for 2nd-quantisation fermionic hilbert spaces ({class}`netket.experimental.hilbert.SpinOrbitalFermions`).

```{eval-rst}
.. currentmodule:: netket.experimental.sampler

```

```{eval-rst}
.. autosummary::
   :toctree: _generated/samplers
   :template: flax_module_or_default
   :nosignatures:


   MetropolisParticleExchange
```

And the corresponding rules
```{eval-rst}
.. autosummary::
   :toctree: _generated/samplers
   :template: flax_module_or_default
   :nosignatures:


   rules.ParticleExchangeRule
```

netket/experimental/sampler/rules/fermion_2nd.py
```diff
@@ -0,0 +1,119 @@
...
+        return f"ParticleExchangeRule(# of clusters: {len(self.clusters)})"
+
+
+@jax.jit
+def _compute_hoppable_clusters_mask(clusters, σ):
+    # mask the clusters to include only feasible moves (occ -> unocc, or the inverse)
+    hoppable_clusters_mask = ~jnp.isclose(
+        σ[..., clusters[:, 0]], σ[..., clusters[:, 1]]
+    )
+    return hoppable_clusters_mask
```
PhilipVinc, line 119: Now that I think about it, this condition makes sense for Exchange Rule itself, because it simply asks that it does not swap two identical configurations.

Why can't this be used in the standard Exchange Rule? (cc @gcarleo )

PhilipVinc: Ok.. Maybe we can do that in a follow up PR.

PR #1693, opened 2024-01-17
Substitute `variables` with `parameters` in automatic QGT selector

alleSini99:
In the function `default_qgt_matrix`, which determines the default metric tensor depending on the variational state and the solver, it is used `variables` as argument of `nkjax.tree_ishomogeneous`. I think this should be substituted with `parameters`, since in general `variables` can contain quantities that are not numbers and are not differentiated (for instance an operator U if we want to sample from U|psi> where |psi> is an arbitrary state).


PhilipVinc (approved):
Nice catch

PR #1746, opened 2024-03-05
Remove unused sample_fun from MCState

inailuig:
I tried to use this to do some importance sampling without success,
because it has never been plugged in (since the release of nk 3).

Remove it for now, and consider adding it back later in some derived class.


PhilipVinc (approved):
yes, you're right.

PR #1750, opened 2024-03-06
Reset struct.Pytree property_cached

inailuig:


PhilipVinc (approved):
should be like this, otherwise the data class decorated ones do not benefit from it.

PR #1812, opened 2024-05-21
support non-multiple chunk_size and has_aux in vjp_chunked

inailuig:


PhilipVinc:
What we need this for ? I'm not against it but I think it will increase considerably the complexity of this already impeerscrutable code?

PR #1846, opened 2024-06-20
Make doubled spaces more easily indexable

jwnys:
Minor fix: make doubled spaces more easily indexable for certain low-particle-number fock spaces.


PhilipVinc (approved):
Can you just add a quick test with a case that is fixed by this PR to make sure it does not regress in the future?

Good otherwise.

PR #1854, opened 2024-07-04
Allow `machine_pow` to be any positive number

attila-i-szabo:
There doesn't seem to be a good reason to require `machine_pow` in samplers to be an integer. This PR replaces the checks to allow it to be any positive real number.

I don't think this affects any of the tests. Do we need additional ones?


PhilipVinc (approved):
Good point. 

I just want the check to pass if we create samplers in a jax.jit context, so the imaginary check should be on the dtype not on the value, and the positivity check (which we were not doing before) should only be done if it is not a tracer.

I pushed a commit to fix this, it's good for me now.

Can you just write a line in the changelog? Then I will merge.

attila-i-szabo:
Thanks for the fixes! I've added the changelog and am happy to merge.

PR #1860, opened 2024-07-11
Add line in changelog for PR 1687

alleSini99:
Add line  in changelog for variance observable PR 1687


PhilipVinc (approved):
usually you add it at the bottom of the list, but since we love you, well accept it at the top

PR #1892, opened 2024-08-15
Deprecated `rcond` in solvers

Daniel-Haas-B:
First PR here, so please let me know if anything needs adjustment. The idea was to address #1871. I've followed the contribution guidelines, and all tests are passing.

**Comments:**
- `linalg.lstsq` still utilizes `rcond`, so `solvers.svd` was not changed.
- In alignment with other deprecated examples in the project, I did not include the deprecated argument in the docstrings.
- I made an adjustment to the ising1d_sr_pinv example to prevent any warnings from being generated.

----------------
**Edit:** decided to disregard item 2 above since upon checking, the documentation page would load without any explanation of the deprecated arguments.

### DIFF
--- a/Examples/Ising1d/ising1d_sr_pinv.py
+++ b/Examples/Ising1d/ising1d_sr_pinv.py
@@ -37,7 +37,7 @@ op = nk.optimizer.Sgd(learning_rate=optax.linear_schedule(0.1, 0.0001, 500))
 
 # SR
 sr = nk.optimizer.SR(
-    solver=partial(nk.optimizer.solver.pinv, rcond=1.0e-6), diag_shift=0.01
+    solver=partial(nk.optimizer.solver.pinv, rtol=1.0e-6), diag_shift=0.01
 )
 
 # Variational state
--- a/netket/optimizer/solver/solvers.py
+++ b/netket/optimizer/solver/solvers.py
@@ -79,6 +79,8 @@ def pinv_smooth(
         rtol_smooth : Regularization parameter used with a similar effect to `rtol`
             but with a softer curve. See :math:`\epsilon` in the formula
             above.
+        rcond: (deprecated) Alias for `rtol`. Will be removed in a future release.
+        rcond_smooth: (deprecated) Alias for `rtol_smooth`. Will be removed in a future release.
     """
     del x0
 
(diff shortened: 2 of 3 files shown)


PhilipVinc (commented):
Hi @Daniel-Haas-B, sorry for the delay but everyone's on holiday break...

Overall the PR is good (thanks!) however there are 2 minor issues that should be addressed in order to merge it:

 - NetKet supports several jax versions (because installing the last one is sometimes complicated or unsupported on some GPUs), with the minimum version nowadays being 0.4.16, while the new `rtol` argument has been introduced between jax `0.4.29` and 0.4.30.  We need therefore to support both old and new jax versions. The solution for this would be to do something like
   ```python
     # TODO: Cleanup when we support only jax 0.4.29 or later
     if module_version(jax) >= (0,4,29):
         A_inv = jnp.linalg.pinv(A, rtol=rtol, hermitian=True)
     else:
         A_inv = jnp.linalg.pinv(A, rcond= rtol, hermitian=True)
    ```
    (you can import `module_version` from netket.utils`). 
    This way both old and new jax versions will work, but users of recent netket versions will always be pushed to use the newer rtol keyword argument.

 - We need a test to check that both new and old arguments work. Something easy that checks with `pytest.warns` that if you use the old syntax a warning is raised but the result is identical to calling the function with `rcond` would be enough.

Is this feasible?

Daniel-Haas-B:
I have implemented the suggestions! 

On a side note, I have noticed that `utils.warn_deprecation` uses category `FutureWarning`. Some other deprecation warnings such as the ones in `SpinOrbitalFermions` are using a `DeprecationWarning` category, as they are explicit and don't use the function from `utils`. I do not know if this is intended, but wanted to mention the apparent inconsistency. Maybe this is a another good issue for newcomers :)

PhilipVinc (approved):
Thanks, looking good now!

A thing that bothers me is that `jnp.linalg.lstsq` still uses rcond, but it seems it was not changed in numpy either, and this is not super consistent, but well... 

-- 
About the warning. You are right, it's inconsistent.
Someone should check which is the correct one and harmonise the warnings that we raise. 
I think there's a subtle difference between those two warnings, where FutureWarning is supposed to warn about a future deprecation while DeprecationWarning is about something already deprecated.
I think the deprecation one should be correct, but should check...

PR #1924, opened 2024-09-18
Make constraints more flexible for SpinOrbitalFermions

Z-Denis:
This lifts the requirement for a fermion-number constraint whenever a custom constraint is specified by the user. Previously, while a priori supported, this would trigger an error. This also makes that exception print a slightly clearer error message.


Z-Denis:
I use it locally on a project and it properly works for any combination of occupation and custom constraint.

PR #1990, opened 2025-01-31
Added warning when attempting to initialize a periodic lattice on a small lattice

wttai004:
This is a follow-up on the issue [here](https://github.com/netket/netket/issues/1989).

I added a warning class InitializePeriodicLatticeOnSmallLatticeWarning in netket.errors.py and used it in the Grid and Triangular Class in common_lattices.py. I think these two lattice classes are the only ones where the periodic lattice problem would happen; in more complicated lattices like diamond or Kagome the periodic looped edge is not equivalent to one of the edges in the open boundary case.

Please advise if the wording in the warning class is appropriate or if I should edit something.

### DIFF
--- a/netket/errors.py
+++ b/netket/errors.py
@@ -1128,3 +1128,38 @@ def concrete_or_error(force, value, error_class, *args, **kwargs):
         )
     except ConcretizationTypeError as err:
         raise error_class(*args, **kwargs) from err
+
+
+#################################################
+# Graph errors                                  #
+#################################################
+
+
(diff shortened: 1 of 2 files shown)


PhilipVinc (commented):
Can you add a coupled of test cases, checking that the error is raised (and that the error is not raised)?

Otherwise, the wording is good.

wttai004:
I've added some test cases in test/graph/test_graph.py to verify the implementation of InitializePeriodicLatticeOnSmallLatticeWarning.

PhilipVinc (approved):
Thanks!

PR #2025, opened 2025-04-15
History: complexification of histdict arrays

lgravina1997:
This PR is to merge the real and imaginary part stored as {'real':..., 'imag'...} when loading from Json data files


PhilipVinc (approved):
Thank you Luca!

PR #2041, opened 2025-05-14
Correcting wrong normalization in PT sampler (diffusion statistics)

lindamauron:
I just saw that the normalization of the diffusion of temperatures in the Parallel Tempering sampler is wrong (I obtained values > 1, while it should be normalized). Comparing to previous code (see [here](https://github.com/netket/netket/blob/87d469aa8c23f71c4838cf09d7ed7b87ff2ea01f/netket/legacy/sampler/numpy/metropolis_hastings_pt.py) for example), I found that a factor was supposed to be outside the square root, in opposition to what is done now. Nothing else was changed.


lindamauron:
I added the test for normalization in the same place where we verify that the acceptance is well defined

PhilipVinc (approved):
thanks!

PhilipVinc (approved):

PR #2051, opened 2025-05-26
Rename `SpaceGroupBuilder` to `SpaceGroup`, make it a subclass of `PermutationGroup`

attila-i-szabo:
Here's an implementation of the changes suggested in #2042

The main changes are the following:
* `SpaceGroupBuilder` is renamed `SpaceGroup`. It is a subclass of `PermutationGroup` so it can be directly returned by `Lattice.space_group`
* Introduced another subclass of `PermutationGroup` called `TranslationGroup` which deals with translations along any number of axes.
* The product table of both is implemented much more efficiently than a generic `PermutationGroup` (on my laptop, the product table of the space group of a 16x16 square lattice takes ~1 second with the new code and ~30 seconds with the old one)
* Since the "default space group" is doing a lot of work, I've memoised it in the Lattice object.

I've also deprecated `Lattice.space_group_builder()` and `SpaceGroup.space_group()` since both return SpaceGroup objects (`self` in the latter case).

Also I've noticed that `SpaceGroupBuilder` does not appear in the documentation. This is a bit of a problem, since a lot of the symmetry architecture is in this class (or `SpaceGroup` as it's now called). Could we change this?


attila-i-szabo:
@PhilipVinc @Adrien-Kahn  @gcarleo do you have any requests/suggestions to this PR so it can be merged?

gcarleo (approved):
this looks overall superficially good to me, and would tend to merge it soon to maybe build more features later

Adrien-Kahn:
I don't have any comments about this PR, but I do have questions/remarks about the group code in general, in particular about the encoding of permutations. Should I message you on slack or open an issue?

attila-i-szabo:
@gcarleo @PhilipVinc could we merge this?

netket/graph/space_group.py
```diff
@@ -148,91 +291,70 @@ def point_group_(self) -> PointGroup:
...
+    )
+    def space_group(self) -> "SpaceGroup":
         """
         The space group generated by `self.point_group` and `self.translation_group`.
         """
-        return self._full_translation_group @ self.point_group
+        return self
+
+    @struct.property_cached
+    def product_table(self) -> Array:
```
PhilipVinc, line 326: @attila-i-szabo is this code being tested down the line somewhere?

netket/graph/space_group.py
```diff
@@ -79,25 +79,145 @@ def _ensure_iterable(x):
...
+    but the product table is computed much more efficiently than a generic
+    `PermutationGroup`.
+    """
+
+    lattice: Lattice
+    """The lattice whose translation group is represented."""
+    axes: tuple[int]
+    """Axes translations along which are represented by the group."""
+    group_shape: tuple[int]
+    """Size of the translation group along each axis"""
```
PhilipVinc, line 135: We do not need to keep this, right? 
It can be computed from `lattice` and `axes`.
To minimise the risk of bugs, where this would get 'out of sync' or be set erroneously, I would just make it a property. 

I would not keep this as an attribute, and instead have it as a property (you can also cache it if you want and if you think it's expensive to compute).

PhilipVinc: something like
```python
@property # or @struct.property_cached
def group_shape(self):
        translation_by_axis = [_translations_along_axis(self.lattice, i) for i in self.axes]
        return tuple(len(x) for x in translation_by_axis)
```

netket/graph/space_group.py
```diff
@@ -79,25 +79,145 @@ def _ensure_iterable(x):
...
+            perms.append(Identity())
+        else:
+            # note that we need the preimages in the permutation
+            perm = lattice.id_from_position(p.preimage(lattice.positions))
+            perms.append(Permutation(perm, name=str(p)))
+    return PermutationGroup(perms, degree=lattice.n_nodes)
+
+
+@struct.dataclass
+class TranslationGroup(PermutationGroup):
```
PhilipVinc, line 121: Nice, thanks!

netket/graph/space_group.py
```diff
@@ -79,25 +79,145 @@ def _ensure_iterable(x):
...
+    """
+    Class to handle translation symmetries of a `Lattice`.
```
PhilipVinc, line 123:
```suggestion
    Class to handle translation symmetries of a `Lattice`. Corresponds to a representation of the translation group
    on the given lattice as a permutation group of `N_sites` variables.
```

netket/graph/space_group.py
attila-i-szabo: It actually follows from `lattice.pbc` and `lattice.extent` so it's quite cheap to compute!

netket/graph/space_group.py
```diff
@@ -79,25 +79,145 @@ def _ensure_iterable(x):
...
 
     Also generates space group irreps for symmetrising wave functions.
     """
 
     lattice: Lattice
+    """The lattice underlying the space group."""
     _point_group: PointGroup
+    """The geometric point group underlying the space group."""
+    point_group: PermutationGroup
+    """The point group as a `PermutationGroup` acting on the sites of `self.lattice`.
```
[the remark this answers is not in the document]
PhilipVinc: hmm.. no ok. But then can you add a comment like

```python
    point_group: PermutationGroup
    """The point group as a `PermutationGroup` acting on the sites of `self.lattice`. This is computed from '...' at construction and should not be changed directly."""
```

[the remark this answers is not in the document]
attila-i-szabo: I think it's a moot point
```
>>> import netket
>>> g = netket.graph.Square(4)
>>> sg = g.space_group()
>>> sg.point_group = '123'
  File "<string>", line 4, in __setattr__
dataclasses.FrozenInstanceError: cannot assign to field 'point_group'
```

netket/graph/space_group.py
```diff
@@ -79,25 +79,147 @@ def _ensure_iterable(x):
...
+
+    Group elements are listed in the order they appear in `self._point_group`."""
```
attila-i-szabo, line 217:
```suggestion
    Group elements are listed in the order they appear in `self._point_group`.
    Computed from `_point_group` upon construction, must not be changed after."""
```

PR #2052, opened 2025-05-27
Introduce geometry module for particle Hilbert spaces

gcarleo:
## Summary
- add geometry.Cell and FreeSpace as jax-compatible PyTrees
- require geometry object when creating `hilbert.Particle`
- update continuous-space tutorial to use geometry classes
- adjust unit tests and add geometry-specific tests

## Testing
- `pre-commit run --files netket/geometry/cell.py netket/hilbert/particle.py test/hilbert/test_hilbert.py test/geometry/test_cell.py docs/tutorials/gs-continuous-space.ipynb` *(fails: `fatal: unable to access 'https://github.com/astral-sh/ruff-pre-commit/': Couldn't connect to server`)*
- `pytest test/hilbert/test_hilbert.py::test_particle_with_geometry test/geometry/test_cell.py::test_distance_jit -q`

### DIFF
--- a/docs/api/api.md
+++ b/docs/api/api.md
@@ -21,6 +21,7 @@ drivers
 errors
 exact
 graph
+geometry
 hilbert
 jax
 logging
--- a/docs/api/geometry.md
+++ b/docs/api/geometry.md
@@ -1,0 +1,19 @@
+(netket_geometry_api)=
+# netket.geometry
+
+```{eval-rst}
+.. currentmodule:: netket.geometry
+```
+
+The :mod:`netket.geometry` module contains helper classes to describe the geometry of continuous spaces.
+These objects can be passed to {class}`~netket.hilbert.Particle` and provide utilities such as distance computations.
+
+```{eval-rst}
+.. autosummary::
+   :toctree: _generated/geometry
+   :template: class
+   :nosignatures:
+
+   Cell
+   FreeSpace
+```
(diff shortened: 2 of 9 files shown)


gcarleo:
I asked my new software engineer to implement my suggestion in #1575. Seems OK to me.

gcarleo:
It's ready to go. I didn't touch yet the abstract class for continous space Hilbert (it still takes L and pbc... super ugly) but I will do in a separate PR. 

core of this PR is separating the geometry from the constructor of the particle: 

```python
    geo = nk.geometry.Cell(d=2, L=(np.inf, 10.0), pbc=(False, True))
    hi1 = nk.hilbert.Particle(N=5, geometry=geo)
```

gcarleo:
yes in fact I conoletely forgot we had discussed the abstract class in a never merged PR ~2 years ago .... ehm. Indeed distances are necessary and different types of distances too (L1, L2 etc). For random I'm not sure yet how to do it. 

in any case, common geometries we need are essentially periodic cells with and without basis. in fact there is some common code with the corresponding graph class we already have. 

to proceed I would propose to merge this one, the I'll do another one moving stuff to experimental. Then another with the abstract class for geometry, even though that's not so urgent I feel. The most urgent is also having a more flexible particle class, that can support multiple types of particles (e.g. protons and electrons) etc... all of this now is very hacky.

gcarleo:
we will need to iterate fast on these classes, the question now is if we want to do it within netket or if we do it separately and likely merge later

PR #2053, opened 2025-05-28
Update lattice-fermions.ipynb

gcarleo:
The sampler in the fermion tutorial had not been updated when fermions were moved to main netket, I fixed this


gcarleo:
yes I used it yesterday for a school that's why I asked to make the changes

PR #2054, opened 2025-05-28
Update Heisenberg tutorial to nnx

gcarleo:
## Summary
- update the Heisenberg tutorial to use `flax.nnx`


PhilipVinc (approved):
I would remove the `jax.random.split` and just call `rngs.params()` twice. But anything is fine

PR #2064, opened 2025-06-20
Implement SpinfulParticle and ParticleSet

gcarleo:
## Summary
Following the discussion we had on the APIs, this implements the ideas, with some tweaks: 

- Introduced an experimental `ParticleSet` Hilbert space and `SpinfulParticle` dataclasses, enabling spin and position particles inside continuous cells
- `SpinfulParticle` is constructed reusing `netket.hilbert.Spin` and taking tensor products 
- Introduce convenience spinful particles, such as `Electron` and `Proton`, these are just partial instantiations of `SpinfulParticle`, more can be added if needed 
- `position_indices` is added to `ParticleSet` in order to extract what indices in the array of configurations correspond to positions, and what correspond instead to spins. 
- `Particle` and `SpinfulParticle` behave as single particle objects. Previously it was possible to declare `Particle(N=...)` at the moment this is being removed (see also discussion below) but we can decide whether it makes sense to put it back or not. 

This PR also does: 
- QuTiP conversion tests were updated for version 5’s collapsed dimension behavior, requiring only `q_obj.dims[1] == [1]`

The main thing I don't like is `position_indices` in `ParticleSet`, but it is impossible to do otherwise at the moment, given the NetKet limitation of samplers returning Arrays only. If we allowed samplers to return pytrees, we could much more easily handle these "hybrid space" cases. 
Another possible issue (that I don't find problematic, though) is that when the spin is fixed for a particle, then it is anyways stored in the Array of the sampler. Again, having a mechanism for samplers to work only on a subset of quantum numbers and ignore the others would remove this issue more broadly. 

Example usages: 

```python
# construct hilbert space for two electrons in two d, fully sampled spin  
cell = nkx.geometry.Cell(d=1, L=5.0, pbc=True)
hilb = nkx.hilbert.ParticleSet([nkx.hilbert.Electron(), nkx.hilbert.Electron()])

# hilbert space where first electron has fixed m_z projection, second one is being sampled 
hilb1 = ParticleSet([Electron(m_z=0.5), Electron()], cell)

# hilbert space with fixed proton (Born oppenheimer case) 
hibo = ParticleSet([Proton(position=(0.2,)), Electron()], cell)
```

More syntactic sugar will be added afterwards, most notably to construct molecules and maybe a shortcut to construct a set of identical particles. Now it is done this way: 

```python
cell = nkx.geometry.Cell(d=1, L=5.0, pbc=True)
N=128 
hilb = nkx.hilbert.ParticleSet(N*[nkx.hilbert.Electron()])
```

More broadly, one can ask if `ParticleSet` can be completely hidden, and one only works with `Particle` and `SpinfulParticle`.... This maybe is not too complicated to do and amounts to return a tensor product automatically when doing something like `hilb=Electron()*Proton()`, however I personally don't like this product syntax very much in this context....

### DIFF
--- a/test/hilbert/test_hilbert.py
+++ b/test/hilbert/test_hilbert.py
@@ -14,7 +14,7 @@
 
 import itertools
 from math import prod
-from functools import partial, reduce
+from functools import partial
 import netket as nk
 import numpy as np
 import pytest
@@ -683,7 +683,7 @@ def test_tensor_combination():
     assert len(hit._hilbert_spaces) == 5
     assert isinstance(repr(hit), str)
 
-    hi3 = reduce(lambda a, b: a * b, [part] * 5)
+    hi3 = nkx.hilbert.ParticleSet([part] * 5, geo_default)
     hit2 = hi1 * hi3
     assert isinstance(hit2, nk.hilbert.TensorHilbert)
     assert hit2.size == hi1.size + hi3.size
@@ -738,7 +738,7 @@ def test_tensor_combination():
     assert len(hit._hilbert_spaces) == 1
     assert isinstance(repr(hit), str)
 
-    hit = nk.hilbert.TensorHilbert(reduce(lambda a, b: a * b, [part] * 5))
+    hit = nk.hilbert.TensorHilbert(nkx.hilbert.ParticleSet([part] * 5, geo_default))
     assert isinstance(hit, nk.hilbert._tensor_hilbert.TensorGenericHilbert)
     assert len(hit._hilbert_spaces) == 1
     assert isinstance(repr(hit), str)
@@ -751,7 +751,7 @@ def test_errors():
     with pytest.raises(TypeError):
         hi * 1
 
-    hi = reduce(lambda a, b: a * b, [part] * 5)
+    hi = nkx.hilbert.ParticleSet([part] * 5, geo_default)
     with pytest.raises(TypeError):
         1 * hi
     with pytest.raises(TypeError):
--- a/test/variational/test_exact_state.py
+++ b/test/variational/test_exact_state.py
@@ -141,7 +141,7 @@ def test_qutip_conversion(vstate):
     assert q_obj.type == "ket"
     assert len(q_obj.dims) == 2
     assert q_obj.dims[0] == list(vstate.hilbert.shape)
-    assert q_obj.dims[1] == [1 for i in range(vstate.hilbert.size)]
+    assert q_obj.dims[1] == [1]
 
     assert q_obj.shape == (vstate.hilbert.n_states, 1)
     np.testing.assert_allclose(q_obj.data.to_array(), ket.reshape(q_obj.shape))
--- a/test/variational/test_variational.py
+++ b/test/variational/test_variational.py
@@ -339,7 +339,7 @@ def test_qutip_conversion(vstate):
     assert q_obj.type == "ket"
     assert len(q_obj.dims) == 2
     assert q_obj.dims[0] == list(vstate.hilbert.shape)
-    assert q_obj.dims[1] == [1 for i in range(vstate.hilbert.size)]
+    assert q_obj.dims[1] == [1]
 
     assert q_obj.shape == (vstate.hilbert.n_states, 1)
     np.testing.assert_allclose(q_obj.data.to_array(), ket.reshape(q_obj.shape))


test/hilbert/test_hilbert.py
```diff
@@ -675,7 +683,7 @@ def test_tensor_combination():
     assert len(hit._hilbert_spaces) == 5
     assert isinstance(repr(hit), str)
 
-    hi3 = nk.experimental.hilbert.Particle(N=5, geometry=geo_default)
```
PhilipVinc, line 678: I would feel more confident if those tests which are now of experimental features where moved to a different file.

As I don't trust 100% AI to not do random changes here and there, at least it's easy to check that they do not change the tests of other features to make sure they pass.

test/nn/test_blocks.py
```diff
@@ -105,13 +106,14 @@ def test_deepset():
...
+    cell = nkx.geometry.Cell(d=len(L), L=L, pbc=True)
+    hilb = nkx.hilbert.ParticleSet(
+        [nkx.hilbert.Electron() for _ in range(n_particles)],
+        cell,
     )
-    sdim = len(hilb.domain)
+    sdim = cell.dimension
     key = jax.random.PRNGKey(42)
-    x = hilb.random_state(key, size=1024)
+    x = hilb.random_state(key, size=1024)[:, : n_particles * sdim]
```
PhilipVinc, line 116: this is misterious.

PhilipVinc (commented):
I will not be reviewing this because I'm not sure I have something intelligent to say but as long as it does not make changes to 'established parts of netket' I'm ok with this.

I would feel better if this experimental stuff was in a separate package while experimenting, as we've done in the past, until a satisfying design is reached.
I personally feel that this stuff is super-experimental and people will need to use it to understand what makes sense to do and what does not in terms of a good code for dealing with particles in continuous systems.

I also personally don't like the code the AI is generating. 
It's not even using the same docstring notation as the rest of netket which might give issues in the documentation generation.

If it was confined to a separate package it would be easier for you to play with everything, you could let the AI do whatever it wants, and nobody has to check that it breaks something else.
You can just define a separate 'Hilbert space' and new 'operators'. The ones we have are terrible anyway.

But regardless, this seems not to break stuff so I'm ok with this.

jwnys:
Do we need "proton"? Can't we just define "nucleus" more generally? Maybe defining an intermediate "SpinHalfParticle" like in tensor network codes would be useful. This avoids the idea that it would also have a mass etc immediately associated to it, it's more abstract.

Would also be useful to have a simpler interface (again similar to e.g. itensor, or any tensor network code), that defined "hilb = nkx.hilbert.ParticleSet(N*[nkx.hilbert.Electron()])" in a different way. Just a constructor that takes the local hilbert space through nkx.hilbert.Electron() and makes a homogeneous hilbert space based on that. But that's personal taste.

I would also try to keep some degree of backward compatibility or so, or at least deprecation warnings and suggestions? There's +- 4 years of code being developed on the continuous part of netket, which now will be challenging to run. Although I understand the need to make an overhaul of the Continuous version, I still feel that this is a separate package, or needs to be at least first developed as one, and then maybe merged later. It took >3 years to put a working 2nd quantized fermions into the main code, and now things are changing rapidly without the time for public feedback.

We should also maybe convert "n_per_spin" to "n_particles_per_spin", similar to the fermionic code. Furthermore, developing the Hilbert space with so much flexibility without addressing the operators in parallel makes this hard to review. I am mainly wondering how we're going to nicely handle this there.

gcarleo:
I understand... we can make it back compatible with some converters, I am
just saying this is not what I would prioritize right now since nobody
wants to write those codes and I had to resort to Codex to do the job...
which is really not ideal but what can I do... In any case doing the
electron gas with the new interface is relatively simple, maybe I'll need
to add examples. Maybe the right thing to do at this point is to do an
external library to allow for faster developments

PhilipVinc:
> the main question is if there is a better way to isolate indices of
subparts of the Hilbert spaces without introducing pytrees for
configurations

I would actually be in favor of supporting pytrees for configurations.
It would require some changes everywhere, especially in the sampler, but should otherwise be doable and I think It makes sense.

Concatenating stuff makes us use the wrong dtype at times, and also the ordering is all confusing, so just having a well designed pytree to hold configurations would be good

PR #2067, opened 2025-06-24
Restructure Permutation

Adrien-Kahn:
The clarification of names in `Permutation` as discussed with @PhilipVinc and @attila-i-szabo. The objective is to solve the confusion that might be caused by the property `permutation` actually encoding the inverse permutation.

**Changes**:
- The constructor of `Permutation` now takes in either the permutation array or its inverse. `name` is now a keyword argument.
- The `Permutation` property `permutation` is deprecated in favor of `permutation_array` and `inverse_permutation_array`.
- `__repr__` now returns the array of the permutation instead of its inverse.
- `__array__` is deprecated and replaced by `Permutation.inverse_permutation_array` for more clarity.

---

As far as I can tell, the PR is functional right now, but I think it's better to wait for Attila's PR to be merged so we can update the usage of the `Permutation` class in `SpaceGroup` and the likes, and avoid an avalanche of deprecation warnings.

Two other points:
- There is the function `apply_to_id` in `Permutation`. From what I can see, it's only used twice inside the tests to obtain the permutation (instead of its inverse). Since it's effectively identical to `p.permutation_array[x]` now, would it make sense to deprecate it?
- There is this `HashableArray` business going on, and I'm not sure what it's for (being able to hash `permutation` I guess?). I think what I wrote in that regard makes sense, but you should probably check.

### DIFF
--- a/netket/utils/group/_permutation_group.py
+++ b/netket/utils/group/_permutation_group.py
@@ -30,13 +30,14 @@ from ._semigroup import Element
 
 
 class Permutation(Element):
-    def __init__(self,
-                 permutation: Array | None = None,
-                 *,# change one line somewhere
-                 name: str | None = None,
-                 permutation_array: Array | None = None,
-                 inverse_permutation_array: Array | None = None,
-                 ):
+    def __init__(
+        self,
+        permutation: Array | None = None,
+        *,  # change one line somewhere
+        name: str | None = None,
+        permutation_array: Array | None = None,
+        inverse_permutation_array: Array | None = None,
+    ):
         r"""
         Creates a `Permutation` from either the array of images `permutation_array` or preimages `inverse_permutation_array`. Note that the left action of a permutation on an array `a` is `a[inverse_permutation_array]`.
 
@@ -52,16 +53,22 @@ class Permutation(Element):
 
         arg_list = [permutation, permutation_array, inverse_permutation_array]
         if sum([arg is not None for arg in arg_list]) != 1:
-            raise TypeError("Only one among `permutation`, `permutation_array` and `inverse_permutation_array` must be specified.")
+            raise TypeError(
+                "Only one among `permutation`, `permutation_array` and `inverse_permutation_array` must be specified."
+            )
 
         if permutation is not None:
-            warn_deprecation("The argument `permutation` is deprecated. In order to clarify notations, you should either pass the array of images `permutation_array` or preimages `inverse_permutation_array`.")
+            warn_deprecation(
+                "The argument `permutation` is deprecated. In order to clarify notations, you should either pass the array of images `permutation_array` or preimages `inverse_permutation_array`."
+            )
             inverse_permutation_array = permutation
-        
+
         if permutation_array is not None:
             inverse_permutation_array = np.argsort(permutation_array)
-        
-        self._inverse_permutation_array = HashableArray(np.asarray(inverse_permutation_array))
+
+        self._inverse_permutation_array = HashableArray(
+            np.asarray(inverse_permutation_array)
+        )
 
         self.__name = name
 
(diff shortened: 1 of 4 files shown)


netket/utils/group/_permutation_group.py
```diff
@@ -23,35 +23,79 @@
...
+        self,
+        permutation: Array | None = None,
+        *,  # change one line somewhere
+        name: str | None = None,
+        permutation_array: Array | None = None,
+        inverse_permutation_array: Array | None = None,
+    ):
         r"""
-        Creates a `Permutation` from an array of preimages of :code:`range(N)`.
+        Creates a `Permutation` from either the array of images `permutation_array` or preimages `inverse_permutation_array`. Note that the left action of a permutation on an array `a` is `a[inverse_permutation_array]`.
```
attila-i-szabo, line 42: Break up to lines of up to 70-80 characters as per PEP 8. Also, the first tagline should be no more than 1-2 of these lines, because it makes the autogenerated documentation weird. The "Note that" bit can come after an empty line - and probably expanded.

netket/utils/group/_permutation_group.py
```diff
@@ -23,35 +23,79 @@
...
 
-        Arguments:
-            permutation: 1D array listing :math:`g^{-1}(x)` for all :math:`0\le x < N`
-                (i.e., :code:`V[permutation]` permutes the elements of `V` as desired)
-            name: optional, custom name for the permutation
+        Args:
+            permutation: (deprecated) 1D array listing :math:`g^{-1}(x)` for all :math:`0\le x \le N-1`.
+            name: Optional, custom name for the permutation.
+            permutation_array: 1D array listing :math:`g(x)` for all :math:`0\le x \le N-1`.
+            inverse_permutation_array: 1D array listing :math:`g^{-1}(x)` for all :math:`0\le x \le N-1`.
```
attila-i-szabo, line 48: Shoudl mention that you only take one of the three arguments

netket/utils/group/_permutation_group.py
```diff
@@ -23,35 +23,79 @@
...
         Returns:
-            a `Permutation` object encoding the same permutation
+            A `Permutation` object encoding the same permutation.
         """
-        self.permutation = HashableArray(np.asarray(permutation))
+
+        arg_list = [permutation, permutation_array, inverse_permutation_array]
+        if sum([arg is not None for arg in arg_list]) != 1:
+            raise TypeError(
+                "Only one among `permutation`, `permutation_array` and `inverse_permutation_array` must be specified."
```
attila-i-szabo, line 57:
```suggestion
                "Precisely one of `permutation`, `permutation_array`,\n"
                "and `inverse_permutation_array` must be specified."
```
not specifying any of them is also bad

netket/utils/group/_permutation_group.py
```diff
@@ -23,35 +23,79 @@
...
+            )
+
+        if permutation is not None:
+            warn_deprecation(
+                "The argument `permutation` is deprecated. In order to clarify notations, you should either pass the array of images `permutation_array` or preimages `inverse_permutation_array`."
```
attila-i-szabo, line 62:
```suggestion
                "The argument `permutation` is deprecated.\n\n"
                "For clarity, you should either pass the array of images\n"
                "`permutation_array` or preimages `inverse_permutation_array`.\n"
                "The latter matches the the behaviour of `permutation`."
```

attila-i-szabo (commented):
I'm happy with this, the suggestions are basically all on formatting

Adrien-Kahn:
I also modified the class `Translation` that inherits from `Permutation` so that it matches the new version.

@attila-i-szabo, while going through the code looking for places where I should switch to the new arguments of `Permutation` and `Translation`, I made a change inside `_translations_along_axis` and `_pg_to_permutation`, and I wanted to make a sanity check that the tests would fail if I passed the argument as `permutation_array` instead of `inverse_permutation_array` (which is the correct one), but the tests still passed. From what I could see, these functions are used to construct the full space group, so maybe that's why it doesn't matter if everything is inverted, but I still wanted to mention it.

PhilipVinc (approved):
This looks good now, thank you.

As for the missing tests, yeah, that's life. This is not clearly tested but it's ok because it means that not much is directly dependent on it.

PR #2068, opened 2025-06-26
Fix qutip conversion tests

Adrien-Kahn:
`QuTiP` now automatically converts trivial `Qobj` dims like `[1, 1, 1, 1]` to `[1]`. This PR fixes the tests to account for this change of behaviour.


gcarleo:
was being fixed in #2064 actually...

PR #2075, opened 2025-07-04
Allow specifying `characters` directly in `SymmSumExp`

attila-i-szabo:
Given that space group irreps can be generated much more conveniently than by looking at the (gigantic) full character table, we should allow specifying characters as arrays. This PR does that for `SymmSumExp`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
 
 ## New features
 
+* `SymmSumExp` supports specifying group characters as an array in addition to their index in the full character table. [#2075](https://github.com/netket/netket/pull/2075).
+
 ## Deprecations and Removals
 
 
--- a/netket/nn/blocks/symmetry_sum.py
+++ b/netket/nn/blocks/symmetry_sum.py
@@ -96,7 +96,7 @@ class SymmExpSum(nn.Module):
 
     chi: HashableArray | None = None
     """Character of the space group to project onto.
-    
+
     Only one of `chi` and `character_id` may be specified.
     If neither is specified, the character is taken to be all 1,
     yielding a trivially symmetric state.
--- a/netket/nn/blocks/symmetry_sum.py
+++ b/netket/nn/blocks/symmetry_sum.py
@@ -146,7 +146,7 @@ class SymmExpSum(nn.Module):
         # Compute the log-wavefunction obtaining (-1,) and reshape to (N_symm, ...)
         psi_symm = self.module(x_symm).reshape(*x_symm_shape[:-1])
 
-        characters = self._chi.reshape((-1,) + tuple(1 for _ in range(x.ndim - 1)))
+        characters = np.expand_dims(self._chi, range(1, x.ndim))
 
         # If those are all positive, then use standard logsumexp that returns a
         # real-valued, positive logsumexp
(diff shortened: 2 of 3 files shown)


PhilipVinc:
Can I get a comment by @Adrien-Kahn and @rajnutakki ? 
What do you think?

rajnutakki:
This makes a lot of sense to me. I always think about symmetrization in terms of what space group irrep I am using and avoiding the character table removes an unnecessary intermediate step

Adrien-Kahn:
I don't really understand the kind of context in which one would have $\chi$ but not the full character table. Except maybe in the case where one would repeatedly project to the same character and know it by heart, but even then does the generation of the character table take long enough to justify taking the risk of passing an invalid character?

Adrien-Kahn:
Either way, I would argue that the argument `character_id` is probably the one most users would use, and is also the safest, so maybe it should be listed before `\chi`.

attila-i-szabo:
> Either way, I would argue that the argument `character_id` is probably the one most users would use, and is also the safest, so maybe it should be listed before `chi`.

The order the parameters are listed has zero effect on behaviour.

PhilipVinc:
@attila-i-szabo can you add an example of passing the irrep directly, like the one you mentioned above in the discussion?

Also, wouldn-t it be better to call this new argument `irrep` instead of `chi` ? 
(This makes me think that we could rename `character id` to `irrep_id` as well, but ok..)

Adrien-Kahn:
Ah, ok, I didn't know of this way of indexing space group irreducible representations. Then I guess it makes sense.

PhilipVinc:
ok, thank you. 
 I'd prefer something more explicit than just 'chi'

PhilipVinc:
@attila-i-szabo in the main docstring of the SymmSumExp class. There's already an example
```python
    Examples:
       Constructs a :ref:`netket.nn.blocks.SymmExpSum` for a bare
       :ref:`netket.models.RBM`, summing over all translations of a
       2D Square lattice
       >>> import netket as nk
       >>> graph = nk.graph.Square(3)
       >>> print("number of translational symmetries: ", len(graph.translation_group()))
       number of translational symmetries:  9
       >>> # Construct the bare unsymmetrized machine
       >>> machine_no_symm = nk.models.RBM(alpha=2)
       >>> # Symmetrize the RBM over all translations
       >>> ma = nk.nn.blocks.SymmExpSum(module = machine_no_symm, symm_group=graph.translation_group())
       If you have a Convolutional NN that is already invariant under translations, you might
       want to only symmetrize over the point-group (mirror symmetry and rotations).
       >>> import netket as nk
       >>> graph = nk.graph.Square(3)
       >>> print("number of point-group symmetries: ", len(graph.point_group()))
       number of point-group symmetries:  8
       >>> # Construct the bare unsymmetrized machine
       >>> machine_no_symm = nk.models.RBM(alpha=2)
       >>> # Symmetrize the RBM over all translations
       >>> ma = nk.nn.blocks.SymmExpSum(module = machine_no_symm, symm_group=graph.point_group())

```

Just add another one (before this one if possible).

As for the more general tutorial. Yes, I'd agree. Adrian had written something that was quite nice, which could be completed and complemented to showcase this stuff. Maybe a tutorial to show on netket...

PhilipVinc:
thanks!

PhilipVinc (approved):

PR #2102, opened 2025-07-17
Change condition for scientific notation when printing Stats object

Adrien-Kahn:
When printing a `Stats` object, the current condition for it to be displayed in scientific notation is `std<1e-7`. This PR proposes to change the condition to `value<1e-3`. The reason is that when optimising a quantity that should converge to 0, such as the infidelity or a centred energy, it's much more convenient to read and track progress of a number like `1.234e-5` rather than `0.00001234`, especially on a progress bar.

### DIFF
--- a/netket/stats/mc_stats.py
+++ b/netket/stats/mc_stats.py
@@ -28,7 +28,7 @@ from ._autocorr import integrated_time
 
 
 def _format_decimal(value, std, var):
-    if math.isfinite(std) and std > 1e-7:
+    if math.isfinite(abs(value)) and value > 1e-3:
         decimals = max(int(np.ceil(-np.log10(std))), 0)
         return (
             "{0:.{1}f}".format(value, decimals + 1),
--- a/netket/stats/mc_stats.py
+++ b/netket/stats/mc_stats.py
@@ -36,10 +36,11 @@ def _format_decimal(value, std, var):
             "{0:.{1}f}".format(var, decimals + 1),
         )
     else:
+        n_digits = max(int(np.ceil(-np.log10(std / abs(value)))), 0) + 1
         return (
-            f"{value:.3e}",
-            f"{std:.3e}",
-            f"{var:.3e}",
+            f"{value:.{n_digits}e}",
+            f"{std:.{n_digits}e}",
+            f"{var:.{n_digits}e}",
         )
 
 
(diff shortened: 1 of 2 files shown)


Adrien-Kahn:
@PhilipVinc I made the number of digits depend on the standard deviation in scientific notation like in the decimal case, but I'm not a huge fan of it because now the progress bar squiggles even more than before due to the varying number of digits.

PhilipVinc:
(Can you show a video of the problem?)Il giorno 17 lug 2025, alle ore 20:19, Adrien-Kahn ***@***.***> ha scritto:﻿Adrien-Kahn left a comment (netket/netket#2102)
@PhilipVinc I made the number of digits depend on the standard deviation in scientific notation like in the decimal case, but I'm not a huge fan of it because now the progress bar squiggles even more than before due to the varying number of digits.

Adrien-Kahn:
> (Can you show a video of the problem?)Il giorno 17 lug 2025, alle ore 20:19, Adrien-Kahn ***@***.***> ha scritto:﻿Adrien-Kahn left a comment ([netket/netket#2102](https://github.com/netket/netket/pull/2102)) @PhilipVinc I made the number of digits depend on the standard deviation in scientific notation like in the decimal case, but I'm not a huge fan of it because now the progress bar squiggles even more than before due to the varying number of digits. —Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you were mentioned.Message ID: ***@***.***>


https://github.com/user-attachments/assets/59951288-f5a6-4709-a1c5-4d6ab5922050

At the beginning, it's pretty much unreadable.

Adrien-Kahn:
One thing we can do to fix the squiggling is to reserve a fixed number of character per variable field (like `value` or `std`), and pad them with nothing if the space is not fully occupied. This way we have fixed-length strings and we minimise the squiggling.

PhilipVinc:
Yes, I think it's a good idea.
I think for example R hat could be always printed with 3 digits of decimal precision (even 2 maybe...). Sometimes we print it with 5 which is completely useless.

The variance as well, could be printed with limited number of digits, and we could swap to scientific notation when it's below 1e-2 or so.

This should already reduce the squiggling. 

(By the way, this is the kind of code that chatgpt is good at getting right..)

Adrien-Kahn:
No ChatGPT in this PR.

PhilipVinc (approved):
Thank you. Seems good to me now.

PR #2106, opened 2025-07-21
Update state within sweeps of ParallelTempering sampler

lindamauron:
Modifies the state of the sampler during the sweep, as done for the standard `MetropolisSampler`.


PhilipVinc (approved):
thanks!

PR #2120, opened 2025-07-29
Update automorphism

Adrien-Kahn:
Very minor change to `graph._compute_automorphisms` to remove the deprecation warning due to the updated `Permutation` class.

I retained the previous behaviour so nothing changes, but technically it was sending each permutation as an inverse permutation. It doesn't change anything when constructing the full group though.

(I messed with the other branch update_automorphism with an underscore, it can be deleted)


PhilipVinc (approved):
thanks

PR #2122, opened 2025-08-14
Symmetry Overhaul

Adrien-Kahn:
This PR (in collaboration with @PhilipVinc and @louishrm) proposes to overhaul the symmetrisation process in NetKet by making it closer to the mathematical formalism and allowing more general usage than currently possible.

Currently, NetKet only supports symmetries that are lattice symmetries on spin systems. At the mathematical level, there is a confusion between group and representation, where the representation is implicitly assumed to be the permutation representation on basis element labels. Therefore, it does not support non-lattice symmetry operators such as spin-flip (the operator $\prod_{k} \sigma_k^{(x)}$), or lattice symmetry operators of fermionic systems, which differ by a fermionic sign from their spin counterpart.

This PR introduces the class `Representation` to fill this gap, allowing arbitrary representation of a given group. From the `Representation` object, we can obtain the projection operator on the subspace associated to any irreducible representation, and use it to construct the projected variational state. Below is an example showcasing the comparison between the old and new workflow.

```python
import matplotlib.pyplot as plt
import numpy as np

import netket as nk


hilbert_space = nk.hilbert.Qubit(4)
graph = nk.graph.Square(2, pbc=False)

sampler = nk.sampler.MetropolisLocal(hilbert_space)

hamiltonian = nk.operator.IsingJax(hilbert_space, graph, 1, 1)
exact_gs = nk.exact.lanczos_ed(hamiltonian).item()
hamiltonian -= exact_gs

model = nk.models.RBM(alpha=4)

# Old workflow
group = graph.space_group()
symmetrized_model = nk.nn.blocks.SymmExpSum(model, group, character_id=0)
sym_vstate_old = nk.vqs.MCState(sampler, symmetrized_model)

log_old = nk.logging.RuntimeLog()
driver = nk.VMC(hamiltonian, nk.optimizer.Sgd(5e-3), variational_state=sym_vstate_old)
driver.preconditioner = nk.optimizer.SR(solver=nk.optimizer.solver.cholesky, diag_shift=1e-6)
driver.run(1000, out=log_old)

# New workflow
unsym_vstate = nk.vqs.MCState(sampler, model)
representation = graph.space_group_representation(hilbert_space)
sym_vstate_new = representation.project(unsym_vstate, 0)

log_new = nk.logging.RuntimeLog()
driver = nk.VMC(hamiltonian, nk.optimizer.Sgd(5e-3), variational_state=sym_vstate_new)
driver.preconditioner = nk.optimizer.SR(solver=nk.optimizer.solver.cholesky, diag_shift=1e-6)
driver.run(1000, out=log_new)

# Plot
plt.plot(np.abs(log_old['Energy']['Mean']), label='old')
plt.plot(np.abs(log_new['Energy']['Mean']), label='new')

plt.yscale('log')
plt.legend()
```

In more details, this PR introduces:
- `netket.symmetry.Representation`: It allows the construction of arbitrary representations for a given group by providing the operator associated to each group element. The methods `get_projector` and `project` allow the construction of the projector associated to a given irreducible representation and the projection of a variational state respectively. The method `get_irrep_subspace_dims` computes the dimension of the subspaces associated to each irreducible representation.
- `netket.operator.permutation.PermutationOperator` and `netket.operator.permutation.PermutationOperatorFermion`: The permutation operators for spin and fermionic systems respectively. They are standard `DiscreteJaxOperator` that are used to construct representations of lattice symmetries.
- `netket.vqs.apply_operator`: A function that applies an operator on a variational state and returns the transformed variational state by wrapping its `apply_fun`. It is used to construct the variational state projected on a symmetry sector.

I made the `group` module available as `netket.symmetry.group` since I think it makes sense to have all the symmetry-related tools in one place. It is still available as `netket.utils.group` as well. I also left all the new tests in `test/symmetry` so they are all in one place, but eventually they should be moved where they belong.

The PR is not 100% ready for merge yet, but I wanted to put it out now so we can discuss about the changes. Among other features, we want to have a way to combine different representations (for example lattice symmetry and spin-flip), that is designed with progressive symmetrisation in mind. We also need to write more documentation, about both the theory and practical usage.

@attila-i-szabo

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -120,9 +120,6 @@ jobs:
         with:
           python-version: "3.13"
 
-      - name: Install coverage for merging
-        run: uv pip install --system coverage
-
       - uses: actions/download-artifact@v4
         with:
           pattern: coverage-*


netket/_src/symmetry/representation.py
```diff
@@ -0,0 +1,180 @@
...
+    A representation of a group. For a given group, it is effectively a dictionary
+    that maps the elements of that group to their corresponding operators.
+    """
+
+    def __init__(
+        self,
+        group: FiniteGroup,
+        representation_dict: dict[Element, DiscreteJaxOperator],
+    ):
+
```
PhilipVinc, line 21: docstring. Especially discussing what is representation dict with an exaple

netket/_src/symmetry/representation.py
```diff
@@ -0,0 +1,180 @@
...
+        operator = next(iter(representation_dict.values()))
```
PhilipVinc, line 22: Can you add a check at the beginning?
```
if not isinstance(group, FiniteGroup):
    raise TypeError("group must be a FiniteGrup")
```

netket/_src/symmetry/representation.py
```diff
@@ -0,0 +1,180 @@
...
+        if isinstance(key, Element):
+            return self.representation_dict[key]
+        elif isinstance(key, int):
+            return self.representation_dict[self.group[key]]
+        raise TypeError("Index should be integer or group element")
+
+    def __iter__(self):
+        return iter(self.representation_dict.items())
+
+    def get_projector(self, character_index: int) -> DiscreteJaxOperator:
```
PhilipVinc, line 45: I am against `get_` functions (it's something people used to do in C or when wrapping C libraries because classes did not exist. No reason to use it in python).

I think this can be just `projector(character_index)

netket/_src/symmetry/representation.py
```diff
@@ -0,0 +1,180 @@
...
+        try:
+            character = [op.trace() for perm, op in self]
+        except NotImplementedError:
+            raise NotImplementedError(
+                "At least one operator of the representation"
+                " does not implement the trace"
+            )
+        return np.array(character)
+
+    def get_irrep_subspace_dims(self):
```
PhilipVinc, line 79:
```suggestion
    def get_irrep_subspace_dims(self) -> int:
```

netket/_src/symmetry/representation.py
```diff
@@ -0,0 +1,180 @@
...
+        """
+        Return the dimension of the subspace associated to each irreducible
+        representation. Requires that each operator of the representation
+        implements the trace method.
```
PhilipVinc, line 83:
```suggestion
        Return the dimension of the subspace associated to each irreducible
        representation. 
        
        Requires that each operator of the representation implements the trace method.
```

netket/_src/operator/permutation/permutation_operator_fermion.py
```diff
@@ -0,0 +1,138 @@
...
+
+    def __eq__(self, other):
+        if isinstance(other, PermutationOperatorFermion):
+            return (
+                self.hilbert == other.hilbert and self.permutation == other.permutation
+            )
+        else:
+            return False
+
+    def get_signs(self, x):
```
PhilipVinc, line 100:
```suggestion
    def _get_signs(self, x):
```

netket/_src/symmetry/representation.py
```diff
@@ -0,0 +1,180 @@
+import numpy as np
+import jax.numpy as jnp
+
+from functools import reduce
+
+from netket.utils.group import Element, FiniteGroup
+from netket.operator import DiscreteJaxOperator
+
+
+class Representation:
```
PhilipVinc, line 10: can you add an `__repr__` to this class

netket/_src/symmetry/representation.py
```diff
@@ -0,0 +1,180 @@
...
+            irrep_dims.append(projector_rank)
+
+            cob_matrix[:, current_index : current_index + projector_rank] = (
+                selected_eigvecs
+            )
+            current_index += projector_rank
+
+        return cob_matrix, np.array(irrep_dims)
+
+    # def construct_commuting_product(self, other, check_commute: bool = False):
```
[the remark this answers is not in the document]
Adrien-Kahn: I'll put a comment next to it to explain that.

netket/_src/symmetry/representation.py
```diff
@@ -39,11 +66,36 @@ def __getitem__(self, key):
...
+        if type(self) is type(other):
+            return (
+                self.hilbert_space == other.hilbert_space
+                and self.group == other.group
+                and self.representations == other.representations
+            )
+        return False
+
+    @cached_property
+    def representation_dict(self) -> dict[Element, DiscreteJaxOperator]:
```
Adrien-Kahn, line 84: @PhilipVinc, why did you remove `representation_dict` as a field and put `representations` instead? It should be called `operators` rather than `representations`, but is there a problem with keeping `representation_dict` in the first place?

PhilipVinc:
As far as I am concerned the checklist of things that are still missing are:

- [ ] Reduce the number of test cases for Permutation from 1400 to something reasonable (~10)
- [ ] Add a test case for automatically constructing the representation of a group on a fermionic Hilbert space
- [ ] Add a test case for this projector
- [ ] Add 1/2 example scripts to the Example/ directory
- [ ] Cleanup all leftovers from the notebook/documentation that you have not finished writing yet (you can finish them in a future PR)

PhilipVinc:
I also added a flag to disable the newly added check for the validity of permutation indicess (@Adrien-Kahn ) and disabled in internal methods. The idea is that we validate only user input, not internally created objects because we assume the internal logic is correct.

@louishrm that was not the good changeling file! I moved it to another.

Is there anything else missing?

PhilipVinc:
@attila-i-szabo if you have a moment, I would like your comment on how this turned out.
Can you give a look to https://netket--2122.org.readthedocs.build/en/2122/tutorials/symmetry_tutorial.html and give me an opinion?

PhilipVinc (approved):
Lovely, thanks a lot guys, this is great!
I'll merge it. We might have to iterate a bit more on it in the near future and maybe tweak a few things, but it's already a huge improvement.

PR #2151, opened 2025-09-29
Fix get_local_estimator not hitting the flattening of leading axes

Z-Denis:
Minimal fix. The function `get_kernels` takes samples flattened over the leading dimensions. While in `infidelity_sampling_inner` this reshaping was carried out before calling `get_kernels`, `get_local_estimator` would not do it, leading to an obscure ValueError message. This moves the reshaping directly into `get_kernels`.


PhilipVinc (approved):
Could you add a single, short test case that was not working without this fix?
At the end of this file https://github.com/netket/netket/blob/master/test/driver/test_infidelity_vmc.py

It should be a function 
```
def test_batch_supported():
    vs = ... (something small)
    H = ....

PR #2158, opened 2025-10-03
Added GlobalFlipRule

alessandro-santini:
Added GlobalFlipRule

### DIFF
--- a/docs/api/sampler.md
+++ b/docs/api/sampler.md
@@ -118,6 +118,7 @@ Sampler. Rules with `Numpy` in their name can only be used with
 
   rules.MetropolisRule
   rules.LocalRule
+  rules.GlobalSpinFlipRule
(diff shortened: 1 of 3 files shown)


PhilipVinc:
Please:
 - call it `GlobalSpinFlipRule`
 - add the following code to error inside of `transition` (you will have to import `from netket.hilbert import Spin`) to avoid people getting it wrong...
```python
        if not isinstance(sampler.hilbert, Spin):
            raise TypeError(
                "GlobalSpinFlipRule is only compatible with Spin hilbert spaces"
            )
```
 - add it to this list so it shows up in the documentation https://github.com/netket/netket/blob/3d5bf68da9fdd3b58fa424095350c4a8af415411/docs/api/sampler.md?plain=1#L123

PR #2171, opened 2025-11-06
Extend infidelity_sr to FullSumState

Z-Denis:
This adds the possibility to input importance-sampling weights in SR. `FullSumState` is dealt with as a particular case, extending the new `Infidelity_SR` driver to such states.

Changes:
- `Infidelity_SR` now accepts `FullSumState` as a variational and target state.
- `_sr_srt_common(..., use_ntk=False)` now accepts importance-sampling weights.
- `_sr_srt_common(..., use_ntk=False)` now accepts full summation upon providing `samples=hilbert.all_states()` and `weights=state.probability_distribution()`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@
 
 ## NetKet 3.21 (In development)
 
+### New Features
+* {class}`netket.driver.VMC_SR` and {class}`netket.experimental.driver.Infidelity_SR` now also support {class}`netket.vqs.FullSumState` [PR #2171](https://github.com/netket/netket/pull/2171).
+
+
 ## NetKet 3.20 (15 Octobre 2025)
 
 ### Breaking Changes


PhilipVinc:
@lgravina1997 can you check that this is ok ? (it seems ok to me) 
Mainly I'd like to know if we can then use the same logic by passing weight=something to do IS with this.

Z-Denis:
It is meant to, I did some simple tests locally putting a flat weight distribution and using a `MCState` with a ton of samples but I can test more thoroughly.

In this PR I assumed the IS weights not to be properly self-normalised so I rescale them. We might want to check in the future, should NetKet provide full support for IS, that this is not done twice: in `_sr_srt_common` and in the driver.

The idea behind this PR is not to treat full summation and IS separately.

lgravina1997:
I think this should work. However, in the implementations we have, the weights are all computed in _prepare_derivative, within the driver both in the case of MC and FS. For me this is the most intuitive way to do it. If you do this in the end you dont even need the logic separating pdf and weights in _sr_srt_common.

Z-Denis:
I removed some of the hackery I had to rely upon within `expect.py` and avoided recomputing weights. There are still a bunch of ugly if statements checking whether weights are None that one could avoid at the cost of carrying a uniform weight vector for the bare Monte Carlo case. This is, I believe, what @lgravina1997 does in his code.

@PhilipVinc, I really dislike defining `get_samples_and_pdf` within `_sr_srt_common`. This part should be pretty much agnostic of the NetKet high-level stuff like vqs. Could we at least move it to `vmc_sr.py`?

netket/_src/ngd/sr_srt_common.py
```diff
@@ -33,17 +62,18 @@ def _prepare_input(
...
-    N_mc = O_L.shape[0]
-
     local_grad = local_grad.flatten()
-    de = local_grad - jnp.mean(local_grad)
+    de = local_grad - jnp.sum(scaling_factor * local_grad)
 
-    O_L = O_L / jnp.sqrt(N_mc)
-    dv = 2.0 * de / jnp.sqrt(N_mc)
+    dv = 2.0 * de * jnp.sqrt(scaling_factor)
+    if jax.numpy.ndim(scaling_factor) != 0:
```
PhilipVinc:
```suggestion
    if jnp.ndim(scaling_factor) != 0:
```

Z-Denis:
I agree it simplifies the code and avoids a conditional statement. However, it involves an extra allocation of `n_samples` many floats when using a Monte Carlo state, which is the reason I did not simplify the logic that way, but I can if that's something we are ok with. @PhilipVinc, any thoughts on this?

Z-Denis:
@PhilipVinc, tests pass now.

Issue #2183, opened 2025-11-23
ParticleNumberAndSpinConserving operator conversion bug


PhilipVinc:
cc @inailuig 

The conversion from FermionOperator2nd to PNCAS conserving fails...
It took me a long time to track this down. Do you have an idea of what might cause it?
It only shows up when you have an hopping term between subsectors.
Or maybe this kind of term is not supported (but then we should error properly)

```python
from netket.hilbert import SpinOrbitalFermions
from netket.operator import FermionOperator2nd
from netket.experimental.operator import ParticleNumberAndSpinConservingFermioperator2nd
import numpy as np

N = 3  
n = 2 
hi = SpinOrbitalFermions(N, s=0.5, n_fermions_per_spin=(n, n))

# Off-diagonal mixed-sector term
ha = FermionOperator2nd(hi, terms=["0^ 3^ 3 1"], weights=[-1.0])
ha_pnc = ParticleNumberAndSpinConservingFermioperator2nd.from_fermionoperator2nd(ha)

np.testing.assert_allclose(ha.to_dense(), ha_pnc.to_dense())
```

PR #2184, opened 2025-11-24
Fix a bug in ParticleNumberAndSpinConservingFermioperator2nd where the ordering of spin sectors was reversed

inailuig:
Fix https://github.com/netket/netket/issues/2183


PhilipVinc (approved):
Thank you

PR #2194, opened 2026-01-26
(Minor redundancy) operators list initialized twice in Representation.projector

chenjl517:
Hi,

I noticed a minor redundancy in `netket/_src/symmetry/representation.py`. The `operators` list is initialized twice (see line 140 and line 144):

https://github.com/netket/netket/blob/a8675bdc6f457b922262b0acfc5916183fc6c2c1/netket/_src/symmetry/representation.py#L137-L146

line 144 seems redundant.


PhilipVinc:
Sorry for the delay, and thank you!

PR #2200, opened 2026-02-22
NTK distributed solver using jax_mg 

PhilipVinc:
This allows to go up to ~100k samples . Credits to Roeland Wiersma @therooler


therooler:
Hey, I check the HPC configuration, which requires checking how many nodes are active with the allgather. I technically only need to know if we're in SPMD or MPMD mode and could remove the allgather in the newest version I'm working on.

Related to our discussions at CCQ @inailuig: I had a chat today with the NVIDIA Math libraries team and they will connect me with some people there working on JAX and XLA which will hopefully smooth the way for developing JAXMp at some point in the future and figure out how to expose the XLA communicator.

PR #2211, opened 2026-03-13
Vit architecture

riccardo-rende:
Fix vit tutorial and add transformer architecture in models

@llviteritti


netket/models/transformer.py
```diff
@@ -0,0 +1,334 @@
...
+
+        return x
+
+@partial(jax.vmap, in_axes=(None, 0, None), out_axes=1)
+@partial(jax.vmap, in_axes=(None, None, 0), out_axes=1)
+def roll2d(spins, i, j):
+    side = int(spins.shape[-1] ** 0.5)
+    spins = spins.reshape(spins.shape[0], side, side)
+    spins = jnp.roll(jnp.roll(spins, i, axis=-2), j, axis=-1)
+    return spins.reshape(spins.shape[0], -1)
```
attila-i-szabo, line 114: Should we use the existing convolutional network layers instead? (See #2215 for a proposal to make them more user-friendly.)

netket/models/transformer.py
```diff
@@ -0,0 +1,334 @@
...
+
+        # concatenate the different heads
+        x = rearrange(
+            x, "batch n_patches n_heads d_eff ->  batch n_patches (n_heads d_eff)"
+        )
+
+        # the representations of the different heads are combined together
+        x = self.W(x)
+
+        return x
```
attila-i-szabo, line 106: Couldn't this be incorporated into FMHA with an optional gamma parameter?

Examples/Transformer/heisenberg_square_J1J2.py
```diff
@@ -0,0 +1,83 @@
+import os
+os.environ['NETKET_EXPERIMENTAL_SHARDING'] = '1'
+
```
PhilipVinc, line 3:
```suggestion
```

this stuff is not needed anymore since a while.

Examples/Transformer/heisenberg_square_J1J2.py
```diff
@@ -0,0 +1,83 @@
...
+import jax
+jax.distributed.initialize()
+
```
PhilipVinc, line 6:
```suggestion
```

I mean I'm not against this, but then put it behind a flag like... if there's the SLURM env variable or something. otherwise the script won't run.

Examples/Transformer/heisenberg_square_J1J2.py
```diff
@@ -0,0 +1,83 @@
...
+import matplotlib.pyplot as plt
+
+import netket as nk
+
+import jax.numpy as jnp
+from netket.models.transformer import ViT
```
PhilipVinc, line 12: See discussion in the comment below. but I'd rather you put the full model code in here.
```suggestion
# insert full model here
```

Examples/Transformer/heisenberg_square_J1J2.py
```diff
@@ -0,0 +1,83 @@
...
+            variational_state=vstate,
+            mode='complex',
+            chunk_size_bwd=N_samples,
+            on_the_fly=True,
+            use_ntk=True)
+
+log = nk.logging.RuntimeLog()
+
+N_opt = 800
+vmc.run(n_iter=N_opt, out=log)
```
PhilipVinc, line 83: format

Examples/Transformer/heisenberg_square_J1J2.py
```diff
@@ -0,0 +1,83 @@
...
+N_params = nk.jax.tree_size(vstate.parameters)
+print("Number of parameters = ", N_params, flush=True)
+
+# Variational monte carlo driver
+vmc = nk.driver.VMC_SR(hamiltonian=hamiltonian,
+            optimizer=optimizer,
+            diag_shift=1e-4,
+            variational_state=vstate,
+            mode='complex',
+            chunk_size_bwd=N_samples,
```
PhilipVinc, line 76: Is there a reason for that ? this is basically the default.

Examples/Transformer/heisenberg_square_J1J2.py
```diff
@@ -0,0 +1,83 @@
+import os
+os.environ['NETKET_EXPERIMENTAL_SHARDING'] = '1'
+
+import jax
+jax.distributed.initialize()
+
+import matplotlib.pyplot as plt
```
PhilipVinc, line 7:
```suggestion
```

if you don't use it don't import it

PhilipVinc (changes requested):
Hi @riccardo-rende , you should have pinged me. I forgot about this.

Remarks on the notebook:
 - the image in the tutorials should be fixed. it was broken before and it's broken now. Can you just embed as encoded hex data?
 - You write 'Consider using the spatial attention, which improves the performance when studying local Hamiltonians (see Ref. arXiv:2602.02665)' Can you expand the discussion to an at least 3-4 lines paragraph about what it's doing? And why it's a good thing and when it might not be a good thing (I imagine if one has long range couplings for example). Also define the variable (gamma0)
 - Also add the mathematical formula please. can you define what is gamma0 and what is a reasonable value for it (explicitly quoting your paper for the value if it comes from there to avoid future confusion if someone else suggests another value)?
 - Can you rename gamma0 to gamma or there's a reason not to?

Remarks on the example file:
 - please copy paste the whole definition of the model in there. Also rename the example as `Examples/architectures/transformer_heisenberg_square_......py`

Remarks on the transformer in netket:
I ask that you remove `netket/models/transformer.py`.

As a general statement I don't want to merge any 'real' architecture inside of netket again. There's largely 2.5 reasons:  
- one is maintenance, because it's hard to keep the architecture called 'transformer.ViT' in netket and ensure checkpoint and serialised state backward compatibility. Essentially, either the architecture is frozen in time, as hugginface does, or we are bound to break some stuff in the future. We already had a few complications with GCNN, and I don't want to repeat this again. 

I think huggingface approach is great, just that netket is not the place where one should have the model repository because 
- second is 'biasing research'. Until 2024 there's a huge correlation between netket papers and RBM/GCNN usage, and a huge correlation between jax-vmc and CNN usage. Basically because netket does not provide a built in CNN wf but does for RBM/GCNN. I don't want this to happen again with the ViT. I think it's a great architecture, but wouldn't want to have it as 'an easy thing to do' from within netket. Admittedly we should not have done the same with GCNN or ARNN but the past is the past.

- Third, and most importantly, I think a thing that huggingface got right is the 'single file architecture' idea (https://huggingface.co/blog/transformers-design-philosophy ). This encourages people to test/modify/change architectures. Shoving architectures inside of netket is discouraging people from doing that. Having in a repo a single file that does that, encourages it. 

Therefore I'd encourage you to create a separate repository that people can `uv add` with a barebone importable module with the network. 
(This is also what I think should be done for your paper under review. )

NetKet can have specific layers or tools that benefits general architectures in its `nk.nn` submodule, functions to compute distances between sites (which I think we already have), but not architectures. We can advertise them, but they either belong to their own repo or to something that someone still has to create, a nqs-architectures depot a la transformers.

gcarleo:
@PhilipVinc I do not disagree but we must be consequential then and also remove all other models

attila-i-szabo:
@PhilipVinc I disagree with the general sentiment. Most people who would like to do NQS are not experts in the same way as we are, working on Netket, knowing its quirks in detail, and used to implementing complex architectures. Also, most physicists want to do physics rather than tweak code. 

This means that efficient implementations of successful NQS architectures that can be used with Netket with minimal friction is a very large part of the appeal. Otherwise, we'll simply push users to use other libraries that can do what Netket can, but also provide architectures.

> * 'biasing research'. Until 2024 there's a huge correlation between netket papers and RBM/GCNN usage, and a huge correlation between jax-vmc and CNN usage. Basically because netket does not provide a built in CNN wf but does for RBM/GCNN. I don't want this to happen again with the ViT. I think it's a great architecture, but wouldn't want to have it as 'an easy thing to do' from within netket.

Again, I think you have the logic backwards here - if Netket hadn't provided those architectures (or _some_ architectures), most of those papers would simply have been done with another library. Likewise, without an easily accessible ViT architecture, people will just not use it, or switch to another library if they're very keen on this architecture...

> We can advertise them, but they either belong to their own repo or to something that someone still has to create, a nqs-architectures depot a la transformers.

I don't insist on keeping the architectures built into Netket (although if we keep them frozen, it shouldn't be too hard - the issues with GCNN came more from iterating on the layers themselves). However, if we don't, we definitely need to provide a depot, invite users (not just us) to deposit their new architectures, and advertise it almost as part of the library.

gcarleo:
@attila-i-szabo yes that's why I am saying, we either drop all models and create a repo under the netket organization (would be very much happy if that happens.... was something always we wanted to do) or add this one to netket too, like all other successful models. if we decide to not include this one specifically, it would be an inconsistent/ not well motivated choice in my opinion

gcarleo:
doing the model repository is something relatively easy to do, provided one pins the netket version they work with. one can then add automatic tests / actions that automatically bump up compatible versions if all tests for a given model pass. it's a simple project for Claude code...

PhilipVinc:
I agree that we should be consequential, but I don't think the consequence has to be "merge every successful architecture into `netket.models` until we have removed the old ones".
I think there's a difference between (i) breaking existing code by removing RBM/GCNN/ARNN today and (ii) deciding not to add new long-term API surface for models.

Once an architecture is in `netket.models`, users reasonably expect it to be documented, tested, maintained, checkpoint-compatible, and stable across NetKet releases.
That is a stronger commitment than "here is a good architecture that works with NetKet". For this PR, and differently from the GCNN, the model is not integrated with NetKet abstractions anyway: it does not use NetKet lattice information, does not validate that the input geometry/patching is consistent, does not give NetKet-style error messages, and is not documented at the same level as the rest of the public API. 

So the user-facing benefit of putting it in core is mostly just the import path.

I agree with @attila that users need low-friction access to good architectures. 
But I don't think "inside core NetKet" is the only low-friction option. 
A small package/repository would be just as easy for users.

```bash
uv add git+https://github.com/RiccardoRende/vit-nqs
python
>>>from vit_nqs import ViT
```

and what i was proposing in my past message was to link it in the tutorial, examples, and possibly a “known architectures” page.
That gives users an easy path without increasing NetKet's API surface and without committing NetKet itself to specific models.

It could also be in the NetKet organisation if Riccardo/Luciano agree to write tests, document it, and maintain it in the near future.

> I don't insist on keeping the architectures built into Netket (although if we keep them frozen, it shouldn't be too hard - the issues with GCNN came more from iterating on the layers themselves). However, if we don't, we definitely need to provide a depot, invite users (not just us) to deposit their new architectures, and advertise it almost as part of the library.

yeah, it's a few months I've been playing with ways to do this reasonably well. I tried to get people to work with me on it but nobody has the time or desire to, and my time is limited.
And I don't think that Claude can do a decent job at doing something that works for years to come.

A few weeks ago I did some progress and I have something that 75% works. Hopefully I'll be able to finish it before my death

gcarleo:
honestly I think we are overcomplicating things. having a working repository with working models can be done in a matter of a few hours. I don't understand what the main problem would be, it's just a collection of jax models that are compatible with some frozen netket repository. that would be good enough for most people, no need to die before that.  I can have a proof of principle done this week.

gcarleo:
not to mention that having a separate repo for models would remove a lot of complications that we have now when we want (for some reason) add a new model and make it generally available. ideally, that should be used as a low barrier repository where everybody can push whatever they think works. essentially like huggingface

attila-i-szabo:
There is a downside of the huggingface approach for us, though - for physics we are rarely interested in reproducing exactly what somebody did already, but rather use their methods (mutatis mutandis) for our new problems. This might be just as hard as writing the model from scratch if you just get somebody else's code that was never designed for this kind of modification. (The beauty of the GCNN code in netket is precisely this, you can very easily repurpose it to a spin model on any lattice, while the code in this PR seems tailored for the square lattice - that's also why I was suggesting using the existing Netket convolution layers...)

For sure, we should encourage people to open up their code (a la huggingface) but I think it'd be valuable for us to maintain a middle level (within Netket org, possibly not in Netket library) with architectures that we consider successful and good to expose to users, written in a general enough way. I'd definitely put ViT in this category, and I'm also open to moving GCNN/ARNN/etc. to such a repo/repos.

> without committing NetKet itself to specific models.

I still don't see the problem with this. We don't really commit Netket to anything, users can write any ansatz they like. If they don't want to (which is probably the majority of users), we can at best push them away from Netket, not make them write new ansätze...

PhilipVinc:
> (The beauty of the GCNN code in netket is precisely this, you can very easily repurpose it to a spin model on any lattice, while the code in this PR seems tailored for the square lattice - that's also why I was suggesting using the existing Netket convolution layers...)

That's why I was saying that NetKet it's not the place to merge this. The code in this PR has many assumptions (also does not play well with multiple sites per unit cell). 
I did write a code for a ViT-like wavefunction based on netket's lattices, and it took a bit to get it right for arbitrary lattices. I think some pieces are in the last few papers we did. But we never cleaned it up properly, and I don't want NetKet to commit to something.

Part of the problem is what we are using to save MCStates and neural networks for 'mid term storage'.
It's been playing well for the last 2 years, but a constant pain point is that when someone adds/changes some fields of an object that someone has saved, we need to write a conversion function. For most of netket this is not a problem, as it rarely changes the internals, but for models we did have many issues and had to learn to work in a certain way to be able to save states reliably.

As netket needs to be constantly upgraded to keep up with jax changes/different gpus, I don't want to oblige people to upgrade the models as well. It should be a separate repo.

>  I'd definitely put ViT in this category, and I'm also open to moving GCNN/ARNN/etc. to such a repo/repos.

indeed. +1

PR #2214, opened 2026-03-25
Allow different probabilities for different moves in `ExchangeRule`

attila-i-szabo:
There are situations where allowing further-range exchanges in Monte Carlo is useful, but you don't normally want to use all of them at the same probability. This PR is therefore to allow weighting the different proposals allowed by `ExchangeRule`

**Usage 1:** if `clusters` are specified manually, `probabilities` specifies the relative probability of proposing each one of them:
```python
sampler = nk.sampler.MetropolisExchange(hilbert, clusters=[(1,2),(2,3),(3,4)], probabilities=[1,2,3])
```

**Usage 2:** if `graph` and `d_max` are specified, `probabilities` specifies the relative probabilities of clusters with graph distance 1,2,...
```python
sampler = nk.sampler.MetropolisExchange(hilbert, graph=graph, d_max=2, probabilities=[1.0, 0.5])
```

**Miscellaneous improvements:**
* lowered the `compute_clusters` function to numpy instead of big nested loops
* expanded the documentation of `MetropolisExchange` to cover every argument

### DIFF
--- a/netket/sampler/rules/exchange.py
+++ b/netket/sampler/rules/exchange.py
@@ -83,6 +83,11 @@ class ExchangeRule(MetropolisRule):
     The Exchange rule will swap the two sites of a random row of this
     matrix at every Metropolis step.
     """
+    probabilities: jax.Array
+    r"""1-Dimensional array :math:`p_{i}` of length :math:`N_\text{clusters}`
+    containing the relative probability of choosing each particular cluster
+    for the exchange update.
+    """
 
     def __init__(
         self,
@@ -90,6 +95,7 @@ class ExchangeRule(MetropolisRule):
         clusters: list[tuple[int, int]] | None = None,
         graph: AbstractGraph | None = None,
         d_max: int = 1,
+        probabilities: list[float] | None = None,
     ):
         r"""
         Constructs the Exchange Rule.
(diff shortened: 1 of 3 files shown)


netket/sampler/rules/exchange.py
```diff
@@ -127,11 +151,11 @@ def transition(rule, sampler, machine, parameters, state, key, σ):
         @jax.vmap
         def _update_samples(key, σ, hoppable_clusters):
             # pick a random cluster, taking into account the mask
-            n_conn = hoppable_clusters.sum(axis=-1)
+            n_conn = hoppable_clusters @ rule.probabilities
```
PhilipVinc, line 154: This does not seem correct? I'd assume that the default is for rule.probabilities to be 1/Nconn for each entry?

PhilipVinc: ah, maybe it's no longer an `n_conn` but it is still the right weight?

attila-i-szabo: Yes, it's more like `total_w_conn`. Note that the default init for probabilities is ones, so it would recover the original behaviour. (rule.probabilities are only proportional to the actual probabilities, since you have to normalise them dynamically.)

attila-i-szabo: @PhilipVinc  any specific changes I should make on this?

PhilipVinc: can you rename it? I was very confused when I read this line.

PhilipVinc (approved):
I also changed it to avoid storing the probabilities by default (the past setup) and avoid doing the product. Minor thing, but this sort of thing can quickly get expensive on GPUs when you combine many different rules.

PR #2225, opened 2026-04-14
Change times in pbar for TDVP

lindamauron:
Currently, the progressbar in `TDVP` drivers only works correctly if `t0 = 0`: 
- if `t0 > 0`, percentages start at `t0 / T` instead of `0` breaking prediction time
- if `t0 < 0`, percentages are negative and prediction time breaks too

These small changes fix this issue by shifting the origin correctly in the progressbar, displaying iterations from `0` to `T` (`T` is the total evolution time). No logic was changed in the driver, only the progressbar is affected.

### DIFF
--- a/netket/experimental/driver/tdvp_common.py
+++ b/netket/experimental/driver/tdvp_common.py
@@ -303,10 +303,11 @@ class TDVPBaseDriver(AbstractVariationalDriver):
         callbacks = to_iterable(callback)
         callback_stop = False
 
-        t_end = np.asarray(self.t + T)
+        t_init = np.asarray(self.t)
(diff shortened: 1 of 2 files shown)


PhilipVinc:
Thanks Linda,
Don't worry about failing tests, those are unrelated.

Could you also add a simple regression test to ensure this does not break in the future?

lindamauron:
I added `test_pbar_starts_zero_ends_hundred`, testing that the `pbar` has no negative percentages (happened for `t0<0`) and no sudden jumps (happened for `t0>0`). I checked that the old code raised errors while the new one passes them.

PR #2229, opened 2026-04-23
Fix QGTOnTheFly chunked solve silently returning n_devices × correct under sharding

### LINKED ISSUE #2228: QGTOnTheFly chunked solve returns x_ref / n_devices when chunk_size < samples_per_shard (with pdf)
## Summary

Under `NETKET_EXPERIMENTAL_SHARDING_CPU > 1`, solving a weighted `QGTOnTheFly` with `chunk_size < samples_per_shard` (where `samples_per_shard = n_samples // jax.device_count()`) returns a result that is **exactly** `x_ref / n_devices` instead of `x_ref`, silently corrupting any downstream SR update. Setting `chunk_size >= samples_per_shard` or `chunk_size=None` gives the correct answer.

The factor is exactly `1 / n_devices` for every `chunk_size` below the per-shard size, which strongly suggests a missing shard-axis reduction in the chunked multi-chunk-per-shard code path when a `pdf` is provided.

## Versions

- netket `3.22.0.dev3`
- jax `0.10.0`
- flax `0.12.7`
- Python `3.13.5`
- macOS (Darwin 23.6.0), CPU backend

## Minimal reproducer

Standalone, uses only stock NetKet primitives:

```python
"""NETKET_EXPERIMENTAL_SHARDING_CPU=4 python mre_qgt_chunk.py"""
from __future__ import annotations

import os
os.environ.setdefault("NETKET_EXPERIMENTAL_SHARDING_CPU", "4")

import jax
import jax.numpy as jnp
import numpy as np
import netket as nk
from jax.flatten_util import ravel_pytree
from netket.optimizer.qgt.qgt_onthefly import QGTOnTheFly_DefaultConstructor
from netket.optimizer.solver import pinv_smooth


def solve_qgt(vstate, samples, pdf, rhs, chunk_size):
    qgt = QGTOnTheFly_DefaultConstructor(
        vstate._apply_fun, vstate.parameters, vstate.model_state,
        samples, pdf=pdf, holomorphic=True, chunk_size=chunk_size,
    )
    x, _ = qgt.solve(pinv_smooth(rcond=1e-8, rcond_smooth=1e-8), rhs)
    x_flat, _ = ravel_pytree(x)
    return x_flat


def main():
    n_sites = 8
    n_samples = 4096
    alpha = 2
    seed = 0

    hi = nk.hilbert.Spin(s=0.5, N=n_sites)
    model = nk.models.RBM(alpha=alpha, param_dtype=complex)

    sampler = nk.sampler.MetropolisSampler(
        hi, rule=nk.sampler.rules.LocalRule(),
        n_chains=max(16, jax.device_count() * 4),
    )
    vstate = nk.vqs.MCState(sampler, model, n_samples=n_samples, seed=seed)

    samples = jnp.asarray(
        np.asarray(vstate.samples).reshape(-1, n_sites)
    )
    rng = np.random.default_rng(seed)
    omega = rng.uniform(0.1, 1.0, size=samples.shape[0])
    pdf = jnp.asarray(omega / omega.sum())

    params_flat, unravel = ravel_pytree(vstate.parameters)
    rhs_flat = rng.normal(size=params_flat.size) + 1j * rng.normal(
        size=params_flat.size
    )
    rhs = unravel(jnp.asarray(rhs_flat))

    n_dev = jax.device_count()
    per_shard = n_samples // n_dev
    print(f"jax.device_count()={n_dev}, n_samples={n_samples}, "
          f"samples_per_shard={per_shard}")
    print(f"{'chunk_size':>14} {'max|Δ|':>14} {'|x_cs|/|x_ref|':>18}")
    print("-" * 55)

    x_ref = solve_qgt(vstate, samples, pdf, rhs, None)
    for cs in [None, per_shard * 2, per_shard, per_shard // 2,
               per_shard // 4, 128, 64]:
        if cs is None:
            x = x_ref
        else:
            x = solve_qgt(vstate, samples, pdf, rhs, cs)
        diff = float(jnp.max(jnp.abs(x - x_ref)))
        ratio = float(jnp.linalg.norm(x) / jnp.linalg.norm(x_ref))
        print(f"{str(cs):>14} {diff:>14.3e} {ratio:>18.5f}")


if __name__ == "__main__":
    main()
```

## Observed output

```
jax.device_count()=4, n_samples=4096, samples_per_shard=1024
    chunk_size         max|Δ|     |x_cs|/|x_ref|
-------------------------------------------------------
          None      0.000e+00            1.00000
          2048      0.000e+00            1.00000
          1024      0.000e+00            1.00000
           512      7.986e+02            0.25000
           256      7.986e+02            0.25000
           128      7.986e+02            0.25000
            64      7.986e+02            0.25000
```

Every row with `chunk_size < 1024` gives `|x_cs| / |x_ref| = 0.25 = 1 / n_devices`, and all such rows agree on the wrong vector (same `max|Δ|`), so the output is a deterministic rescaling of the correct result — not a stochastic error.

## Expected behavior

All rows should agree to ~1e-12 regardless of `chunk_size`, as is the case when `chunk_size >= samples_per_shard`.

## Likely source

`netket/optimizer/qgt/qgt_onthefly_logic.py::_Odagger_DeltaO_v` (~L135–145):

```python
def _Odagger_DeltaO_v(forward_fn, params, model_state, samples, v, chunk_size, pdf=None):
    w = _O_jvp(forward_fn, params, model_state, samples, v, chunk_size)
    if pdf is None:
        w = w * (1.0 / samples.shape[0])
        w = w - jnp.mean(w)  # w/ JAX sharding
    else:
        w = pdf * (w - pdf @ w)
    res = _OH_w(forward_fn, params, model_state, samples, w, chunk_size)
    return res
```

`_O_jvp` and `_O_vjp` are each wrapped with `sharding_decorator(..., reduction_op_tree=jax.lax.psum)` (for `_O_vjp`) or plain `sharding_decorator` (for `_O_jvp`), so inside each you see the per-shard slice of `samples` and `v`. The `scanmap`+`chunk(samples, chunk_size)` inside those functions splits the per-shard slice into chunks; `scan_append` and `scan_reduce` accumulate over those chunks (per-shard), and the outer `sharding_decorator` is responsible for stitching the shards back together.

At `chunk_size >= samples_per_shard`, `chunk()` produces exactly one chunk per shard and the accumulation reduces to a no-op — the path goes through what is effectively the "1 chunk" case and the outer `psum`/reshape combines shards correctly.

At `chunk_size < samples_per_shard`, the per-shard scan produces multiple partial results that get accumulated via `scan_append`/`scan_reduce` **before** the decorator's shard reduction runs, and the cross-shard normalization/`pdf @ w` step (outside the `@sharding_decorator` wrappers) sees a `w` whose scaling has already been divided by `n_devices` but hasn't been `psum`ed. The precise factor of `1 / n_devices` in the reproducer matches exactly this: the inner `_O_jvp` result is implicitly an average per shard rather than a sum, so `pdf @ w` produces the wrong normalization and every subsequent step carries the `1/n_devices` factor.

Suggested repro fix to investigate: make sure the `pdf * (w - pdf @ w)` expression in `_Odagger_DeltaO_v` is evaluated on a correctly shard-reduced `w`, either by wrapping the whole `_Odagger_DeltaO_v` in a `sharding_decorator` (with appropriate `reduction_op_tree`) or by adjusting the chunked inner reduction so that its output is shard-summed before the `pdf`-weighted step. The non-`pdf` branch relies on `jnp.mean(w)` (which JAX + NamedSharding handles correctly), so it doesn't exhibit the bug — but the `pdf` branch's manual `pdf @ w` has to be either explicitly `psum`-reduced or executed inside the same shard-map as the rest.

## Workaround

`chunk_size = None` or `chunk_size >= n_samples // jax.device_count()` (i.e. at most one chunk per shard). In practice this means chunking inside a shard is currently unusable when a nontrivial `pdf` is passed.

gcarleo:
## Summary

Fixes the silent correctness bug in `QGTOnTheFly` reported in #2228. Under `NETKET_EXPERIMENTAL_SHARDING=1`, passing `chunk_size` smaller than `n_samples / jax.device_count()` routed the solve through `mat_vec_chunked_factory`, which returned a result scaled by exactly `jax.device_count()` — no error, no warning, just a factor-of-`n_devices` error in every `S v` product and thus in every SR step.

## Root cause

The internal `_O_vjp` was wrapped with `reduction_op_tree=jax.lax.psum` on the outer `sharding_decorator`, but did not mark the replicated `params`/`model_state` arguments with `pvary_args_tree`. As a result, `jax.vjp` inside the `shard_map` implicitly `psum`-reduced the gradient wrt those replicated inputs (see the discussion at https://github.com/jax-ml/jax/discussions/29608), and this compounded with the explicit `reduction_op_tree=jax.lax.psum` — so the chunked result was `n_devices × correct`.

The exact same pattern (differentiating through a replicated input inside a shard_map + explicit outer `psum`) is already handled correctly in `netket.jax._vjp_chunked`, which uses `pvary_args_tree` on the replicated, non-`nondiff` args and specifically comments:

> "jax will psum in every vjp, but we want to do one by hand at the end so we need to set the sharded args we differentiate as varying"

This PR ports that fix to `_O_vjp`.

## Fix

```python
@partial(
    sharding_decorator,
    sharded_args_tree=(False, False, False, True, True, False),
    reduction_op_tree=jax.lax.psum,
    pvary_args_tree=(False, True, True, False, False, False),  # <- added
)
def _O_vjp(forward_fn, params, model_state, samples, w, chunk_size):
    ...
```

Single-device runs (where `mat_vec_chunked_factory` is never selected, and where `shard_map` is not active anyway) are unaffected.

## Regression test

Added `test/sharding/test_sharding.py::test_qgt_onthefly_chunked_matches_unchunked`. It builds `QGTOnTheFly` at several chunk sizes (including `< samples_per_shard`, which is what triggers the bug) with and without a nontrivial `pdf`, and asserts the chunked and unchunked `S @ v` match to 1e-10. Without this PR the test fails with `ratio ≈ n_devices` at every `chunk_size < samples_per_shard`.

## Verification

Before the fix, the MRE from the issue printed

```
    chunk_size         max|Δ|     |x_cs|/|x_ref|
-------------------------------------------------------
          None      0.000e+00            1.00000
          1024      0.000e+00            1.00000
           512      7.986e+02            0.25000   <- bug
           256      7.986e+02            0.25000
           128      7.986e+02            0.25000
```

After the fix:

```
           512      1.359e-10            1.00000
           256      1.539e-10            1.00000
           128      1.798e-10            1.00000
            64      1.384e-10            1.00000
```

All existing tests still pass locally:
- `test/optimizer/test_qgt_itersolve.py`: 362 passed
- `test/sharding/test_sharding.py` (excluding the pre-existing unrelated `test_loggers[HDF5Log]` that needs `h5py`): 31 passed, including the two new `test_qgt_onthefly_*` variants.

Closes #2228.


gcarleo:
failing tests unrelated, we should fix them elsewhere

gcarleo:
looks like a rather critical bug to me, would love to merge this fix

inailuig:
lgtm, but I would first fix the failing tests, then rebase this on top of that

Tests fail because of recent changes in jax; We need to update `.vma` to `.manual_axis_type.varying`

PhilipVinc:
I'll let @inailuig comment here. If tests pass it's good for me.
However I'd suggest rewriting the test. as it's written I don't understand it.

As a proof of concept I gave this prompt to claude
```bash
❯ Look at the test added in this PR currently checked out. i think the test is very
  bad, because it wants to compare code between sharding and no sharding but ends up
  doing a mess and the human me cannot follow/understand what it's doing.
  A good test is clear, has transparent meaning and is short/simple.

  I think it can be heavily simplified by simply constructing an MCState and buildign
  the QGT under a 'jax.sharding.with_mesh' or similar context and using an empty mesh
  with only 1 GPU vs the default case of many GPUs.

  Can you do that? before changing the test, do a few experiments to understand how to
   use the context, assuming we execute the test with many gpus (this happens by
  default because +JAX_NUM_CPU_DEVICES is set) .

 Also make sure to use jax.tree.map instead of zip through flattened lists, and keep the test simple by simply comparing a single calculation between single-gpu mesh and multi-gpu mesh.
```
which gives this (more comprehensible) test
```python
    L = 4 
    ma = nk.models.RBM(alpha=1, param_dtype=np.complex128)

    full_mesh = jax.sharding.get_mesh()
    single_mesh = jax.make_mesh((1,), ("S",), axis_types=full_mesh.axis_types[0])

    n_states_per_shard = hi.n_states // jax.device_count()
    chunk_size = max(n_states_per_shard // 2, 1)

    # Shared parameters as plain numpy — no device affiliation
    params_np = jax.tree.map(np.asarray, nk.vqs.FullSumState(hi, ma).parameters)

    # Reference: single-device mesh — shard_map runs on 1 device, no psum scaling possible
    jax.sharding.set_mesh(single_mesh)
    try:
        vs_ref = nk.vqs.FullSumState(hi, ma)
        vs_ref.parameters = jax.device_put(params_np, NamedSharding(single_mesh, P()))
        S_ref = nk.optimizer.qgt.QGTOnTheFly(vs_ref, chunk_size=chunk_size, holomorphic=True)
        res_ref = jax.tree.map(np.asarray, S_ref @ vs_ref.parameters)
    finally:
        jax.sharding.set_mesh(full_mesh)

    # Test: full mesh
    vs = nk.vqs.FullSumState(hi, ma)
    vs.parameters = jax.device_put(params_np, NamedSharding(full_mesh, P()))
    S = nk.optimizer.qgt.QGTOnTheFly(vs, chunk_size=chunk_size, holomorphic=True)
    res = S @ vs.parameters

    jax.tree.map(lambda r_ref, r: np.testing.assert_allclose(r, r_ref, rtol=1e-10, atol=1e-10), res_ref, res)
```

gcarleo:
feel free to add the test you propose

netket/optimizer/qgt/qgt_onthefly_logic.py
```diff
@@ -112,6 +112,15 @@ def __O_jvp(forward_fn, params, model_state, samples, v):
...
     sharded_args_tree=(False, False, False, True, True, False),
     reduction_op_tree=jax.lax.psum,
+    # jax.vjp inside a shard_map implicitly psum's the gradient wrt any
+    # replicated input (see https://github.com/jax-ml/jax/discussions/29608).
+    # Without pvary on (params, model_state), the implicit psum inside vjp
+    # compounds with the explicit psum from reduction_op_tree above, giving
+    # a result scaled by jax.device_count() (silently wrong).  Marking
+    # params and model_state as varying tells JAX the backward pass should
+    # not psum them; the outer reduction_op_tree then does the one psum we
+    # actually want.  Same pattern as netket.jax._vjp_chunked.
```
PhilipVinc: This comment is AI slop that is long and incomprehensible. I would prefer something that a human clean up.

```suggestion
    # The gradient w.r.t. params must be psum'd exactly once over devices
    # (via reduction_op_tree), but jax.vjp inside shard_map implicitly psum's
    # replicated inputs too.  pvary suppresses the implicit psum.
    # (Same pattern as netket.jax._vjp_chunked; see jax-ml/jax#29608.)
```

CHANGELOG.md
```diff
@@ -97,6 +97,7 @@
   Those bugs had apparently been sitting there unnoticed for quite a while.
 * The `Midpoint` and `Heun` tableaus now expose consistent embedded lower-order formulas [PR #2210](https://github.com/netket/netket/pull/2210).
 * Calling {meth}`netket.vqs.MCMixedState.expect_and_grad` with a physical operator now raises a dedicated error with guidance on defining a custom dispatch rule [commit e4fc4a965](https://github.com/netket/netket/commit/e4fc4a96575b6bd61a723e76f677148def50e9b8).
+* Fixed a silent correctness bug in {class}`netket.optimizer.qgt.QGTOnTheFly` under `NETKET_EXPERIMENTAL_SHARDING=1`: when the ``chunk_size`` argument was strictly smaller than the number of samples per shard, the chunked S-matrix-vector product returned a result scaled by a factor of `jax.device_count()`. The internal `_O_vjp` lacked `pvary_args_tree` for the replicated `params`/`model_state` arguments, so `jax.vjp`'s implicit backward-pass `psum` compounded with the explicit `reduction_op_tree=jax.lax.psum` (see the same pattern already used in {func}`netket.jax._vjp_chunked`). The fix only affects multi-device sharded runs; single-device results were already correct.
```
PhilipVinc:
```suggestion
  * Fixed a silent correctness bug in {class}`netket.optimizer.qgt.QGTOnTheFly` with sharding: when chunking was active, the S-matrix-vector product was
  silently scaled by `jax.device_count()`. The fix adds `pvary_args_tree` on `params` so that `jax.vjp`'s implicit backward-pass `psum` does not compound with
  the explicit `reduction_op_tree` psum [#2229](https://github.com/netket/netket/pull/2229).
```

PhilipVinc (approved):

PR #2230, opened 2026-04-24
update .vma -> .manual_axis_type.varying

inailuig:


inailuig:
do we need some backward compatibility?

PhilipVinc:
This was on my radar, but it should be backwards compatible with pre jax 0.9 because I would not want to drop it just yet.

PhilipVinc:
@inailuig you could have a look at https://github.com/netket/netket/tree/pv/fixes-0.10 .
It was claude-generated, so the method he proposed (see below) is terrible but could be cleaned up.
I never got around to finishing it but maybe you could? all those getattr make me suffer.

```python
def aval_varying_axes(x) -> tuple:
    """
    Return the manual mesh axes over which a traced value is varying.
    JAX <= 0.9 exposed this as ``jax.typeof(x).vma``. JAX 0.10 replaced that
    attribute with ``manual_axis_type.varying``.
    """
    aval = jax.typeof(x)

    if hasattr(aval, "vma"):
        return tuple(aval.vma)

    manual_axis_type = getattr(aval, "manual_axis_type", None)
    if manual_axis_type is None:
        return ()

    varying = getattr(manual_axis_type, "varying", ())
    if not varying:
        return ()

    aval_sharding = getattr(aval, "sharding", None)
    mesh = getattr(aval_sharding, "mesh", None)
    axis_names = getattr(mesh, "axis_names", ())
    ordered_axes = tuple(axis for axis in axis_names if axis in varying)

    if len(ordered_axes) == len(varying):
        return ordered_axes

    return ordered_axes + tuple(axis for axis in varying if axis not in axis_names)

```

PR #2234, opened 2026-05-13
Add ladder operators to PauliStrings

### LINKED ISSUE #1206: Ladder operators in PauliStrings

Cooooked:
## Summary

- Allow `PauliStrings` operator strings to contain `+` and `-` ladder operators.
- Treat ladder operators as state flips with zero matrix elements when the operator cannot act on the current local state.
- Support both JAX and Numba PauliStrings paths.
- Expand ladder operators into Pauli terms for symbolic `@` multiplication.
- Add focused regression coverage for dense matrices, Z phases, and ladder-operator multiplication.

Fixes #1206.

## Testing

- `git diff --check`

I could not run the Python test suite in this workspace because no usable Python interpreter is installed here. Opening as draft so repository CI can validate the patch before review.

### DIFF
--- a/netket/operator/_pauli_strings/base.py
+++ b/netket/operator/_pauli_strings/base.py
@@ -31,7 +31,7 @@ from netket.utils.numbers import dtype as _dtype, is_scalar
 from .._abstract_operator import AbstractOperator
 from .._discrete_operator import DiscreteOperator
 
-valid_pauli_regex = re.compile(r"^[XYZI]+$")
+valid_pauli_regex = re.compile(r"^[XYZI+-]+$")
 
 
 def _standardize_matrix_input_type(op):
@@ -97,7 +97,7 @@ def canonicalize_input(hilbert: AbstractHilbert, operators, weights, *, dtype=No
     if not consistent:
         raise ValueError(
             """Operators in string must be one of
-            the Pauli operators X,Y,Z, or the identity I"""
+            the Pauli operators X,Y,Z, ladder operators +,-, or the identity I"""
         )
 
     weights = _standardize_matrix_input_type(weights)
@@ -538,7 +538,6 @@ def _split_string(s):
     return [x for x in str(s)]
 
 
-@jit(nopython=True)
 def _make_new_pauli_string(op1, w1, op2, w2):
     """Compute the (symbolic) tensor product of two pauli strings with weights
     Args:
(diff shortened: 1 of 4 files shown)


PhilipVinc:
are you a human or a bot?

gcarleo:
do a quick Turing test

Cooooked:
hi Netket/Netket this was a bot,  no wories

PhilipVinc:
I wory

PR #2238, opened 2026-05-25
Make MCState.local_estmator extensible, add builtin delta-method propagation to compute variance of (nonlinear) observables. 

PhilipVinc:
Makes writing correct error/variance estimation of observables easy, and automatically extends support of `èxpect_to_precision ` to more complex objects...


gcarleo:
can you provide an example of where this is now used / could be used?

PhilipVinc:
Yes, sorry! A nice example is the V-score, where the error bar is nontrivial to compute:

```python
  class VScore(AbstractObservable):
      def __init__(self, operator, *, trace_diagonal: float):
          super().__init__(operator.hilbert)
          self._operator = operator
          self._operator_squared = operator @ operator
          self._trace_diagonal = float(trace_diagonal)

  @local_estimators.dispatch
  def _(vstate: MCState, op: VScore, chunk_size):
      h_loc  = compute_local_estimator(vstate, op._operator).real
      h2_loc = compute_local_estimator(vstate, op._operator_squared).real
      td = op._trace_diagonal
      return LocalEstimatorsBatch(
          # the two estimators that are accumulated when you add more samples
          data=jnp.stack([h_loc, h2_loc], axis=-1).reshape(n_chains, -1, 2),
          # how you combine the means of the estimators above to get the Vscore
          combinator=lambda mu: (mu[1] - mu[0]**2) / (mu[0] - td)**2,
      )
```

Another example: connected correlators, which we don't support today but become trivial with this:

```python
  class Covariance(AbstractObservable):
      """Cov(A, B) = <AB> - <A><B>"""
      def __init__(self, op_A, op_B):
          super().__init__(op_A.hilbert)
          self._op_A, self._op_B, self._op_AB = op_A, op_B, op_A @ op_B

  @local_estimators.dispatch
  def _(vstate: MCState, op: Covariance, chunk_size):
      a_loc  = compute_local_estimator(vstate, op._op_A).real
      b_loc  = compute_local_estimator(vstate, op._op_B).real
      ab_loc = compute_local_estimator(vstate, op._op_AB).real
      return LocalEstimatorsBatch(
          data=jnp.stack([a_loc, b_loc, ab_loc], axis=-1).reshape(n_chains, -1, 3),
          combinator=lambda mu: mu[2] - mu[0] * mu[1],
      )
```

Note that, [as shown in the doc page preview](https://netket--2238.org.readthedocs.build/en/2238/advanced/custom-operators/index.html), you can already define such operators today via `@expect.dispatch` returning a Stats. 
While this PR makes those operators easier to define, that's not the key point.

The important point is that if you define them according to the existing interface you can only define how to compute them for a given number of samples (`MCState.expect`) but they won't work with `MCState.expect_to_precision(observable, rtol=1e-3)` which performs a single pass accumulation over samples.

The problem is that, right now, `expect_to_precision` assumes a linear accumulation like 

$$
\langle O \rangle = \mathbb{E}_x[O_\mathrm{loc}(x)]
$$ 

but V-score, connected correlators, etc. don't have that form.
This PR makes it possible to express arbitrary nonlinear accumulations structures like that of Vscore which is  

$$
  \langle V \rangle = \frac{\mathbb{E}_x[H_\mathrm{loc}^2(x)] - \mathbb{E}_x[H_\mathrm{loc}(x)]^2}{\left(\mathbb{E}_x[H_\mathrm{loc}(x)] -
  e_0\right)^2}
$$

To finish, another example, a bit more contrived, is the Quantum Fisher/Generalized Susceptivity in foundation-style NQS. To estimate the QFI in those settings up to a set precision (which is something very useful, especially because the QFI has wildlyvarying variance as you move in  the parameter space) we have been using a code somewhat like this, which we have been copy pasting in tens of places so I'd rather it be in one place.

gcarleo:
ok I See makes sense ..! but maybe the simple solution is doing bootstrap or jackknife to compite error bars on nonlinear quantities, I think we should maybe also stabilize that at some point, shouldn't be too hard it the values of the observables are available for all configurations

PhilipVinc:
simple but expensive ? You need to have more samples/do subsampling to do it effectively?
In any case, I think it would be useful to have, one day. Especially for gradient/natural gradient. it's always a mess to do and it's not an uncommon thing to do.

In any case it's quite orthogonal to this. 
This PR is just the standard error propagation formula, essentially, which is good enough in most cases.

gcarleo:
no, you don't need to generate new samples it's just resampling what you already have, it's quite cheap I think. but I'm it's orthogonal I agree, this one here can be merged

PR #2245, opened 2026-07-04
Fix/tensorboard jax scalars

### LINKED ISSUE #2244: TensorBoardLog silently drops `acceptance` (and possibly other raw-array scalars) on 3.22
**Environment**
- NetKet version: 3.22.3
- JAX version: 0.10.2
- Python version: 3.11.15

## Description
 
`acceptance` does not appear in TensorBoard event files, even though it is present in `log_data` at every step (confirmed via a custom `AbstractCallback.on_step_end`). Other quantities such as `Energy` (which arrive wrapped in `netket.stats.mc_stats.Stats`) log correctly.
 
Printing the type of the value shows a difference between the two:
```
type(log_data["acceptance"])  ->  <class 'jaxlib._jax.ArrayImpl'>
type(log_data["Energy"])  ->  <class 'netket.stats.mc_stats.Stats'>
```
My hypothesis is that `TensorBoardLog`'s write path branches on "is this a `Stats`/`History`-like object" or a small set of known Python scalar types (`int`/`float`/`np.number`), and a `jaxlib._jax.ArrayImpl` falls through both checks and is silently skipped, with no warning or error raised.
 
I have not tested this against NetKet versions prior to 3.22, so I can't say whether this is a regression from the recent callback rewrite or has always been the case.
 
## Steps to reproduce
 
See attached minimal example below. Run it, then inspect the log directory directly (bypassing the TensorBoard browser UI, to rule out a display/tag-filter issue):
```
tensorboard --inspect --logdir=issue_logdir
```
`acceptance` is absent from the list of tags, despite being printed by the debug callback at every step.
 
```python
import netket as nk

# Tiny 4-site transverse-field Ising model — small enough to run in seconds
g = nk.graph.Chain(length=4, pbc=True)
hi = nk.hilbert.Spin(s=0.5, N=g.n_nodes)

sampler = nk.sampler.MetropolisLocal(hi, n_chains=16)
model = nk.models.RBM(alpha=1)
vstate = nk.vqs.MCState(sampler, model, n_samples=512)

optimizer = nk.optimizer.Sgd(learning_rate=0.01)
driver = nk.driver.VMC(ha, optimizer, variational_state=vstate)

# Debug callback: show exactly what type "acceptance" has when it reaches the loggers
class DumpStatsType(nk.callbacks.AbstractCallback):
    def on_step_end(self, step, log_data, driver):
        if "acceptance" in log_data:
            print(f"step: {step}:\ntype(log_data['acceptance']): {type(log_data['acceptance'])}, value: {log_data['acceptance']}")
            print(f"type(log_data['Energy']): {type(log_data['Energy'])}, value: {log_data['Energy']}")
        return None

logger = nk.logging.TensorBoardLog("issue_logdir")

driver.run(
    n_iter=10,
    out=logger,
    callback=[ DumpStatsType(), CastAcceptance()],
)

# After running: inspect what actually landed on disk
# $ tensorboard --inspect --logdir=issue_logdir
# -> "acceptance" tag is absent, even though DumpStatsType printed it every step
```
 
## Expected behavior
 
`acceptance` should be logged as a scalar tag. It is currently logged correctly via `MLFlowLog` in the same run, which handles this value without issue, so the discrepancy appears to be specific to `TensorBoardLog`'s handling of raw JAX-array scalars.
 
## Actual behavior
 
`acceptance` is silently dropped from the TensorBoard output without exception or warning.
 
## Workaround
 
Casting the value to `float` before it reaches `TensorBoardLog`, via an extra `callback=[...]` entry that runs first, fixes it:
```python
class CastAcceptance(nk.callbacks.AbstractCallback):
    def on_step_end(self, step, log_data, driver):
        if "acceptance" in log_data:
            log_data["acceptance"] = float(log_data["acceptance"])
        return None
 
driver.run(n_iter=..., out=tb_logger, callbacks=[CastAcceptance()])
```

PhilipVinc:
Closes #2244

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
 
 ## NetKet 3.23 (In development)
 
+### Bug Fixes
+* {class}`netket.logging.TensorBoardLog` no longer silently drops metrics that arrive as `jax`/`numpy` scalar arrays (such as `acceptance`); scalar (0-dim or single-element) arrays are now logged, with complex scalars split into `/re` and `/im` tags [issue #2244](https://github.com/netket/netket/issues/2244).
+
 ...
 
 ## NetKet 3.22 (4 June 2026)
--- a/netket/_src/callbacks/save_state.py
+++ b/netket/_src/callbacks/save_state.py
@@ -100,7 +100,11 @@ class SaveVariationalState(AbstractCallback, mutable=True):
             if jax.process_index() == 0:
                 pattern = re.compile(rf"^{re.escape(self._file_name_root)}_(\d+)\.nk$")
                 saved = sorted(
-                    (p for p in self._path.glob(f"{self._file_name_root}_*.nk") if pattern.match(p.name)),
+                    (
+                        p
+                        for p in self._path.glob(f"{self._file_name_root}_*.nk")
+                        if pattern.match(p.name)
+                    ),
                     key=lambda p: int(pattern.match(p.name).group(1)),
                 )
                 for old in saved[: -self._max_to_keep]:
--- a/netket/_src/logging/mlflow_log.py
+++ b/netket/_src/logging/mlflow_log.py
@@ -217,7 +217,9 @@ class MLFlowLog(AbstractCallback):
         data: list[tuple[str, Any]] = []
         if self._ignore:
             item = {k: v for k, v in item.items() if k not in self._ignore}
-        walk_tree_with_path(item, "", visit_leaf=_visit_leaf, expand_node=_expand_node, data=data)
+        walk_tree_with_path(
+            item, "", visit_leaf=_visit_leaf, expand_node=_expand_node, data=data
+        )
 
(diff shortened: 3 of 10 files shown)


gcarleo:
genuine question: do people still use tensorboard?

PhilipVinc:
It seems at least `@FerdinandToelkes` who opened the bug report.
I use MLFlow because it works on our cluster but it's the same thing.

FerdinandToelkes:
> It seems at least `@FerdinandToelkes` who opened the bug report. I use MLFlow because it works on our cluster but it's the same thing.

Yes I use Tensorboard because it was the easiest to setup on our cluster without having root/sudo rights. MLFlow's file based approach is legacy by now and using MLFlow with a database file without a tracking server leads to problems when running multiple runs in parallel. The tracking server was (at least for me) not straight forward to setup on our server without having root rights.

Issue #2258, opened 2026-07-31
Breaking changes in JAXMg


therooler:
My collaborator and I just released a new version of [JAXMg (v1.0.0)](https://flatironinstitute.github.io/jaxmg/latest/) that supports multi-GPU and (the much requested) multi-node capabilities.

This was a significant rewrite of the original code but should be both faster and more flexible going forward and allow us to implement more of the new features of [NVIDIA's cuSOLVERMp API](https://docs.nvidia.com/cuda/cusolvermp/usage/functions.html#dense-linear-algebra-apis).

For NetKet, this means that the current version of JAXMg must be pinned to <1.0.0 in order to avoid breaking the distributed solvers.

I intend to make a PR in the future to switch NetKet to new API, which should be straightforward but will require some care.

I'm particularly excited to try to see if the QGT construction and inversion can be done on the 2D GPU grid that JAXMg now supports, avoiding the need for any all-gathers of the QGT blocks.

PR #2263, opened 2026-08-12
Pin jaxmg to <1.0.0 for the distributed solvers

PhilipVinc:
Addresses #2258.

[JAXMg 1.0.0](https://flatironinstitute.github.io/jaxmg/latest/) is a rewrite on top of NVIDIA's cuSOLVERMp: it changes the API and *requires one python process per GPU*, while {`cholesky_distributed`, `pinv_smooth_distributed`} are written against the 0.0.x cuSOLVERMg API and drive several GPUs from a single process.

This is only the stop-gap pin asked for in the issue — porting the solvers to the new API is left to the follow-up PR @therooler announced.

Note that this is not merely hypothetical: `main` already declares `jaxmg>=0.0.9; sys_platform == 'linux'` in the `extra` extra, so `pip install netket[extra]` on Linux currently installs the incompatible 1.0.0.

## Changes

* `pyproject.toml`: bound the requirement to `jaxmg>=0.0.9,<1.0.0`.
* `import_optional_dependency` now actually *enforces* version bounds. `minimum_version=` existed but was only used to format the `pip install` hint in the not-found message; it is now checked, and `maximum_version=` (exclusive) and `extra_msg=` were added. If the version cannot be determined (no `__version__`, no distribution metadata) the check is skipped rather than failing.
* Both distributed solvers pass those bounds, so an unsupported jaxmg fails upfront with an actionable message instead of erroring obscurely inside jaxmg:

  ```
  `jaxmg` version 1.0.0 is not supported by
  `cholesky_distributed solver`, which requires `jaxmg>=0.0.9,<1.0.0`.

  To install a supported version, run

      pip install 'jaxmg>=0.0.9,<1.0.0'

  `jaxmg` 1.0 is a rewrite on top of NVIDIA's cuSOLVERMp, with a different
  API and execution model (it requires one python process per GPU). Support
  for it in NetKet is not implemented yet, see
  https://github.com/netket/netket/issues/2258 .
  ```
* Docs warning + changelog entry.

## Tests

* New `test/utils/test_optional_deps.py` covering the bound checking.
* New `test_unsupported_jaxmg_version` in `test/optimizer/test_solver_distributed.py`, which injects a fake `jaxmg` 1.0.0 and checks both solvers refuse it. Since the guard fires before any GPU work, this one runs on CPU — the module-level `pytestmark` GPU skip became a per-test `@requires_gpu` decorator on the existing (kernel-executing) tests so it isn't skipped everywhere.

`test/optimizer` + `test/utils` pass locally (1955 passed, 10 skipped). The GPU-only solver tests were not exercised.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


PhilipVinc:
Superseded by #2265, which ports the solvers to the new JAXMg 1.0 API instead of pinning `jaxmg<1`.

PR #2264, opened 2026-08-12
Speed up CI by caching XLA compilations and using all cores

PhilipVinc:
The test suite is compilation-bound, and that compilation was being repeated both across xdist workers within a run and across every CI run.

- Enable jax's persistent compilation cache, stored in the Actions cache and keyed on a fingerprint of the compilation environment (the ubuntu fleet is mixed-CPU and keys track the CPU target — see comments below). Written on pushes to `main`, saved with `always()`.
- Run the standard-tests-under-sharding script with 2 xdist workers instead of serially.
- `-n logical` was tried and reverted: 4 workers is faster but OOM-killed the runner twice.

| Job | baseline (median of 8 runs on main) | warm cache |
|---|---|---|
| Ubuntu | 47.6 min | **19.9 min** |
| macOS | 46.4 min | **18.0 min** |
| Sharding (Standard) | 43.9 min | 21.7 min (cold) |

Critical path **~47 → ~20 min**. A hit requires identical HLO and target, so this cannot mask a change in netket.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -62,11 +62,12 @@ jobs:
             sharding_script: "run_test_sharding_distributed.sh"
 
     env:
-      # Reuse XLA compiled executables across xdist workers and across CI runs.
-      # The test suite is dominated by compilation, not execution, so this is
-      # worth roughly a 2x reduction in CPU time. The cache key is derived from
-      # the HLO, so a cache hit is only ever an identical computation and this
-      # cannot mask a change in netket.
+      # Reuse XLA compiled executables. The test suite is dominated by
+      # compilation, not execution, so this pays off twice: the xdist workers of
+      # a single run stop compiling the same kernels independently, and on macOS
+      # the cache is also carried between runs (see the restore/save steps).
+      # A cache key is derived from the HLO, so a hit is only ever an identical
+      # computation and this cannot mask a change in netket.
       JAX_COMPILATION_CACHE_DIR: ${{ github.workspace }}/.jax-cache
       # Default is 1s, which skips the many small compilations that make up most
       # of the suite's time.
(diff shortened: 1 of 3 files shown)


PhilipVinc:
### First pass, cold caches

| Job | before | after |
|---|---|---|
| Ubuntu | 50.1 min | runner died at 81% |
| macOS | 48.4 min | 36.5 min |
| Sharding (Standard) | 43.9 min | 24.1 min |

`-n logical` did give the ubuntu runner all 4 vCPUs (`created: 4/4 workers`), but two runs in a row died with `The runner has received a shutdown signal` at 81% and 93% — no failing test, no crashed worker, i.e. an OOM-killed agent. Reverted.

Superseded by the results below.

PhilipVinc:
### Warm cache, plus an `always()` fix

macOS reuses its cache properly — steady 169MB, 46.4 → 21 min. Linux hit nothing and its cache doubled every run (184 → 367MB), so at this point persistence was made macOS-only.

Also fixed here: the save runs under `always()`. A compiled executable is valid whether or not the test asserting on it passed, and one flaky test had been skipping the save in two jobs at once.

Linux root cause in the next comment.

PhilipVinc:
## Root cause: mixed-CPU runner fleet

Cache filenames *are* jax's keys. Printing them on four `ubuntu-latest` runners at once:

| runner | CPU | first key |
|---|---|---|
| 1 | EPYC 9V74 | `jit__lambda-7bb5c61a…` |
| 2–4 | EPYC 7763 | `jit__lambda-3b5d4a32…` (identical) |

Keys track the host CPU target and the fleet is mixed, so consecutive runs share nothing; macOS is uniform Apple silicon, hence unaffected. This is invisible from jax's source (`platform_version` is just `'cpu'`), which is why I wrongly ruled it out earlier. Also excluded, on a fixed host where reuse is perfect: xdist, the tar/restore round-trip, dependency drift, and `--xla_cpu_max_isa` capping.

**Fix:** keys are stable per CPU model, so cache per compilation environment, fingerprinted by hashing the filenames one trivial jit emits. Each fingerprint now holds a stable ~186MB cache instead of one that doubled and never hit.

| Job | baseline (median of 8 on main) | warm |
|---|---|---|
| Ubuntu | 47.6 min | **19.9** |
| macOS | 46.4 min | **18.0** |
| Sharding (Standard) | 43.9 min | 21.7 (cold) |

One cold run per CPU model as the caches fill (3 linux fingerprints seen so far), then steady hits.

PR #2265, opened 2026-08-12
Port the distributed solvers to JAXMg 1.0 (cuSOLVERMp)

### LINKED ISSUE #2258: Breaking changes in JAXMg

PhilipVinc:
Closes #2258 by moving to the new API rather than pinning `jaxmg<1`.

[JAXMg 1.0](https://flatironinstitute.github.io/jaxmg/) rewrote everything on top of NVIDIA's cuSOLVERMp, replacing the deprecated cuSOLVERMg backend of the 0.0.x series. In exchange for requiring **one python process per GPU**, it works **across several nodes** — which the old backend never did, so the "single node only" warning in the docs is gone.

The new interface distributes the matrix over a **two-dimensional** process grid, which is most of the work here.

## What changed in the solvers

`cholesky_distributed` / `pinv_smooth_distributed` now:

* **Build a dedicated 2D device mesh** over NetKet's devices. cuSOLVERMp requires *both* matrix dimensions to be mapped to a named mesh axis (`P(row_axis, col_axis)`; `P("S", None)` is rejected), and NetKet's own mesh only has the single `S` axis. The grid defaults to `(n_devices, 1)` — which is exactly how NetKet already shards QGT/NTK matrices, so nothing is redistributed, and it is the configuration JAXMg's own README uses. The new **`process_grid=(process_rows, process_cols)`** argument opts into a genuinely 2D grid, which cuts the communication volume of the factorisation at the cost of redistributing `A` first.
* **Enter that mesh as the context mesh** around the JAXMg call. Without it, JAXMg's internal `jax.shard_map` raises `ValueError: The context mesh AbstractMesh('S': 4, …) should match the mesh passed to shard_map`. `jax.sharding.use_abstract_mesh` is used rather than `jax.set_mesh`, because the latter raises ``set_mesh` can only be used outside of `jax.jit`` and these solvers run inside jitted driver code.
* **Pick a tile size that cuSOLVERMp accepts**: it requires every process row/column to own at least one block-cyclic tile, i.e. `ceil(N/T_A) >= max(process_rows, process_cols)`. The old default (`max(64, N // n_devices)`) could violate that. On the default grid the new default equals the local row count, so tiles align with the jax shards and no padded copy of `A` is allocated.
* **Check the matrix is evenly distributable** over the grid. cuSOLVERMp requires `N % process_rows == 0` and `N % process_cols == 0`; JAXMg's own message (`logical_rows must be divisible by process_rows`) doesn't say what to do about it, so NetKet raises a message that names the sizes and points at `process_grid=`.
* **Replicate the solution**, which also detaches it from the grid mesh (otherwise an eager call returns an array whose sharding refers to the JAXMg mesh, and later cross-mesh `with_sharding_constraint` calls fail).
* Let JAXMg reshape a 1D rhs itself, since 1.0 does that (and restores the rhs sharding on the way out).

Supporting changes:

* `import_optional_dependency` now actually **enforces** version bounds — `minimum_version=` existed but was only used to format the `pip install` hint — plus `maximum_version=` and `extra_msg=`. An older jaxmg now fails with an actionable message instead of a cryptic one deep inside jaxmg. Its keyword arguments became keyword-only so the new parameter can't silently capture a positional `descr`.
* **jaxmg moves out of the `extra` extra into its own `jaxmg` extra.** jaxmg 1.0 declares `jax==0.10.1` on Linux (0.0.9 only asked for `>=0.6.2`), and `main` currently lists `jaxmg>=0.0.9` under `extra` — so `pip install netket[extra]`, including CI's `uv sync --extra extra`, would silently pin everyone's jax to 0.10.1. The extra requests `jaxmg[cuda12]`, matching NetKet's existing `cuda` extra, since the base package does not pull the cuSOLVERMp runtime.
* The two `*_with_sharding` GPU tests passed `mesh=`/`in_specs=` kwargs that no solver has ever accepted (they're always skipped, so it never showed); they now shard `A` the way NetKet does and let the solver build the grid. GPU test matrix sizes scale with the device count, since one-process-per-GPU makes tiny matrices undistributable.

## Testing

The cuSOLVERMp kernels are CUDA-only and want one process per GPU, so **none of this can run in CI, and I have no GPU here**. What I did instead:

1. **New CPU tests** (`test/optimizer/test_solver_distributed.py`) install a fake `jaxmg` that solves with plain jax, and assert NetKet passes a 2D `Mesh` covering every device, `matrix_specs` naming both of its axes, a matrix size divisible by both grid dimensions, and a tile size satisfying the ownership rule — for *every* process grid the available devices allow, eager and under `jax.jit`. On the 4 CPU devices the test conftest provides that covers `(4,1)`, `(2,2)` and `(1,4)`; the "Sharding (Standard)" CI job re-runs them on 2 devices.
2. **Locally, against real JAXMg 1.0**: I installed jaxmg 1.0.0's python layer and faked *only* the native FFI pipeline, so its real validation, `shard_map`, rank mapping, padding and redistribution code all ran. All 16 combinations (2 solvers × 4 grids × eager/jit) passed with correct numerics, including with `A` sharded `P("S", None)` and `b` replicated as the drivers do, and with a tile size that forces the padding path. A `VMC_SR` run (3 iterations, `use_ntk` both ways, both solvers) also completed. Not committed, since it needs jaxmg installed.

So the plumbing is verified as far as the Python layer goes; **the actual cuSOLVERMp kernels have not been exercised**, and the `(n_devices, 1)` vs 2D grid performance trade-off is unmeasured — the default is the conservative choice that preserves today's data layout, and is worth benchmarking on real hardware before changing.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


pyproject.toml
```diff
@@ -52,7 +52,13 @@ dependencies = [
 [project.optional-dependencies]
 cuda = ["jax[cuda12]; sys_platform == 'linux'"]
 pyscf = ["pyscf>=2.0"]
-extra = ["tensorboardx>=2.0.0", "openfermion>=1.0.0", "h5py>=3.7.0", "qutip>=4", "mlflow>=2.0.0", "jaxmg>=0.0.9; sys_platform == 'linux'"]
+extra = ["tensorboardx>=2.0.0", "openfermion>=1.0.0", "h5py>=3.7.0", "qutip>=4", "mlflow>=2.0.0"]
+# jaxmg is kept in its own extra, and out of `extra`, because it pins an exact
+# jax version (jax==0.10.1 for jaxmg 1.0.0) which we do not want to force onto
```
therooler, line 57: Just mentioning here that the jax==0.10.1 is there to pin the XLA docker container, which will be largely unchanged through newer jax versions.

We'll make sure that newer Jax versions are supported by testing them in the CI.

therooler:
I'll test this implementation later this week, thanks for making the PR.

pyproject.toml
PhilipVinc: I think you should set it at least as `jax~=0.10.1` in JAXMg (this means >= 0.10.1, <0.11) so we get bug fixes.

In general, I'd suggest you don't put upper bounds because they are a pain in the python ecosystem to manage. 
I tried in the past, and it's very annoying. I'd suggest you don't do it..

PhilipVinc:
## Now validated on real GPUs (4× H100, Jean Zay, one process per GPU)

All green with stock `jaxmg==1.0.0`:

* both solvers × the four process grids `None`/`(4,1)`/`(2,2)`/`(1,4)` × eager and `jax.jit` — 16/16 correct against `jnp.linalg.solve`;
* `VMC_SR` for 5 iterations, both solvers, `use_ntk` either way.

### One real bug the GPUs found, now fixed

`jaxmg.potrs`/`syevd` wrap their pipeline in `jax.jit(donate_argnums=...)`, so **they delete the caller's `A`**. Eagerly, that made

```python
nk.optimizer.solver.nan_fallback(cholesky_distributed, cholesky)
```

die with `RuntimeError: Array has been deleted with shape=float64[2048,2048]` — `nan_fallback` reads `A` again after the primary solver. Under `jit` the inner donation is dropped, so it only bit outside `jit`.

Fixed by calling jaxmg's `potrs_shardmap_ctx`/`syevd_shardmap_ctx` instead: same pipeline, no donating jit. Re-verified on GPU — `A` and `b` survive in both modes and the combinator works. There's a CPU regression test for it.

pyproject.toml
therooler: Alright, I can do that.

therooler:
Not donating the buffers under jit can result in the copy of the local shard, which results in OOM if you're near the memory limit. `nan_fallback` should probably warn against that.

therooler:
Posting this in here in reply to your question about 1D sharding vs 2D sharding from this PR https://github.com/flatironinstitute/jaxmg/pull/28

Here you see that at scale, the 4x4 grid sharding can be much faster than the 16x1 sharding.

image

PR #2277, opened 2026-09-16
Fold force_to_grad factor of two into the real-parameter branch

### LINKED ISSUE #2190: Jacobian of complex parameters multiplied by 2 in VMC
It seems to be a mismatch between the docstring and code: For complex-parameterized models, the Ō_grad will still be multiplied by 2?
https://github.com/netket/netket/blob/63b594245843ac71a6d1390a4fed268dbba66360/netket/vqs/mc/common.py#L64-L80

shaneraphel:
Fixes #2190.

`force_to_grad` is documented as the identity on a complex holomorphic target and as `2 Re[F]` on real parameters. Current still takes `x.real` for real leaves and then multiplies the whole tree by 2, so a complex parameter's force is doubled. `expect_and_grad` already routes through this helper and is not rewritten.

Unique rewrite folds the factor of two into the real branch:

```python
        lambda x, target: (x if jnp.iscomplexobj(target) else 2 * x.real).astype(
```

Measured on dummy scalar forces (`n_paired=3`; not a VMC run, not JAX):

| parameters | current | after rewrite |
|---|---|---|
| real, F=0.4 | 0.8 (control) | 0.8 |
| complex, F=0.4+0.3j | 0.8+0.6j | 0.4+0.3j |
| complex, F=0.3j | 0.6j | 0.3j |


PhilipVinc:
@shaneraphel good find. I suspect this could be related to why we have not been able to solve the Hubbard model yet with netket. Can you check if it solves it?

shaneraphel:
jax and netket are not importable in the environment that produced this PR, so I cannot run a Fermi-Hubbard VMC. The dummy scalar lattice in the description is not a VMC run and is not a Hubbard energy.

`force_to_grad` is documented as the identity on a complex holomorphic target. The leftover doubled a complex leaf by applying `2` after taking `.real` only on real parameters. The rewrite puts the factor of two only on the real branch. That is the documented map. It is necessary for holomorphic forces and is not a Hubbard solve.

